PR #2, opened 2023-01-25
folder.to() like functionality between data sources

jlewitt1:
This PR adds the `folder.to()` like functionality between various data store sources (local, cluster, s3, gcs).  This enables a more fluid copying of folders between each of the sources. Some other small changes were made to the Cluster, Table and Blob classes to support this.

_Folder_: a couple new APIs:

- `upload`: upload a folder to a remote data store 
- `download`: download folder from remote data store
- `to_data_store`: Copy folder from one data store to another (ex: GCS —> S3)
- `to_cluster`: Copy folder from a local or remote data store onto a cluster
- `delete_in_fs`: Delete the folder itself (along with its contents)
- `empty_folder`: Delete the folder contents, but keep the folder

_Cluster_: 
- Replace SFTP file system with the faster SSHFS

_Table_: 
- Add option to stream Huggingface dataset batches as dict rather than converting them back to a Dataset object


runhouse/rns/resource.py
```diff
@@ -53,7 +53,7 @@ def _resource_string_for_subconfig(self, resource):
                 # Calls save internally and puts the resource in the current folder
                 resource.name = rns_client.resolve_rns_path(resource.rns_address[1:])
             # TODO [DG] not sure if we should save here
-            resource.save()
+            # resource.save()
```
jlewitt1, line 56: @dongreenberg I think we don't need this now that we enforce calling `save()` more explicitly on resources
wdyt?

runhouse/rns/folders/gcs_folder.py
```diff
@@ -6,33 +8,91 @@ class GCSFolder(Folder):
     DEFAULT_FS = 'gcp'
 
     def __init__(self, dryrun: bool, **kwargs):
-        # TODO [JL]
         super().__init__(dryrun=dryrun, **kwargs)
 
     @staticmethod
     def from_config(config: dict, dryrun=True):
```
dongreenberg, line 14: Also override from_name to create cluster if needed as fs

runhouse/rns/folders/folder.py
```diff
@@ -678,9 +762,16 @@ def put(self, contents, overwrite=False):
             with fss_file as f:
                 f.write(raw_file)
 
+    @staticmethod
+    def bucket_name_from_url(url: str) -> str:
+        """Extract the bucket name from a URL (e.g. '/my-bucket/my-folder/my-file.txt' -> 'my-bucket')"""
+        split_urls: list = list(filter(None, url.split('/')))
```
carolineechen, line 768: fyi `.split('/')` may not always work, if `url` can refer to a local path on Windows (which uses forward slash `\`)

runhouse/rns/folders/gcs_folder.py
```diff
@@ -6,33 +9,92 @@ class GCSFolder(Folder):
...
-        pass
+    def to_cluster(self, dest_cluster, url=None, mount=False, return_dest_folder=False):
+        upload_command = self.upload_command(src=self.url, dest=url)
+        dest_cluster.run([upload_command])
+        if return_dest_folder:
+            return GCSFolder(url=url, dryrun=True).from_cluster(dest_cluster)
 
-    def to_blob_storage(self, fs, url=None, data_config=None):
-        pass
+    def to_local(self, dest_url: str, data_config: dict, return_dest_folder: Optional[bool] = False):
```
carolineechen, line 72: `Optional[bool]` params like `return_dest_folder` should just be type `bool`

PR #3, opened 2023-01-26
Object Store, Streaming Logs, and some other stuff

dongreenberg:
1) Introduce Object Store, a wrapper around the ray kv store but allowing string keys. Automatically returns/prints a convenient string to retrieve results or logs from remoted sends.
2) Introduce initial support for streaming stdout and logs during Send execution. 
3) Introduce a way to tell which cluster we're on. 
4) Fix stickiness bug where gRPC server and/or Ray sometimes don't respect module or package updates after already having imported them. 
5) Break out the function call logic from the gRPC server to allow us to call it directly if we detect that the Send is being called on its own cluster (but actually calling that way is still TODO).

Added basic tests at test_obj_store, all pass and print logs (but not yet explicitly testing if the logs came in).

Still todo:

- [ ] test all tutorials (Update 27-Jan: tested some, still debugging and testing more)
- [x] handle in-function exceptions returned by get()
- [x] add more tests
- [ ] refactor the other send call_types to support streaming logs
- [x] Add `cluster.cancel(run_key)` and `runhouse cancel cluster:run_key`
- [x] Add `runhouse logs cluster:run_key`
- [x] Fix bug with setting num_gpus for Ray (based on whether or not GPUs are available)


runhouse/rns/hardware/cluster.py
```diff
@@ -0,0 +1,405 @@
...
+                 ):
+        """
+        Args:
+            name: Name of the cluster
+            dryrun:
+        """
+
+        super().__init__(name=name, dryrun=dryrun)
+
+        self.address = ips[0] if ips else None
```
carolineechen, line 40: I'm a little bit confused about `ips` being a list of ip addresses to pass in, but only the first one is used through `self.address`.

carolineechen: also in what situation would we expect `Cluster` to be initialized with `ips=None`?

runhouse/rns/hardware/cluster.py
```diff
@@ -0,0 +1,405 @@
...
+        # netstat -vanp tcp | grep 5005
+        # lsof -i :5005_
+        # kill -9 <pid>
+
+        creds: dict = self.ssh_creds()
+        connected = False
+        ssh_tunnel = None
+        while not connected:
+            try:
+                if local_port > local_port + num_ports_to_try:
```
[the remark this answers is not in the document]
dongreenberg: 🤦 great catch! It's probably been like this for weeks.

PR #5, opened 2023-02-12
Logout and secrets to cluster

jlewitt1:
**Logout**: Give user 2 main options: 
- `delete_loaded_secrets`: deletes configured secret providers' credential files
- `delete_rh_config_file`: deletes the runhouse config file

**Secrets to Cluster**: Ability to send secrets to a cluster and save them down into the cluster's file system


runhouse/rns/secrets.py
```diff
@@ -117,14 +128,13 @@ def put(
...
             # if a supported cloud provider is given, use the provider's built-in class
-            provider_cls_name = cls.provider_cls_name(provider)
-            p = cls.get_class_from_name(provider_cls_name)
+            p = cls.builtin_provider_class_from_name(provider_name)
             if not from_env and not p.has_secrets_file():
                 # no secrets file configured for this provider
-                raise Exception(f"No local secrets file found for {provider}")
+                raise Exception(f"No local secrets file found for {provider_name}")
 
             secret = p.read_secrets(from_env=from_env, file_path=file_path)
```
dongreenberg, line 139: If the user passed in the secrets or file_path this is throwing an error for no reason (it doesn't matter that we didn't find one locally), or overwriting their provided secrets if we find others locally. We shouldn't enter this block if they passed in a secret, and we shouldn't throw an exception if they provided a file_path.

runhouse/rns/secrets.py
```diff
@@ -146,100 +156,122 @@ def get(
...
 
-        secrets = cls.load_resp_data(resp).get(provider, {})
+        secrets = cls.load_resp_data(resp).get(provider_name, {})
         if not secrets:
-            raise Exception(f"Failed to load secrets for {provider}")
+            raise Exception(f"Failed to load secrets for {provider_name}")
 
-        cls_name = cls.provider_cls_name(provider)
-        p = cls.get_class_from_name(cls_name)
+        p = cls.builtin_provider_class_from_name(provider_name)
```
dongreenberg, line 170: Much better!

runhouse/rns/secrets.py
```diff
@@ -146,100 +156,122 @@ def get(
...
-            logger.info(f"Saving secrets for {provider} to local config")
+            logger.info(f"Saving secrets for {provider_name} to local config")
             p.save_secrets(secrets)
 
         return secrets
 
     @classmethod
-    def delete(cls, providers: List[str]):
-        """Delete secrets from Vault for the specified providers"""
+    def delete_from_vault(cls, providers: Optional[List[str]] = None):
```
dongreenberg, line 179: Much clearer!

runhouse/rns/secrets.py
```diff
@@ -146,100 +156,122 @@ def get(
...
+        if not provider_secrets:
+            # If no secrets saved in local config files, check if they exist in Vault
+            vault_secrets: dict = cls.download_into_env(save_locally=False)
+            # TODO [JL] change this API so we don't have to convert the list to a dict?
+            provider_secrets: list = [
+                {"provider": k, **v} for k, v in vault_secrets.items()
+            ]
+
+        # Send provider secrets over RPC to the cluster, then save each provider's secrets into their respective files
+        hardware.add_secrets(provider_secrets)
```
dongreenberg, line 216: Oh I see, this is where we use the RPC. So inside the cluster class, let's just make send_secrets call `Secrets.to(self` so it goes through the same code path. It's also nice that we load from vault if the secrets aren't saved locally, which we don't have in that method right now.

runhouse/rns/secrets.py
```diff
@@ -146,100 +156,122 @@ def get(
...
+            hardware = cluster(name=hardware)
+
+        hardware_name = hardware.name
+        if not hardware.is_up():
+            raise RuntimeError(
+                f"Hardware {hardware_name} is not up. Run `hardware_obj.up()` to re-up the cluster."
+            )
+
+        provider_secrets: list = cls.load_provider_secrets(providers=providers)
+        if not provider_secrets:
```
dongreenberg, line 207: It looks like we'll only pull from vault if a user passes in a list and none of them are found (vs. if just one or two aren't found locally). Maybe we should check the len of provider_secrets and see if it's shorter than providers (if providers is not None), and if so, try to pull from vault and fill those in? Then, if some still haven't been found, we should raise an exception saying secrets can't be found for x providers.

runhouse/rns/secrets.py
```diff
@@ -385,10 +434,12 @@ def save_secrets(cls, secrets: dict, file_path: Optional[str] = None):
             value=secrets["secret_key"],
         )
 
-        cls.save_to_config_file(parser, dest_path)
+        if overwrite:
+            cls.save_to_config_file(parser, dest_path)
+            cls.save_secret_to_config()
 
 
-class AZURESecrets(Secrets):
```
dongreenberg, line 391: This file is getting pretty huge, do you mind breaking each of the providers into separate files?

runhouse/rns/secrets.py
```diff
@@ -385,10 +434,12 @@ def save_secrets(cls, secrets: dict, file_path: Optional[str] = None):
             value=secrets["secret_key"],
         )
 
-        cls.save_to_config_file(parser, dest_path)
+        if overwrite:
```
dongreenberg, line 437: It looks like we only write down the secrets if overwrite is True (even if there's no conflict with an existing file), and it's false by default. If think we need to check if the file is already present and then check if overwrite is False before we decide not to write down?

runhouse/rns/secrets.py
```diff
@@ -419,7 +472,9 @@ def save_secrets(cls, secrets: dict, file_path: Optional[str] = None):
             value=secrets["subscription_id"],
         )
 
-        cls.save_to_config_file(parser, dest_path)
+        if overwrite:
```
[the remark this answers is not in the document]
dongreenberg: Maybe all the "save to file" methods can take an overwrite argument that does this check, so we can just pass it through and don't need to do it in each provider class.

runhouse/rns/secrets.py
```diff
@@ -471,10 +528,49 @@ def save_secrets(cls, secrets: dict, file_path: Optional[str] = None):
...
+        else:
+            from sky.clouds.lambda_cloud import lambda_utils
+
+            client = lambda_utils.LambdaCloudClient()
+            api_key = client.api_key
+            ssh_key_name = client.ssh_key_name
+        return {
+            "provider": cls.PROVIDER_NAME,
+            "api_key": api_key,
+            "ssh_key_name": ssh_key_name,
```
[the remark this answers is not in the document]
dongreenberg: My apologies if I added this by accident, I don't remember

runhouse/rns/secrets.py
```diff
@@ -485,32 +581,42 @@ def read_secrets(cls, from_env: bool = False, file_path: Optional[str] = None):
...
+    ):
         # TODO check properly if hf needs to be installed
         try:
             import huggingface_hub
         except ModuleNotFoundError:
             subprocess.run(["pip", "install", "--upgrade", "huggingface-hub"])
             import huggingface_hub
-        huggingface_hub.login(token=secrets["token"])
+
+        if overwrite:
```
dongreenberg, line 600: Same overwrite comment as above.

runhouse/rns/secrets.py
```diff
@@ -519,8 +625,8 @@ def read_secrets(cls, from_env: bool = False, file_path: Optional[str] = None):
                 f"Reading secrets from env is not supported for {cls.PROVIDER_NAME}"
             )
         else:
-            private_key_file = file_path or cls.PRIVATE_KEY_FILE
-            public_key_file = file_path or cls.PUBLIC_KEY_FILE
+            private_key_file = cls.PRIVATE_KEY_FILE
+            public_key_file = cls.PUBLIC_KEY_FILE
```
dongreenberg, line 629: Why do we ignore file_path here?

dongreenberg: (I see it was wrong before, we need to add ".pub" to it to get the public key file)

runhouse/rns/secrets.py
```diff
@@ -532,22 +638,26 @@ def read_secrets(cls, from_env: bool = False, file_path: Optional[str] = None):
...
-        sky.authentication._save_key_pair(
-            private_key_path,
-            public_key_path,
-            secrets["ssh_private_key"],
-            secrets["ssh_public_key"],
-        )
-        # TODO do we need to register the keys with cloud providers? Probably not, sky does this for us later.
-        # backend_utils._add_auth_to_cluster_config(sky.clouds.CLOUD_REGISTRY.from_str(self.provider),
-        #                                                   Path(yaml_path).expanduser())
+        if overwrite:
```
dongreenberg, line 650: Same as above

runhouse/rns/secrets.py
```diff
@@ -562,19 +672,32 @@ def read_secrets(cls, from_env: bool = False, file_path: Optional[str] = None):
...
+            cls.save_secret_to_config()
+
+
+class Providers(Enum):
+    AWS = AWSSecrets()
+    AZURE = AzureSecrets()
+    GCP = GCPSecrets()
+    HUGGINGFACE = HuggingFaceSecrets()
+    LAMBDA = LambdaSecrets()
+    SKY = SkySecrets()
```
dongreenberg, line 700: Love it, nice solution. Maybe better to move it to the top of the file above the Secrets class, that's the convention I'm used to seeing for Enums.

runhouse/rns/hardware/cluster.py
```diff
@@ -451,39 +456,12 @@ def run(
...
-            # Send secrets for specific providers from local configs rather than trying to load from Vault
-            from runhouse import Secrets
-
-            secrets: list = Secrets.load_provider_secrets(providers=providers)
-            # TODO [JL] change this API so we don't have to convert the list to a dict
-            secrets: dict = {
-                s["provider"]: {k: v for k, v in s.items() if k != "provider"}
-                for s in secrets
-            }
-            load_secrets_cmd = [
```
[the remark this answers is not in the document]
dongreenberg: See comment below, we can either go through add_secrets or `Secrets.to`. `Secrets.to` would get us the extra hop to check vault, which could be nice, but I guess it's a little circular.

tests/test_secret.py
```diff
@@ -46,7 +46,28 @@ def test_upload_all_provider_secrets():
...
+
+    configured_providers = rh.Secrets.configured_providers()
+
+    rh.Secrets.to(cluster, providers=configured_providers)
+    # Confirm the secrets now exist on the cluster
+    for provider_cls in configured_providers:
+        provider_name = provider_cls.PROVIDER_NAME
+        p_str = str(provider_cls())
+        command = [
+            f"from runhouse.rns.secrets import {p_str}",
```
dongreenberg, line 64: Nice! Good check

dongreenberg (commented):
Awesome PR. Solid cleanup and new functionality. Few things:
0) Breaking up into separate files by provider will make this more readable and manageable.
1) If you can add a check to make sure we're not overwriting SSH keys in the SSHSecrets I think that's needed (for each key, not just checking if .ssh/ is present). If you overwrite an SSH key you can never get it back. 
2) For overwrites in general, If we see a file already exists for a key we want to save I would check whether the contents match before throwing an error.

Great work!

PR #10, opened 2023-02-26
Correctly add relative path to sys.path

carolineechen:
Move adding relative path to sys.path from `RunModule` to `_get_fn_name`. In [multi gpu tutorial example](https://github.com/run-house/tutorials/blob/main/t04_Distributed/p01a_accelerate_multigpu_example.py), where there is nested function calls, only the external relative path (`tutorials/t04_Distributed`) was added to sys.path, and not relative path for the internal rh function (`accelerate/examples`)

### DIFF
--- a/runhouse/servers/grpc/unary_server.py
+++ b/runhouse/servers/grpc/unary_server.py
@@ -1,8 +1,6 @@
 import logging
-import sys
 import traceback
 from concurrent import futures
-from pathlib import Path
 
 import grpc
 
@@ -172,11 +170,6 @@ class UnaryService(pb2_grpc.UnaryServicer):
             module_path = None
 
             if module_name == "notebook":
-                # If relative_path is None, the module is not in the working dir, and should be in the reqs
-                if relative_path:
-                    module_path = str((Path.home() / relative_path).resolve())
-                    sys.path.append(module_path)
-                    logger.info(f"Appending {module_path} to sys.path")
                 fn = fn_name  # Already unpickled above
             else:
                 fn = get_fn_by_name(module_name, fn_name, relative_path)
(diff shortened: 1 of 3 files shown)


runhouse/servers/grpc/unary_server.py
```diff
@@ -170,16 +170,16 @@ def RunModule(self, request, context):
...
 
             module_path = None
-            # If relative_path is None, the module is not in the working dir, and should be in the reqs
-            if relative_path:
-                module_path = str((Path.home() / relative_path).resolve())
-                sys.path.append(module_path)
-                logger.info(f"Appending {module_path} to sys.path")
 
             if module_name == "notebook":
+                # If relative_path is None, the module is not in the working dir, and should be in the reqs
```
carolineechen, line 175: included this here to maintain the same behavior for notebooks, but not 100% sure if this is necessary?

dongreenberg: It's not, we can remove it. If `module_name == "notebook"`(a bit hacky right now, we should clean it up) we're just serializing the function with cloudpickle and sending it over to the cluster, which is deserializing it (in the same line above where it's deserializing the path, fn_name, etc.) and running it. However this isn't even true for all notebooks, the default behavior is actually to save down the function as a module and the user must opt-in to have their function serialized in this way (because cloudpickle is very demanding about the minor versions of libraries matching in the serialization and deserializing environments). So we should probably rename "notebook" to "serialized" or something. Also worth noting that if the function is serialized the relative path is None, so you've already handled it in get_fn_by_name correctly.

dongreenberg (approved):
Looks great!!

PR #14, opened 2023-03-01
Handle folder resource on local cluster

carolineechen:
Automatically set the `system` attribute of a Folder resource to the cluster it is on, rather than needing to specify `folder.to(system).from_cluster(system)`.
- save cluster config onto itself, to be able to load and detect the cluster a resource is on
- remove `from_cluster` function
- remove `return_dest_folder`


runhouse/rns/blob.py
```diff
@@ -228,23 +216,31 @@ def blob(
...
-        # By default we save the blob in its own folder
         blob_name_in_path = (
             f"{generate_uuid()}/{rns_client.resolve_rns_data_resource_name(name)}"
         )
-        if config["system"] == rns_client.DEFAULT_FS:
-            # create random path to store in .cache directory of local filesystem
+
+        if (
+            system == rns_client.DEFAULT_FS
+            or config["system"] == _current_cluster(key="name")
```
carolineechen, line 237: if get a str name for system, can expand rns_address and compare against that

runhouse/rns/folders/folder.py
```diff
@@ -208,15 +224,17 @@ def data_config(self, data_config):
     @property
     def _fs_str(self):
         if isinstance(self.system, Resource):  # if system is a cluster
-            # TODO [DG] Return 'file' if we're on this cluster
+            if self.system.rns_address == _current_cluster("rns_address"):
```
carolineechen, line 227: Is there an easy way to add a test for this case? would require having a folder on a cluster and testing from within the cluster that it returns "file" and not "ssh"

dongreenberg: Yes, you can create a folder with its system set to a cluster, then define a function which accepts a folder object and returns it's _fs_str, then send that function to the cluster and call it on your folder object, you should get back "file" and not "ssh" (but get back "ssh" if you run the function locally). Does that make sense?

runhouse/rns/folders/folder.py
```diff
@@ -208,15 +224,17 @@ def data_config(self, data_config):
...
         else:
             return self.system
 
     @property
     def fsspec_fs(self):
-        if self._fsspec_fs is None:
-            self._fsspec_fs = fsspec.filesystem(self._fs_str, **self.data_config)
+        # recompute this each time, since it is different depending if it is being called
+        # from the same cluster (local filesystem) or somewhere else (ssh filesystem)
+        self._fsspec_fs = fsspec.filesystem(self._fs_str, **self.data_config)
```
dongreenberg, line 237: Can you tell if this affects performance? Also, if we're not caching it, we can just return `fsspec.filesystem(self._fs_str, **self.data_config)`, no need to store it in _fsspec_fs I think (unless you see otherwise).

dongreenberg (approved):
LGTM!!

runhouse/rns/folders/folder.py
carolineechen: I'm not actually too sure the performance impact, and if implemented this way the yea we can get rid of `_fsspec_fs`. Hm not completely sure this is bug proof though, if loading `**self.data_config` from a different file system than it was created on might have issues? (running into a similar issue currently with BERT FT on kitchen branch)

PR #17, opened 2023-03-19
Writing, loading, and saving resources

jlewitt1:
Separate the functionality of writing/saving underlying resource data from the metadata saved to the Runhouse Resource Naming System (RNS), used to facilitate collaboration and portability of resources between environments and users:

- use `write` method for saving underlying resource data
- use `fetch` or `stream` method for loading underlying resource data 
- use `save` method for saving resource metadata to RNS 
- use `from_name` method for loading a resource saved to RNS
- Add `load` param to each resource factory method to indicate whether we try to load an existing config from RNS


runhouse/rns/hardware/cluster_factory.py
```diff
@@ -37,7 +38,8 @@ def cluster(
         ips (List[str], optional): List of IP addresses for the BYO cluster.
         ssh_creds (dict, optional): Dictionary mapping SSH credentials.
             Example: ``ssh_creds={'ssh_user': '...', 'ssh_private_key':'<path_to_key>'}``
-        dryrun:
+        dryrun (bool): Whether or not to save the cluster. (Default: ``False``)
```
carolineechen, line 41: I know this is copied from blob but the wording is a bit confusing and I think mixes up the behavior for dryrun, could you make it clearer for both?

runhouse/rns/blob.py
```diff
@@ -125,11 +125,7 @@ def _save_sub_resources(self):
         if isinstance(self.system, Resource):
             self.system.save()
 
-    def save(
-        self,
-        name: str = None,
-        overwrite: bool = True,
-    ):
+    def write(self):
         """Save the blob to RNS."""
```
carolineechen, line 129: should this be something like "write down blob data to file"

runhouse/rns/tables/table.py
```diff
@@ -172,12 +172,8 @@ def _save_sub_resources(self):
...
             self.system.save()
 
-    def save(
-        self,
-        name: Optional[str] = None,
-        overwrite: bool = True,
-    ):
-        """Save the table to RNS."""
+    def write(self):
+        """Write table to fsspec url. Optionally save down the table's metadata to RNS (defaults to `True`)."""
```
carolineechen, line 176: There's no option to save metadata here

tests/test_folder.py
```diff
@@ -336,11 +336,13 @@ def test_cluster_and_cluster():
 def test_s3_sharing():
     s3_folder = rh.folder(
         name="my-s3-shared-folder", path=DATA_STORE_PATH, system="s3"
-    ).save()
+    ).save_rh()
```
carolineechen, line 339: there's still quite a few references to `save_rh` in tests + readme btw

PR #26, opened 2023-03-22
add `resources` property for Function object to pass into ray.remote

jlewitt1:
_Current behavior_: By default, Runhouse passes in fractional resources `{"num_cpus": 0.0001, "num_gpus": 0.0001}` when calling `ray.remote()` on a particular Function. This results in the scheduling and queueing not always working properly,  since we are basically telling Ray that each job doesn't need any resources.  

This change gives users control over the resources allocated to a particular Function, by adding a`resource` parameter to the `Function` object. Users can specify `num_cpus` (number of CPU cores to reserve for the task) or `num_gpus` (number of GPUs to reserve).

### DIFF
--- a/runhouse/rns/function.py
+++ b/runhouse/rns/function.py
@@ -24,8 +24,6 @@ logger = logging.getLogger(__name__)
 class Function(Resource):
     RESOURCE_TYPE = "function"
     DEFAULT_ACCESS = "write"
-    # By specifying fractional resources we remove queueing when calling ray remote on a function
-    DEFAULT_RESOURCES = {"num_cpus": 0.0001, "num_gpus": 0.0001}
 
     def __init__(
         self,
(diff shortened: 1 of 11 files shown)


runhouse/rns/function.py
```diff
@@ -49,6 +52,7 @@ def __init__(
         self.setup_cmds = setup_cmds or []
         self.access = access or self.DEFAULT_ACCESS
         self.dryrun = dryrun
+        self.resources = resources or self.DEFAULT_RESOURCES
```
dongreenberg, line 55: I'd actually rather leave this as None client-side by default. If we change the default resource behavior (I think we will) any functions which are saved with these resources will break.

dongreenberg (commented):
Looks excellent so far. For tests, it would be nice to test the queuing behavior that motivated this feature - i.e. submitting multiple calls to a cluster and wanting to make sure they run sequentially (not with contention). Two other things:
1) Right now we have an "enqueue" call type that doesn't really do its job (my fault). Could be nice to let that accept a `resources` arg to force queuing. It would do the same thing as setting resources in the fn constructor, but let's you do it one-off for one call without changing it for the whole function definition.
1) Sometimes the ray cluster gets incorrectly locked up with "resources unavailable," would be nice to have a way to flush out the pending jobs without restarting the grpc server and ray. Maybe something like `cluster.cancel(all=True)` (reusing the existing cancel fn).

PR #36, opened 2023-03-24
add tqdm progress bar on server side for map and starmap

jlewitt1:


dongreenberg:
I think someone on the Ray team was working on an experimental ray.tqdm progress bar because this may impact performance. Might be worth doing a quick speed test with and without.

PR #41, opened 2023-03-31
Torch version installation on cluster

jlewitt1:
Ensure the correct torch version is installed on the cluster, based on its cuda version.

### DIFF
--- a/tests/test_package.py
+++ b/tests/test_package.py
@@ -64,10 +64,31 @@ def test_load_shared_git_package():
     assert git_package.config_for_rns
 
 
-def test_cuda_version_install_for_torch():
-    pkg = rh.Package.from_string(specifier="pip:torch==1.13.1")
-    pkg.install()
-    assert pkg.install_target == "torch==1.13.1"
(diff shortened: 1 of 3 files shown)


dongreenberg:
I would do this inside the package class (probably in the install method itself, not in the instantiation), not inside the server. Also, I would use a `nvcc --version` to get the version instead of an existing PyTorch installation.

tests/test_package.py
```diff
@@ -60,10 +60,16 @@ def test_share_git_package():
 
 
 def test_load_shared_git_package():
-    git_package = rh.package(name="@/shared_git_package")
+    git_package = rh.Package.from_name(name="@/shared_git_package")
     assert git_package.config_for_rns
 
 
+def test_cuda_version_install_for_torch():
```
dongreenberg, line 67: This is a case where I think we need to unit test a longer list of install_cmds and the expected command that would be run. There are a bunch of ways users might specify the install string that we already know of. I wouldn't call install, just test the fn which generates the command.

PR #49, opened 2023-04-09
Enable cross cluster resource sharing

jlewitt1:
By default sharing resources between clusters is not enabled. 

If we want to stream data saved on one cluster (c1) from another cluster (c2), the sky keys + sky cluster yaml from c1 must be present on c2 in order to facilitate this sharing. 

By calling `c2.send_secrets(['sky'])` we load up the user's sky keys + cluster yaml configs for other upped sky clusters and save them to path: `~/.sky/generated` on c2.

### DIFF
--- a/runhouse/rns/hardware/cluster.py
+++ b/runhouse/rns/hardware/cluster.py
@@ -559,8 +559,15 @@ class Cluster(Resource):
         return return_codes
 
     def send_secrets(self, providers: Optional[List[str]] = None):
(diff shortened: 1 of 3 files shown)


jlewitt1:
@dongreenberg not sure if at one point we had this automated (maybe in a previous version of sky)? but for now we need the cluster yaml where the data lives to be present in `~/.sky/generated` in the file system of the cluster trying to load/stream that data

PR #51, opened 2023-04-11
Update README

dongreenberg:


README.md
```diff
@@ -14,17 +14,29 @@
...
+interpreter that lets it take detours to remote machines or hold and manipulate remote
+data. So your preprocessing can briefly detour to run on a 64 CPU GCP instance, then
+return to your local environment, then jump to a GPU cluster on Lambda Labs for training.
+The dataframe your preprocessing generates can stream directly to the GPU cluster, so it
+doesn't need to bounce off your laptop. Normally you'd need to translate your code into
+an orchestrator workflow to do all this, but now you can run, test, iterate, and share this program
+like any other Python, even in notebooks. You can drop it into an orchestrator node just
+like any other Python script to schedule or monitor it, or use it with any experiment
+management and lineage tools you like. Runhouse heavily builds on top of Ray, gRPC, and
+SkyPilot, so lots of nice things like queuing, distributed, production-hardness, async,
```
carolineechen, line 30:
```suggestion
SkyPilot, so lots of nice features like queuing, distributed, production-hardness, async,
```

README.md
```diff
@@ -14,17 +14,29 @@
...
-eagerly exactly as they were.
+data infra (with your own cloud creds), all in Python, and continue to interact with
+them normally from your local environment.
+
+**Runhouse is not an orchestrator.** Think of Runhouse as an expansion pack to your Python
+interpreter that lets it take detours to remote machines or hold and manipulate remote
+data. So your preprocessing can briefly detour to run on a 64 CPU GCP instance, then
+return to your local environment, then jump to a GPU cluster on Lambda Labs for training.
+The dataframe your preprocessing generates can stream directly to the GPU cluster, so it
+doesn't need to bounce off your laptop. Normally you'd need to translate your code into
```
carolineechen, line 25:
```suggestion
The preprocessed dataframe can stream directly to the GPU cluster, without bouncing off your laptop. Normally you'd need to translate your code into
```

README.md
```diff
@@ -38,70 +50,89 @@ Take a look at this code (adapted from our first [tutorial](https://github.com/r
...
+         f"--output_dir=dreambooth/output "])
 
 # This goes directly cluster-> s3, doesn't bounce to local
 outputs_s3 = rh.folder(system=gpu, path="dreambooth/outputs").to("s3", path="runhouse/dreambooth/outputs")
 outputs_s3.save("dreambooth_outputs")
+```
 
-# and later:
-rh.folder(name="dreambooth_outputs").to("here")
+And later, if I want to load down the folder in full:
```
carolineechen, line 92:
```suggestion
To load down the folder in full later:
```

README.md
```diff
@@ -38,70 +50,89 @@ Take a look at this code (adapted from our first [tutorial](https://github.com/r
...
+```python
+rh.Folder.from_name("dreambooth_outputs").to("here")
+```
 
-# Load a table in from anywhere (S3, GCS, local, etc)
-my_table = rh.table(system="gcs", path="my_bucket/my_table.parquet").to("here")
-# preprocess...
+We support interactive blob and table primitives too:
```
carolineechen, line 97: The folder and function examples are simple and great, but this one is a bit harder to digest (for a README). I think it's borderline so we can leave it for now if you feel strongly, or we could just saying that we support blob and table primitives, between file systems, clusters, and local without the full code demo. (maybe eventually move this to a quickstart guide page instead of readme?)

README.md
```diff
@@ -14,17 +14,29 @@
 PyTorch lets you send a model or tensor `.to(device)`, so
 why can't you do `my_fn.to('a_gcp_a100')` or `my_table.to('parquet_in_s3')`?
 Runhouse allows just that: send code and data to any of your compute or
-data infra (with your own cloud creds), all in Python, and continue to use them
-eagerly exactly as they were.
+data infra (with your own cloud creds), all in Python, and continue to interact with
+them normally from your local environment.
+
+**Runhouse is not an orchestrator.** Think of Runhouse as an expansion pack to your Python
```
carolineechen, line 20: if this could be simplified in the future or moved down, it's a lot of text at the beginning of the page

dongreenberg: I agree. It feels like we need to add it but I'd like to simplify and spread it around below shortly.

README.md
dongreenberg: I tried to make it into a little bit more of a real-looking application, but I agree it could be more digestible. I'm going to push it for now and then prune it shortly.

PR #53, opened 2023-04-13
Update interactive login/logout flow

### LINKED ISSUE #52: Consider adding `-y` option to `runhouse login` CLI command
## Feature
Simple use case is logging in with `system` command instead of Python API:
```
!runhouse login [TOKEN]
```
Currently, the CLI is hardcoded with `interactive=True`:
https://github.com/run-house/runhouse/blob/560a52880a333e17e8a1aca01c1048f4527fc375/runhouse/main.py#L27

## Motivation
It's a minor quality of life improvement.

## Ideal Solution
See above

## Additional context
Excited to get Runhouse integration up on NatML 😄

carolineechen:
Resolves #52 

* add `--yes` or `-y` option to login and logout cli commands, which sets any upload/download/delete confirmations to yes (token must be provided for truly non-interactive flow, otherwise token will be requested if not provided)
* setting `interactive=False` from Python will throw an error if not all required arguments (token, confirmations) are provided
* if all arguments are provided and `interactive` not explicitly set to `True`, default behavior is to skip interactive steps


runhouse/rns/login.py
```diff
@@ -21,14 +21,23 @@ def login(
...
+    all_options_set = token and not any(
+        arg is None
+        for arg in (download_config, upload_config, download_secrets, upload_secrets)
+    )
+    if interactive is False and not all_options_set:
+        raise Exception(
+            "`interactive` and only be set to `False` if token and all download and upload options are provided."
+        )
+
+    if interactive or (is_interactive() and not (all_options_set)):
```
jlewitt1, line 40:
```suggestion
    if interactive or (is_interactive() and not all_options_set):
```

runhouse/rns/login.py
```diff
@@ -21,14 +21,23 @@ def login(
...
     from runhouse import Secrets
 
-    if is_interactive() or interactive:
+    all_options_set = token and not any(
+        arg is None
+        for arg in (download_config, upload_config, download_secrets, upload_secrets)
+    )
+    if interactive is False and not all_options_set:
+        raise Exception(
+            "`interactive` and only be set to `False` if token and all download and upload options are provided."
```
jlewitt1, line 37:
```suggestion
            "`interactive` can only be set to `False` if token and all download and upload options are provided."
```

jlewitt1:
It looks like if we call `runhouse login -y` from a fresh env without passing in a token we get back a messy traceback from RNS - can we handle that scenario more gracefully?

jlewitt1 (approved):

Issue #45, opened 2023-04-04
[KIT-67] Runs


dongreenberg:
Basic API ideas (WIP):

Create Run object (captures logs, inputs, outputs, other artifacts read or written within call, who ran, where):

`res = fn(**kwargs, name=”my_run”)`

A run is a folder (created inside *local* rh directory by default), and can be sent elsewhere to persist logs, results, artifact info, etc.:

`rh.run(name=“my_run”).to("s3", path="runhouse/nlp_team/bert_ft/results")`

Ideally, we can have a "default log store" setting in the user config so the logs from their runs can be sent to the same place by default when they save, rather than having to send each run one by one.

This could be the way for users to configure for artifacts/logs to flow to an existing MLFlow store, or to flow to W&B, Grafana, Datadog, etc.

Save the run to local or RNS (not all runs need to be saved)

`rh.run(name=“my_run”).save()` 

Creates a run object by tracing the activity within the block - no inputs and outputs, but captures logs (perhaps several logfiles for different calls) and artifacts used:

`with rh.run(name=”my_run”) as r:`

Big feature, essentially the same as auto-caching in orchestrators - check if this run was already completed, and load results if so, otherwise run:

`res = fn.get_or_run(name=”yelp_review_preproc_test”)`

Create/name a CLI run:

`r = my_cluster.run(["python test_bert.py --gpus 4 --model distilbert"], name="test_distilbert_ddp")`

Inspiration: [this MLFlow example](https://github.com/mlflow/mlflow/blob/master/examples/multistep_workflow/main.py#L91)

We can also support event (failure or completion) notifications through [knocknock](https://github.com/huggingface/knockknock) or pagerduty!

Cc @caroline 

<sub>From [SyncLinear.com](https://synclinear.com) | [KIT-67](https://linear.app/runhouse/issue/KIT-67/runs)</sub>

dongreenberg:
Interesting approach using Ray and OpenTelemetry to look into: [https://composable-logs.github.io/composable-logs/home/](https://composable-logs.github.io/composable-logs/home/)

jlewitt1:
Some more context based on the above: 

3 main ways of invoking/creating a run: 
1. Function call 
2. CLI command
3. Context manager

_Logs_
- The `.rh` directory on a cluster will contain logs for each run, with each run having its own unique key (user should be able to overwrite that key name with their own)
- Run object should hold a folder, which contains references to the log files for that run that were generated on the cluster.  
[...]

PR #55, opened 2023-04-16
Runhouse Run object

jlewitt1:
Based on: https://github.com/run-house/runhouse/issues/45


runhouse/rns/function.py
```diff
@@ -55,9 +56,10 @@ def __init__(
         self.access = access or self.DEFAULT_ACCESS
         self.dryrun = dryrun
         self.resources = resources or {}
+
         super().__init__(name=name, dryrun=dryrun)
 
-        self = self.to(self.system, env=self.env)
```
carolineechen, line 60: `.to` returns a new object, so I think in this case we do want to keep the `self = `?

runhouse/rns/function.py
```diff
@@ -114,8 +118,8 @@ def to(
             env = env or self.env
             env = _get_env_from(env)
 
-        if self.env:
-            new_reqs = [req for req in self.env.reqs if req not in env.reqs]
+        if env:
```
carolineechen, line 121: the reason this should be self.env is because in the function constructor (which supports setting the reqs/env), we append `"./"` package to the `self.env.reqs` if not included to handle syncing the local package, and additionally add it here if it's not in the `env` parameter for the `.to` function. It's not the most intuitive, and with potentially getting rid of setting env/system in the function factory this could probably be simplified -- let's discuss more live

runhouse/rns/run.py
```diff
@@ -0,0 +1,557 @@
...
+    def base_cluster_folder_path(name: str):
+        """Path to the base folder for this Run on a cluster."""
+        return f"~/{obj_store.LOGS_DIR}/{name}"
+
+    @staticmethod
+    def current_timestamp():
+        return str(log_timestamp())
+
+    @classmethod
+    def create_run_name(cls, name_run: Union[str, bool], fn_name: str = None) -> str:
```
carolineechen, line 419: should this be a private method?

runhouse/rns/folders/folder.py
```diff
@@ -744,10 +745,10 @@ def exists_in_system(self):
             self.fsspec_url
         ) or rh.rns.top_level_rns_fns.exists(self.path)
 
-    def delete_in_system(self):
-        """Delete all contents in folder from file system."""
+    def delete_in_system(self, path: Optional[str] = None):
```
carolineechen, line 748: would just `delete` make sense as the function name?

runhouse/rns/run.py
```diff
@@ -0,0 +1,560 @@
...
+
+        new_run._folder = self._folder.to(
+            system=system, path=path, data_config=data_config
+        )
+
+        return new_run
+
+    def inputs(self) -> bytes:
+        """Read the pickled function inputs saved on the system for the Run.
+        Note: It's the user's responsibility to unpickle the inputs."""
```
carolineechen, line 241:
```suggestion
        
        Note: It's the user's responsibility to unpickle the inputs."""
```

runhouse/rns/run.py
```diff
@@ -0,0 +1,560 @@
...
+        """Delete a Run from its system. Deletes the specified folder and all its contents from the Run's file system"""
+        path = path or self.path
+        self._folder.delete_in_system(path)
+
+    def exists_in_system(self, path: str = None) -> bool:
+        """Checks whether a path exists in the Run's file system.
+        Optionally provide a custom path to check, otherwise will default to the Run's folder path."""
+        return self._folder.fsspec_fs.exists(path or self.path)
+
+    def register_new_fn_run(self):
```
carolineechen, line 295: I think these register functions should be private?

runhouse/rns/folders/folder.py
```diff
@@ -744,10 +745,10 @@ def exists_in_system(self):
...
+        """Delete all contents in folder from file system. Note: will remove the folder and all its contents."""
```
carolineechen, line 749:
```suggestion
        """Delete all contents in folder from file system.
        
        Note: will remove the folder and all its contents."""
```

runhouse/rns/folders/folder.py
```diff
@@ -758,14 +759,16 @@ def rm(self, name, recursive: bool = True):
         except FileNotFoundError:
             pass
 
-    def put(self, contents, overwrite=False):
+    def put(self, contents, overwrite=False, as_json=False):
         """Put given contents in folder. Contents must be one of the following:
 
-        - Dict with keys being the file names and values being the file-like objects to write
+        - Dict with keys being the file names (without full paths) and values being the file-like objects to write
```
carolineechen, line 765: format as docstrings instead of fn def

runhouse/rns/function.py
```diff
@@ -372,18 +396,32 @@ def enqueue(self, resources: Optional[Dict] = None, *args, **kwargs):
...
-        # ray.init(ignore_reinit_error=True)
+    def run(self, name_run: Union[str, bool] = None, *args, **kwargs):
+        """Run async remote call on cluster.
+
+        Args:
+            name_run (Union[str, bool]): Name of the Run to create. If ``True``, a name will automatically
+             be generated.
+             *args: Optional args for the Function
+             **kwargs: Optional kwargs for the Function
+        Returns:
```
carolineechen, line 407:
```suggestion
        
        Returns:
```

runhouse/rns/function.py
```diff
@@ -392,14 +430,14 @@ def remote(self, *args, **kwargs):
             # TODO print this nicely
             logger.info(
                 f"Submitted remote call to cluster. Result or logs can be retrieved"
-                f'\n with run_key "{run_key}", e.g. '
-                f'\n`{cluster_name}.get("{run_key}", stream_logs=True)` in python '
-                f'\n`runhouse logs "{self.system.name}" {run_key}` from the command line.'
+                f'\n with run name "{run_name}", e.g. '
+                f'\n`{cluster_name}.get("{run_name}", stream_logs=True)` in python '
```
carolineechen, line 434: this reads kinda clunky and outputs a large chunk of text, we should have this somewhere in docs instead of listing out 4 separate examples of using the run_name

runhouse/rns/function.py
```diff
@@ -584,6 +614,56 @@ def notebook(self, persist=False, sync_package_on_close=None, port_forward=8888)
...
+
+        res = self._call_fn_with_ssh_access(
+            fn_type="get_or_call", run_name=name_run, args=args, kwargs=kwargs
+        )
+
+        return res
+
+    def get_or_run(self, name_run: str = None, *args, **kwargs) -> "Run":
+        """Check if Run was already completed. If no cached Run is found on the cluster, create a new one.
+        Note: If the Run has already completed, will not trigger a new Run.
```
carolineechen, line 642:
```suggestion
        
        Note: If the Run has already completed, will not trigger a new Run.
```

runhouse/rns/run.py
```diff
@@ -0,0 +1,560 @@
...
+    LOCAL_RUN_PATH = f"{rns_client.rh_directory}/runs"
+    RUN_CONFIG_FILE = "config_for_rns.json"
+
+    NOT_STARTED_STATUS = "NOT_STARTED"
+    RUNNING_STATUS = "RUNNING"
+    COMPLETED_STATUS = "COMPLETED"
+    ERROR_STATUS = "ERROR"
+
+    # Supported Run types
+    CLI_RUN = "CLI"
```
carolineechen, line 35: would `CMD_RUN = "CMD"` make a little more sense, since we support both run_python and run, and also the differentiating factor is the `cmds` param passed into the function

runhouse/rns/run.py
```diff
@@ -0,0 +1,560 @@
...
+    def base_local_folder_path(self):
+        """Path to the base folder for this Run on a local system."""
+        return f"{self.LOCAL_RUN_PATH}/{self.name}"
+
+    def fn_inputs_path(self):
+        """Path the pickled inputs used for the function which are saved on the system."""
+        return f"{self.path}/inputs.pkl"
+
+    def fn_result_path(self):
+        """Path the pickled result for the function which are saved on the system."""
```
carolineechen, line 397:
```suggestion
        """Path to the pickled result for the function which are saved on the system."""
```

runhouse/rns/run.py
```diff
@@ -0,0 +1,560 @@
...
+        return f"{self.path}/result.pkl"
+
+    @staticmethod
+    def base_folder_name(name: str = None):
```
carolineechen, line 401: private method? same with `base_cluster_folder_path`

runhouse/rns/run.py
```diff
@@ -0,0 +1,560 @@
...
+    name: Optional[str] = None,
+    fn: Optional[Union[str, Callable]] = None,
+    cmds: Optional[List] = None,
+    dryrun: bool = False,
+    load: bool = True,
+    overwrite: bool = False,
+    **kwargs,
+):
+    """Returns a Run object, which can be used to capture logs, inputs and results, artifact info, etc.
+    Currently supports running a particular Runhouse Function object or CLI command(s)
```
carolineechen, line 515:
```suggestion
    Currently supports running a particular Runhouse Function object, or CLI or Python command(s)
```

runhouse/rns/run.py
```diff
@@ -0,0 +1,560 @@
...
+    on a specified system (ex: on a cluster).
+
+    Args:
+        name (Optional[str]): Name to give the Run object, to be reused later on.
+        fn (Optional[str or Function]): The function to execute on the remote system when the function is called.
+            Can be provided as a function name or as a Runhouse ``Function`` object.
+        cmds (Optional[str]): List of CLI or Python commands to execute for the run.
```
carolineechen, line 522:
```suggestion
        cmds (Optional[List]): List of CLI or Python commands to execute for the run.
```

docs/api/run.rst
```diff
@@ -0,0 +1,231 @@
...
+    my_run_on_s3 = my_run.to("s3", path="/s3-bucket/s3-folder")
+
+
+To check if a run was already completed, and load results if so:
+
+.. code-block:: python
+
+    import runhouse as rh
+
+    my_func = rh.Function.from_name("my_func")
```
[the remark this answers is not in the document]
jlewitt1: agreed I dont think we do - I figured we would address this separately as part of the https://github.com/run-house/runhouse/pull/67 PR

runhouse/rns/folders/folder.py
jlewitt1: i was following the convention we used for other resources (blob, folder, table, etc.).. do you think we should rename all of them for simplicity?

runhouse/rns/run.py
jlewitt1: we are calling these functions as part of an RPC or http endpoint when we create a new Run object on the server

carolineechen: private as in we prob don't expect users to touch this and can preface it with an underscore?

runhouse/rns/envs/env.py
```diff
@@ -70,7 +70,7 @@ def _reqs_to(self, system: Union[str, Cluster], path=None, mount=False):
             new_reqs.append(req)
         return new_reqs
 
-    def _setup_env(self, system: Union[str, Cluster]):
+    def _setup_env(self, system: Cluster):
```
dongreenberg, line 73: I don't think this is actually the case - a string would fail here.

runhouse/rns/folders/folder.py
```diff
@@ -758,14 +761,17 @@ def rm(self, name, recursive: bool = True):
...
+        """Put given contents in folder.
+
+        Args:
+            contents (Dict[str, Any] or Resource or List[Resource]): Contents to put in folder.
+                Must be a dict with keys being the file names (without full paths) and values being the file-like
+                objects to write, or a Resource object, or a list of Resources.
+            overwrite (bool): Whether to dump the file contents as json. By default expects data to be encoded.
+                Defaults to ``False``.
+            as_json (bool): Name of the Run to create. If ``True``, a Run will be created
+                Defaults to ``False``.
```
dongreenberg, line 774: 1) I think this was copy pasted
2) I don't think we should handle serialization here, I think it's too automagic. Json serialization fails often and users will blame us, when really it's just debugging they need to do to make sure their object can be written down properly. If there's a big convenience to not needing to convert all the way to bytes before passing into .put, one thing we can do is support an argument like "write_fn" (which accepts a file and data) which can do the serialization, so the user can pass `write_fn=lambda f, data: json.dump(data, f)` which we then call inside put. This would mean the burden is still on them to handle the serialization, but it can also support any other serialization format. We don't need to accept args and kwargs into the write_fn because the user can just just functools `partial` to pre-pass those.

dongreenberg: Btw we should also accept "mode" as an arg into put, users want to be able to control that (just like we do here).

runhouse/rns/folders/folder.py
```diff
@@ -865,7 +875,7 @@ def folder(
     Args:
         name (Optional[str]): Name to give the folder, to be re-used later on.
         path (Optional[str or Path]): Path (or path) that the folder is located at.
-        system (Optional[str]): File system. Currently this must be one of:
+        system (Optional[str]): File system or cluster name. If providing a file system this must be one of:
```
dongreenberg, line 878: Nice catch

runhouse/rns/folders/azure_folder.py
```diff
@@ -15,7 +15,7 @@ def from_config(config: dict, dryrun=True):
         """Load config values into the object."""
         return AzureFolder(**config, dryrun=dryrun)
 
-    def delete_in_system(self):
+    def delete_in_system(self, path: Optional[str] = None):
```
dongreenberg, line 18: Same comment as below, I don't think we want to accept a random path here.

runhouse/rns/folders/gcs_folder.py
```diff
@@ -19,9 +19,9 @@ def from_config(config: dict, dryrun=True):
         """Load config values into the object."""
         return GCSFolder(**config, dryrun=dryrun)
 
-    def delete_in_system(self):
+    def delete_in_system(self, path: Optional[str] = None):
```
dongreenberg, line 22: Same comment as in folder, I don't think we want to accept a random path here.

runhouse/rns/folders/s3_folder.py
```diff
@@ -28,9 +28,9 @@ def from_config(config: dict, dryrun=True):
         """Load config values into the object."""
         return S3Folder(**config, dryrun=dryrun)
 
-    def delete_in_system(self):
+    def delete_in_system(self, path: Optional[str] = None):
```
dongreenberg, line 31: Same comment as in folder, I don't think we want to accept a random path here.

dongreenberg (commented):
First batch of comments, will continue at function.py

runhouse/rns/function.py
```diff
@@ -260,10 +260,33 @@ def _extract_fn_paths(raw_fn: Callable, reqs: List[str]):
 
     # ----------------- Function call methods -----------------
 
-    def __call__(self, *args, stream_logs=False, **kwargs):
+    def __call__(
+        self, *args, stream_logs=False, name_run: Union[str, bool] = None, **kwargs
```
dongreenberg, line 264: I would change stream_logs to true, I don't see a reason not to if everything goes through ray now anyway.

runhouse/rns/function.py
```diff
@@ -500,7 +551,8 @@ def config_for_rns(self):
         return config
 
     def _save_sub_resources(self):
-        self.system.save()
+        if isinstance(self.system, Resource):
```
dongreenberg, line 554: When is system not a resource? It should always be, no?

runhouse/rns/function.py
```diff
@@ -409,7 +444,21 @@ def get(self, run_key):
         """
         return self.system.get(run_key)
 
-    def _call_fn_with_ssh_access(self, fn_type, resources=None, args=None, kwargs=None):
+    def _create_async_run(
```
dongreenberg, line 447: Where is this used?

runhouse/rns/hardware/cluster.py
```diff
@@ -222,6 +223,16 @@ def get(self, key: str, default: Any = None, stream_logs: bool = False):
         self.check_server()
         return self.client.get_object(key, stream_logs=stream_logs) or default
 
+    def get_run(self, run_name: str, folder_path: str = None):
+        self.check_server()
+        return self.client.get_run_object(run_name, folder_path)
+
+    def run_cmds(
```
dongreenberg, line 230: How does this differ from cluster.run?

runhouse/rns/hardware/cluster.py
```diff
@@ -542,6 +556,29 @@ def run(
                 env = Env.from_name(env)
             cmd_prefix = env._run_cmd
 
+        if name_run:
```
dongreenberg, line 559: I'm kinda confused about all the different methods to run on the cluster. If the user doesn't pass name_run, the run happens through the runner like it does today, but if they do it goes over rpc?

runhouse/rns/rns_client.py
```diff
@@ -192,6 +193,36 @@ def resource_request_payload(self, payload) -> dict:
         payload["data"] = data
         return payload
 
+    # Run Stack
```
dongreenberg, line 196: This is great, super straightforward

dongreenberg (commented):
Another batch of comments, and still more to review

runhouse/rns/hardware/cluster.py
jlewitt1: we need a way to basically intercept the command before running it in order to create the Run object + config,  stdout, etc. on the cluster, so that's why we use an RPC here. Do you think we should always use an RPC instead of splitting the functionality like this?

runhouse/rns/run_module_utils.py
```diff
@@ -103,6 +138,27 @@ def deserialize_args_and_kwargs(args, kwargs):
     return args, kwargs
 
 
+def get_fn_by_name(module_name, fn_name, relative_path=None):
```
dongreenberg, line 141: How does this differ from `get_fn_from_pointers` below?

runhouse/rns/run_module_utils.py
```diff
@@ -134,21 +190,300 @@ def get_fn_from_pointers(fn_pointers, serialize_res, num_gpus, *args, **kwargs):
...
+    run_obj = run(
+        name=run_name, cmds=commands, overwrite=True, path=folder_path_on_system
+    )
+
+    run_obj._register_new_fn_run()
+
+    final_stdout = []
+    final_stderr = []
+
+    for command in commands:
```
dongreenberg, line 428: Downside of running via rpc here compared to with the CommandRunner is that we're not streaming logs back to the user.

runhouse/servers/http/http_server.py
```diff
@@ -185,6 +196,57 @@ def get_object(message: Message):
...
+        logger.info(
+            f"Message received from client to get run object: {b64_unpickle(message.data)}"
+        )
+        run_name, folder_path = b64_unpickle(message.data)
+
+        folder_path = folder_path or Run._base_cluster_folder_path(run_name)
+        folder_path_on_system = resolve_absolute_path(folder_path)
+
+        try:
+            HTTPServer.register_activity()
```
dongreenberg, line 215: We don't need this again so soon after calling it above, only if a reasonable amount of time elapses.

runhouse/servers/http/http_server.py
```diff
@@ -185,6 +196,57 @@ def get_object(message: Message):
...
+            # Load config data for this Run saved locally on the system
+            result = Run.from_path(path=folder_path_on_system)
+            return Response(
+                data=pickle_b64(result),
+                output_type=OutputType.RESULT,
+            )
+
+        except Exception as e:
+            HTTPServer.register_activity()
```
dongreenberg, line 224: Same note as above, not needed here

tests/test_run.py
```diff
@@ -0,0 +1,404 @@
...
+CLI_RUN_NAME = "my_cli_run"
+S3_BUCKET = "runhouse-runs"
+
+PATH_TO_CTX_MGR_RUN = f"{rh.Run.LOCAL_RUN_PATH}/{CTX_MGR_RUN}"
+
+
+def setup():
+    from runhouse.rns.api_utils.utils import create_s3_bucket
+
+    create_s3_bucket(S3_BUCKET)
```
dongreenberg, line 21: We should probably just do this with fixtures, no?

tests/test_run.py
```diff
@@ -0,0 +1,404 @@
...
+
+# ------------------------- FUNCTION RUN ----------------------------------
+
+
+@pytest.mark.clustertest
+def test_create_run_on_cluster(summer_func):
+    """Initializes a Run, which will run synchronously on the cluster.
+    Returns the function's result"""
+    res = summer_func(1, 2, name_run=FUNC_RUN_NAME)
+    assert res == 3
```
dongreenberg, line 32: All the tests below relying on this test means it's not really a test, but a fixture (and we generally don't want to have tests which need to be run before other tests because it becomes difficult to keep track of. We may have a few here and there but should avoid it). I would change this into a fixture which returns submitted_run_name and all of the below can use instead of using FUNC_RUN_NAME and relying on this test running first.

tests/test_run.py
```diff
@@ -0,0 +1,404 @@
...
+    # Note: In this test since we are providing a name of "latest", it should return the latest cached version
+    run_output = summer_func.get_or_call("latest")
+
+    assert run_output == 3
+
+
+@pytest.mark.clustertest
+def test_run_fn_async(summer_func):
+    # TODO [JL] Failed - "Error inside function remote: Only one live display may be active at once"
+    """Execute function async on the cluster. If a run already exists, do not re-run. Returns a Run object."""
```
dongreenberg, line 86: Same comment as above, I think this should be a fixture

tests/test_run.py
```diff
@@ -0,0 +1,404 @@
...
+    completion before returning the result."""
+    # Note: In this test since we are providing a name of "latest", it should return the latest cached version
+    run_output = summer_func.get_or_call("latest")
+
+    assert run_output == 3
+
+
+@pytest.mark.clustertest
+def test_run_fn_async(summer_func):
+    # TODO [JL] Failed - "Error inside function remote: Only one live display may be active at once"
```
dongreenberg, line 85: Is this still relevant?

tests/test_run.py
```diff
@@ -0,0 +1,404 @@
...
+@pytest.mark.clustertest
+@pytest.mark.rnstest
+def test_save_fn_run_to_rns(cpu_cluster):
+    """Saves run config to RNS"""
+    func_run = cpu_cluster.get_run(FUNC_RUN_NAME)
+    assert func_run
+
+    func_run.save(name=FUNC_RUN_NAME)
+    loaded_run = rh.Run.from_name(name=FUNC_RUN_NAME)
+    assert rh.exists(loaded_run.name, resource_type=rh.Run.RESOURCE_TYPE)
```
dongreenberg, line 166: Also not testing a ton about the correctness here

dongreenberg (commented):
Amazing PR. We should talk about the run_module stuff more because I think it can be consolidated more and has a few gotchas

tests/test_run.py
```diff
@@ -0,0 +1,455 @@
...
+
+@pytest.mark.clustertest
+@pytest.mark.runstest
+def test_invalid_fn_async_run(summer_func):
+    """Test error handling for invalid function Run. The function expects to receive integers but
+    does not receive any. No result should be saved down on the cluster for this Run, and the Run object returned
+    should have a status of `ERROR`."""
+    run_obj = summer_func.get_or_run(name_run="invalid_async_run")
+
+    assert run_obj.refresh().status == rh.RunStatus.ERROR
```
dongreenberg, line 90: Would be good to grab the exception itself here and check it, like the test above.

dongreenberg (approved):
This is a marvel of a PR. It's makes a huge step forward in functionality and touches a lot of core code, but doesn't make a salad and is super crisp. Great stuff!!

runhouse/servers/http/http_server.py
```diff
@@ -127,34 +127,50 @@ def run_module(message: Message):
...
+
+            if obj_ref is not None:
                 obj_store.put_obj_ref(key=run_key, obj_ref=obj_ref)
-                result = pickle.dumps(run_key)
+
+            result = pickle.dumps(res) if isinstance(res, Run) else res
 
             HTTPServer.register_activity()
-            if isinstance(result, list):
+            if isinstance(result, ray.exceptions.RayTaskError):
```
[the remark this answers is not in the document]
jlewitt1: any other errors thrown would be caught and raised by the http server instead of returning the "result" object. For runs specifically we don't want to immediately raise an exception since we need to capture failed runs (and associated config, stderr, etc.) on the cluster's file system

PR #59, opened 2023-04-23
[WIP] Add test fixtures

jlewitt1:
Introduces pytest fixtures to provide more control over the tests requiring an upped cluster

- Add `cluster_tests.yaml` to GHA: runs all tests marked as `clustertest`
- Add `conftest.py` file to hold all test fixture objects (ex: clusters, tables, blob data) 
- Updated torch install tests - separate tests for detecting the CUDA versions on the cluster and validating the full install commands on the cluster


tests/conftest.py
```diff
@@ -0,0 +1,111 @@
...
+def cpu():
+    return rh.cluster("^rh-cpu").up_if_not()
+
+
+@pytest.fixture
+def v100():
+    # Note: This configuration may not be supported with AWS - if so may need to request 8 GPUS:
+    # return rh.cluster("^rh-v100").up_if_not()
+    return rh.cluster(
+        name="rh-v100", instance_type="V100:8", provider="aws"
```
dongreenberg, line 94: 8 v100's is a lot, and this makes it look like a single v100. We probably just want 1 here? Also, do we need to specify AWS here or should we just take cheapest? I guess it would be more reproducible to pick a single provider, but less comprehensive that we're never testing on other providers.

tests/conftest.py
```diff
@@ -0,0 +1,111 @@
...
+def ray_table():
+    import ray
+
+    ds = ray.data.range(10000)
+    return ds
+
+
+# ----------------- Clusters -----------------
+@pytest.fixture
+def cpu():
```
dongreenberg, line 85: I think cpu_cluster would be clearer (ditto gpu clusters below)

tests/test_cluster.py
```diff
@@ -46,39 +47,38 @@ def test_install():
...
-def test_on_same_cluster():
-    hw = cluster(name="^rh-cpu").up_if_not()
-    hw.restart_grpc_server()
-    hw.up_if_not()
-
-    hw_copy = cluster(name="^rh-cpu")
+@pytest.mark.clustertest
+def test_on_same_cluster(cpu):
+    hw_copy = cpu.copy()
+    cpu.restart_grpc_server()
```
dongreenberg, line 69: Why do we restart here?

tests/test_cluster.py
```diff
@@ -46,39 +47,38 @@ def test_install():
...
+    func_hw = rh.function(is_on_cluster).to(cpu)
+    assert func_hw(cpu)
     assert func_hw(hw_copy)
 
 
-def test_on_diff_cluster():
-    hw = cluster(name="^rh-cpu").up_if_not()
-    diff_hw = rh.cluster(name="test-byo-cluster").up_if_not()
+@pytest.mark.clustertest
+def test_on_diff_cluster(cpu, a10g):
```
dongreenberg, line 78: We don't need an a10 for this, could we add a fixture for another cpu cluster (e.g. the one we use in the byo tests with a different boot image)? We often need to use two in tests anyway.

tests/test_package.py
```diff
@@ -1,20 +1,33 @@
...
+cuda_116_url = "--index-url https://download.pytorch.org/whl/cu116"
+
 
 def setup():
     pass
 
 
+@pytest.fixture
+def cluster(request):
+    return request.getfixturevalue(request.param)
```
dongreenberg, line 18: What does this do?

tests/test_package.py
```diff
@@ -59,45 +73,188 @@ def test_share_git_package():
...
-                                                                         f"torchaudio {extra_index_url} "
-                                                                         f"{cuda_v116_url}",
-        ("torch torchpudding", "11.7"): f"torch {cuda_v117_url} {extra_index_url} torchpudding {cuda_v117_url} "
-                                        f"{extra_index_url}",
-        ("torch>=1.13.0", "11.6"): f"torch>=1.13.0 {cuda_v116_url} {extra_index_url}",
-        ("torch>1.13.0", "11.6"): f"torch>1.13.0 {cuda_v116_url} {extra_index_url}",
-        ("torch==1.13.0", "11.6"): f"torch==1.13.0 {cuda_v116_url} {extra_index_url}",
-        ("torch~=1.13.0", "11.7"): f"torch~=1.13.0 {cuda_v117_url} {extra_index_url}",
-        (f"torch==99.99.999 {cuda_v117_url}", "11.7"): f"torch==99.99.999 {cuda_v117_url} {extra_index_url}",
+        (
```
dongreenberg, line 123: This is just a silly formatting nit, but this doesn't need to be a dict, it can just be a list of lists and then it'd probably be a lot more readable

tests/test_package.py
```diff
@@ -59,45 +73,188 @@ def test_share_git_package():
...
+        # Run the complete install command on the cluster
+        pip_install_cmd = cluster.run_python(
+            [
+                "import runhouse as rh",
+                f"rh.Package.pip_install(install_cmd='{install_cmd}')",
+            ]
+        )
+        assert (
+            "ERROR" not in pip_install_cmd[0][1]
+        ), f"Failed to install command on {cluster.name}"
```
dongreenberg, line 242: Don't we just want to do cluster.install_packages(["torch"]) for each of these to see if the installation works?

tests/test_package.py
```diff
@@ -59,45 +73,188 @@ def test_share_git_package():
...
+
+        # Send a tensor to CUDA using this torch version on the cluster
+        torch_cmds = cluster.run_python(
+            [
+                "import torch",
+                "a = torch.LongTensor(1).random_(0, 10)",
+                "a = a.to(device='cuda')",
+                "print(a)",
+            ]
+        )
```
dongreenberg, line 252: I think it's probably more representative to put this in a function and send it to the cluster than run from the CLI.

dongreenberg (commented):
This is sick, it's a solid foundation for continuing to flesh out the test suite

tests/test_package.py
jlewitt1: this lets us parametrize over multiple fixtures - useful for running the same test to multiple hardware types. for example: `@pytest.mark.parametrize("cluster", ["v100", "k80", "a10g"], indirect=True)`

dongreenberg (approved):

PR #62, opened 2023-05-11
HTTP Server for RPCs

dongreenberg:
This PR replaces the default server for sending RPCs between clusters from gRPC to HTTP.


runhouse/servers/http/http_server.py
```diff
@@ -0,0 +1,390 @@
...
+            yield json.dumps(jsonable_encoder(Response(
+                data=codecs.encode(ret_obj, "base64").decode(),
+                output_type=OutputType.RESULT)))
+        else:
+            yield json.dumps(jsonable_encoder(Response(
+                error=pickle_b64(err),
+                traceback=pickle_b64(tb),
+                output_type=OutputType.EXCEPTION)))
+
+    def PutObject(self, request, context):
```
jlewitt1, line 226: can we remove context from all these? i think may have been leftover from grpc?

runhouse/servers/http/http_utils.py
```diff
@@ -0,0 +1,69 @@
...
+class OutputType:
+    EXCEPTION = "exception"
+    STDOUT = "stdout"
+    STDERR = "stderr"
+    SUCCESS = "success"  # No output
+    RESULT = "result"
+    RESULT_LIST = "result_list"
+
+
+def pickle_b64(picklable):
```
jlewitt1, line 34: why do we need to encode and pickle? is this for serializing python objects in text-based format?

runhouse/rns/hardware/cluster.py
```diff
@@ -9,6 +9,7 @@
 
 import grpc
```
jlewitt1, line 10:
```suggestion

```

runhouse/rns/hardware/cluster.py
```diff
@@ -314,32 +315,38 @@ def check_grpc(self, restart_grpc_server=True):
...
+                requests.exceptions.ConnectionError,
                 sshtunnel.BaseSSHTunnelForwarderError,
             ):
                 # It's possible that the cluster went down while we were trying to install packages.
                 if not self.is_up():
                     self.up_if_not()
                 else:
-                    self.restart_grpc_server(resync_rh=False)
+                    self.restart_server(resync_rh=False)
+        return
```
jlewitt1, line 333:
```suggestion
       
```

runhouse/rns/hardware/cluster.py
```diff
@@ -314,32 +315,38 @@ def check_grpc(self, restart_grpc_server=True):
...
-            self.connect_grpc()
+        if restart_server:
+            self.restart_server(resync_rh=False)
+            self.connect_server_client()
             if self.is_connected():
                 return
 
-            self.restart_grpc_server(resync_rh=True)
-            self.connect_grpc()
+            self.restart_server(resync_rh=True)
```
jlewitt1, line 348: why do we try again here? is this to see if tunnel is already up?

runhouse/servers/http/http_server.py
dongreenberg: Done!

runhouse/servers/http/http_utils.py
dongreenberg: Yes, exactly. Pickle doesn't go over the wire nicely.

runhouse/rns/hardware/cluster.py
dongreenberg: This was legacy, fixed!

dongreenberg:
@jlewitt1 @carolineechen this is just about ready to roll, thoughts and feedback would be super welcome!

docs/debugging_logging.rst
```diff
@@ -37,7 +37,7 @@ being printed out, and can debug normally inside the ``screen``.
     When debugging inside ``screen``, please use ``Ctrl A+D`` to exit out of the screen. Do NOT use ``Ctrl C``,
     which will terminate the GRPC server.
 
-    If you accidentally terminate the GRPC server, you can run ``cluster.restart_grpc_server()`` to restart the
+    If you accidentally terminate the GRPC server, you can run ``cluster.restart_server()`` to restart the
```
carolineechen, line 40: + another reference to GRPC above in this doc

```suggestion
    If you accidentally terminate the server, you can run ``cluster.restart_server()`` to restart the
```

runhouse/rns/hardware/cluster.py
```diff
@@ -431,7 +400,7 @@ def ssh_tunnel(
     #     connected = True
     #     print(f"SSH tunnel is open to {self.address}:{local_port}")
 
-    def restart_grpc_server(
+    def restart_server(
```
carolineechen, line 403: I am quite excited to type 5 fewer characters in my future debugging

runhouse/rns/hardware/cluster.py
```diff
@@ -217,39 +215,41 @@ def install_packages(
...
         return self.client.put_object(key, obj)
 
     def list_keys(self):
         """List all keys in the cluster's object store."""
-        self.check_grpc()
+        self.check_server()
         res = self.client.list_keys()
         return res
 
     def cancel(self, key: Optional[str] = None, force=False, all=False):
```
carolineechen, line 241: I think we can simplify to 

```
def cancel(self, key: str, force=False):
    """Cancel a given run on cluster by its key. If `key` is set to ``"all"``, then all jobs on the
        cluster will be cancelled."""
    self.check_server()
    return self.client.cancel_runs(key, force=force)
```

Also just checking, is the `key` always previously generated by us prior and not user set? (just double checking that "all" isn't a possible regular `key` str)

runhouse/rns/hardware/cluster.py
```diff
@@ -284,8 +284,8 @@ def connect_grpc(self, force_reconnect=False):
                 self._grpc_tunnel = ssh_tunnel
```
carolineechen, line 284: should we rename this too?

runhouse/rns/hardware/cluster.py
```diff
@@ -259,7 +259,7 @@ def on_this_cluster(self):
 
     # ----------------- gRPC Methods ----------------- #
```
carolineechen, line 260:
```suggestion
    # ----------------- HTTP Methods ----------------- #
```

runhouse/rns/hardware/cluster.py
```diff
@@ -441,10 +410,9 @@ def restart_grpc_server(
         # TODO how do we capture errors if this fails?
         if resync_rh:
             self.sync_runhouse_to_cluster(_install_url=_rh_install_url)
-            self.save_config_to_cluster()
-        kill_proc_cmd = 'pkill -f "python3 -m runhouse.servers.grpc.unary_server"'
-        logfile = f"{self.name}_grpc_server.log"
-        grpc_server_cmd = "python3 -m runhouse.servers.grpc.unary_server"
+        logfile = f"cluster_server_{self.name}.log"
```
carolineechen, line 413: can you also update the log file reference name in docs/debugging_logging.rst?

tests/test_cluster.py
```diff
@@ -59,7 +59,7 @@ def test_basic_run(cpu_cluster):
 @pytest.mark.clustertest
 def test_restart_grpc(cpu_cluster):
```
carolineechen, line 60:
```suggestion
def test_restart_server(cpu_cluster):
```

tests/test_obj_store.py
```diff
@@ -83,8 +83,7 @@ def test_cancel_run(cpu_cluster):
     print_fn = rh.function(fn=do_printing_and_logging, system=cpu_cluster)
     key = print_fn.remote()
     assert isinstance(key, str)
-    res = cpu_cluster.cancel(key)
-    assert res == "Cancelled"
+    cpu_cluster.cancel(key)
     try:
```
carolineechen, line 87: this test will still pass here even if cancel doesn't successfully do what it was meant to do

carolineechen (commented):
amazing!!

runhouse/rns/hardware/cluster.py
dongreenberg: Great point! For now it is only set by us, but when Runs land users would be able to set this. I've added another function "cancel_all" so we can change out the implementation underneath as needed later. Ideally this can be handled by the string matching logic of forking. Cc @jlewitt1

tests/test_obj_store.py
dongreenberg: Great point, and turned out it wasn't cancelling correctly! Fixed

runhouse/rns/hardware/cluster.py
dongreenberg: I can't hit the resolve button hard enough on this one. Maybe I should punchj my screen.

Issue #58, opened 2023-04-20
System cannot find the path specified


BStoller:
**Describe the bug**
I'm having an issue when trying to start up a Lang chain llm. After setting up the cluster
```python
gpu = rh.cluster('test', instance_type='T4:1', use_spot=False)
```

I attempt to create the llm that will run my inferences

```python
from langchain.llms import SelfHostedHuggingFaceLLM

llm = SelfHostedHuggingFaceLLM(model_id='dolly-v2-2-8b', hardware=gpu, model_reqs=['pip:./', 'transformers', 'torch'])
```

My code appears to run into some error with creating / finding a file. Hoping you all would be able to support.

```
INFO | 2023-04-20 11:38:47,871 | Setting up Function on cluster.
INFO | 2023-04-20 11:38:47,884 | Upping the cluster test
I 04-20 11:38:53 optimizer.py:617] == Optimizer ==
I 04-20 11:38:53 optimizer.py:628] Target: minimizing cost
I 04-20 11:38:53 optimizer.py:640] Estimated cost: $0.5 / hour
I 04-20 11:38:53 optimizer.py:640] 
I 04-20 11:38:53 optimizer.py:712] Considered resources (1 node):
I 04-20 11:38:53 optimizer.py:760] ---------------------------------------------------------------------------------------------------
I 04-20 11:38:53 optimizer.py:760]  CLOUD   INSTANCE               vCPUs   Mem(GB)   ACCELERATORS   REGION/ZONE   COST ($)   CHOSEN   
I 04-20 11:38:53 optimizer.py:760] ---------------------------------------------------------------------------------------------------
I 04-20 11:38:53 optimizer.py:760]  Azure   Standard_NC4as_T4_v3   4       28        T4:1           eastus        0.53          ✔     
I 04-20 11:38:53 optimizer.py:760] ---------------------------------------------------------------------------------------------------
I 04-20 11:38:53 optimizer.py:760] 
I 04-20 11:38:53 optimizer.py:775] Multiple Azure instances satisfy T4:1. The cheapest Azure(Standard_NC4as_T4_v3, {'T4': 1}) is considered among:
I 04-20 11:38:53 optimizer.py:775] ['Standard_NC4as_T4_v3', 'Standard_NC8as_T4_v3', 'Standard_NC16as_T4_v3'].
I 04-20 11:38:53 optimizer.py:775] 
I 04-20 11:38:53 optimizer.py:781] To list more details, run 'sky show-gpus T4'.
[...]

PR #63, opened 2023-05-12
Prevent RecursionError in import (for Windows, path dependent)

aria1th:
in Windows, if Python path is `C:\\Python`, then it can throw RecursionError on startup.

### DIFF
--- a/runhouse/rns/rns_client.py
+++ b/runhouse/rns/rns_client.py
@@ -80,13 +80,20 @@ class RNSClient:
             )
 
     @classmethod
-    def find_parent_with_file(cls, dir_path, file):
+    def find_parent_with_file(cls, dir_path, file, searched_dirs=None):
         if Path(dir_path) == Path.home() or dir_path == Path("/"):
             return None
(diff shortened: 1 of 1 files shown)


dongreenberg:
Great catch!! Thank you for adding it. Out of curiosity, did you run into any other issue on Windows or was this it?

aria1th:
@dongreenberg in runhouse side - this is the only error I catched, but in skypilot - I had some more issues with it 😭

dongreenberg:
@aria1th were they the same as https://github.com/run-house/runhouse/issues/58?

aria1th:
@dongreenberg Yes, that problem is solved by [pr in skypilot side](https://github.com/skypilot-org/skypilot/pull/1955). Its also in [issue ](https://github.com/skypilot-org/skypilot/issues/1950)
So runhouse does not have responsibility about that problem, but still some problem remains, such as not creating directory before executing rsync, etc. Well manually executing commands could solve but debugging was hard... so it might require separate logger for windows environment...

PR #65, opened 2023-05-16
Update private methods and function run message

carolineechen:


carolineechen:
@dongreenberg @jlewitt1 let me know if any of the methods I made private should stay public or if I missed any

runhouse/rns/folders/azure_folder.py
```diff
@@ -34,17 +34,17 @@ def download(self, dest):
         """Download a folder from a Azure bucket to local dir."""
         raise NotImplementedError
 
-    def download_command(self, src, dest):
```
dongreenberg, line 37: `upload_command` as well?

runhouse/rns/folders/gcs_folder.py
```diff
@@ -48,7 +48,7 @@ def upload(self, src: str, region: Optional[str] = None):
         )
```
dongreenberg, line 48: Part of me wonders if upload should be private too? Not clear when I'd use `upload` vs. `to`.

dongreenberg: Oh actually - `upload` means "this folder's system is already set to gcs, but the data is not necessarily there yet, so upload new local data there please." Essentially if a user creates a GCS Folder with dryrun or has an existing GCS folder but wants to put new data in it from a local path, they could use this command. But they could also use `to` to the same effect, but with the extra step of creating a local folder first to send to GCS. I think this is a confusing overlap so I'd be in favor of making this private until we see a real use case. Also worth noting that we already have `.put` to mean "I have a folder, please put this data into it", which also covers a big part of the use case surface here.

Cc @jlewitt1

dongreenberg (approved):
WOW, amazing turnaround

runhouse/rns/folders/folder.py
```diff
@@ -475,7 +475,7 @@ def to_cluster(self, dest_cluster, path=None, mount=False):
                 )
         else:
             raise TypeError(
-                f"`to_cluster` not supported for filesystem type {type(self.system)}"
+                f"`Sending from filesystem type {type(self.system)} is not supported"
```
carolineechen, line 478: @dongreenberg oh I had a question about this, is sending to a cluster not supported if `self.system` isn't a cluster or file? like if its 's3'

runhouse/rns/folders/gcs_folder.py
carolineechen: do you think `download` should also be made into a private function then? does `.to('here')` essentially cover its use case?

jlewitt1: Agreed there are a bunch of methods that more or less do the same thing, I don't see a great use case for it now that we have `to` and `put`. We only use it in one place when copying a folder to a data store:`new_folder.upload(src=local_folder_path)`, but we could just as easily use `.to` instead. 

I think making private for now makes sense. 

@carolineechen also agree re: `download`

runhouse/rns/folders/folder.py
jlewitt1: I think sending to a data store system like s3 should use `_to_data_store` instead of `_to_cluster`

PR #67, opened 2023-05-17
Cluster Factory Improvements

jlewitt1:
- Create distinct factory methods for each cluster type
- Try loading existing cluster if only `name` is provided
- BC: add `**kwargs` for base cluster factory, show warning but still try to instantiate an on demand cluster with kwargs provided
- Set `dryrun` to `False` by default


runhouse/rns/hardware/cluster.py
```diff
@@ -659,3 +660,54 @@ def remove_conda_env(
...
+            Example: ``ssh_creds={'ssh_user': '...', 'ssh_private_key':'<path_to_key>'}``
+        dryrun (bool): Whether to create the Cluster if it doesn't exist, or load a Cluster object as a dryrun.
+            (Default: ``False``)
+
+    Returns:
+        Cluster: The resulting cluster.
+
+    Example:
+        >>> # BYO Cluster
+        >>> gpu = rh.autocluster(ips=['<ip of the cluster>'],
```
carolineechen, line 689:
```suggestion
        >>> gpu = rh.cluster(ips=['<ip of the cluster>'],
```

runhouse/rns/hardware/on_demand_cluster.py
```diff
@@ -382,3 +382,83 @@ def ssh_creds(self):
...
+        >>>                  instance_type='A100:4',
+        >>>                  provider='gcp',
+        >>>                  autostop_mins=-1,
+        >>>                  use_spot=True,
+        >>>                  image_id='my_ami_string',
+        >>>                  region='us-east-1',
+        >>>                  )
+    """
+    if (
+        instance_type is None
```
carolineechen, line 431: can be simplified with something like 

```
if (
    set(instance_type, num_instances, provider, autostop_mins, image_id, region) == {None}
    and not use_spot and dryrun
)
```

or alternatively `all(param is None for param in (instance_type, num_instances, provider, autostop_mins, image_id, region))` is more intuitive but a little bit slower

tests/test_function.py
```diff
@@ -279,7 +279,7 @@ def delete_function_from_rns(s):
 
     try:
         # Terminate the cluster
-        s.cluster.teardown_and_delete()
+        s.autocluster.teardown_and_delete()
```
carolineechen, line 282: should be `s.cluster` here, `s` is a rh function

carolineechen (commented):
splitting it into different factory functions and having a 1:1 mapping for constructing different cluster types makes sense -- the cluster creator should know what type of cluster they are creating and can be expected to use the right method, and the different factory methods handle their unique operations.

however, for loading existing clusters, I think it would still be nice to support a generic `rh.cluster(name="name")`, which loads and returns the appropriate cluster. after all, they are all a "Cluster" Resource type in RNS, and this will simplify our internal code where we retrieve clusters from name without needing to parse subtypes to return the cluster, and also with cluster sharing or using multiple types of clusters, requires the user to keep track of what type each one is (maybe this is expected of them). what do you think?

runhouse/rns/hardware/cluster.py
jlewitt1: If I have an on demand cluster saved in RNS I can still re-load it with `rh.cluster(name="name", dryrun=True)`, and will get back an `OnDemandCluster` since we resolve its cluster type in the `from_config` method. 

I changed it above to be more explicit, but do you think it's confusing that we now have 2 ways to reload an on demand cluster (since you can now also reload it using `rh.autocluster`)?

dongreenberg (commented):
Submitting first batch of feedback

runhouse/rns/hardware/on_demand_cluster.py
```diff
@@ -378,3 +378,93 @@ def ssh_creds(self):
...
+                image_id,
+                region,
+            )
+        )
+        and not use_spot
+    ):
+        # If only the name is provided and dryrun is set to True
+        return OnDemandCluster.from_name(name, dryrun)
+
+    config = rns_client.load_config(name)
```
dongreenberg, line 449: Not needed

runhouse/rns/hardware/cluster.py
```diff
@@ -670,3 +673,57 @@ def remove_conda_env(
...
+        >>> gpu = rh.cluster(ips=['<ip of the cluster>'],
+        >>>                  ssh_creds={'ssh_user': '...', 'ssh_private_key':'<path_to_key>'},
+        >>>                  name='rh-a10x')
+
+        >>> # Load cluster from above
+        >>> reloaded_cluster = rh.cluster(name="rh-a10x", dryrun=True)
+    """
+    if ips is None and ssh_creds is None:
+        # If only the name is provided and dryrun is set to True
+        return Cluster.from_name(name, dryrun)
```
dongreenberg, line 711: I don't think this will handle the case where the name refers to an on-demand cluster

runhouse/rns/hardware/on_demand_cluster.py
```diff
@@ -378,3 +378,93 @@ def ssh_creds(self):
...
+        >>>                  use_spot=True,
+        >>>                  image_id='my_ami_string',
+        >>>                  region='us-east-1',
+        >>>                  )
+
+        >>> # Load cluster from above
+        >>> reloaded_cluster = rh.cluster(name="rh-4-a100s", dryrun=True)
+
+        >>> # Alternatively can also use the autocluster factory
+        >>> reloaded_cluster = rh.autocluster(name="rh-4-a100s", dryrun=True)
```
dongreenberg, line 430: Why dryrun?

runhouse/rns/packages/package.py
```diff
@@ -405,11 +405,10 @@ def package(
     install_method: str = None,
     install_str: str = None,
     path: str = None,
-    system: str = Folder.DEFAULT_FS,
+    system: str = None,
```
dongreenberg, line 408: Is this going to affect the behavior?

runhouse/rns/packages/package.py
```diff
@@ -434,11 +432,23 @@ def package(
...
+        and install_str is None
+        and path is None
+        and system is None
+        and data_config is None
+        and not local_mount
+    ):
+        # If only the name is provided and dryrun is set to True
+        return Package.from_name(name, dryrun)
+
+    config = rh_config.rns_client.load_config(name)
```
dongreenberg, line 446: Not needed

runhouse/rns/tables/pandas_table.py
```diff
@@ -23,7 +23,7 @@ def __iter__(self):
                 yield row
 
     @staticmethod
-    def from_config(config: dict, dryrun=True):
+    def from_config(config: dict, dryrun=False):
         """Load config values into the object."""
         if isinstance(config["system"], dict):
             config["system"] = OnDemandCluster.from_config(
```
dongreenberg, line 29: This won't work for a static cluster

dongreenberg: Also we should use Resource._check_for_child_configs here.

runhouse/rns/tables/table.py
```diff
@@ -466,9 +464,27 @@ def table(
...
+                data_config,
+                partition_cols,
+                stream_format,
+                metadata,
+            )
+        )
+        and not mkdir
+    ):
+        # Try reloading existing table
+        return Table.from_name(name, dryrun)
```
dongreenberg, line 485: Name needs to be set

runhouse/rns/tables/table.py
```diff
@@ -466,9 +464,27 @@ def table(
...
+
+    config = rns_client.load_config(name)
```
dongreenberg, line 487: Not needed

runhouse/rns/tables/table.py
```diff
@@ -466,9 +464,27 @@ def table(
         >>> )
         >>>
         >>> # Load table from above
-        >>> reloaded_table = rh.table(name="~/my_test_pandas_table")
+        >>> reloaded_table = rh.table(name="~/my_test_pandas_table", dryrun=True)
```
dongreenberg, line 467: Why dryrun?

tests/conftest.py
```diff
@@ -120,7 +120,7 @@ def cluster(request):
 
 @pytest.fixture(scope="session")
 def cpu_cluster():
-    c = rh.cluster("^rh-cpu")
+    c = rh.autocluster("^rh-cpu")
```
dongreenberg, line 123: Can just be cluster

tests/conftest.py
```diff
@@ -155,17 +155,19 @@ def byo_cpu():
 
 @pytest.fixture(scope="session")
 def v100_gpu_cluster():
-    return rh.cluster("^rh-v100", provider="aws").up_if_not()
+    return rh.autocluster("^rh-v100", provider="aws").up_if_not()
```
dongreenberg, line 158: Won't work anymore, because we don't allow overwriting configs if a name is passed

tests/conftest.py
```diff
@@ -199,7 +201,7 @@ def summer(a: int, b: int):
 
 
 def save_and_load_artifacts():
-    cpu = rh.cluster("^rh-cpu").save()
+    cpu = rh.autocluster("^rh-cpu").save()
```
dongreenberg, line 204: Should just be cluster

tests/test_blob.py
```diff
@@ -30,7 +30,7 @@ def test_create_and_reload_local_blob_with_name(blob_data):
     del blob_data
     del my_blob
 
-    reloaded_blob = rh.Blob.from_name(name)
+    reloaded_blob = rh.blob(name=name, dryrun=True)
```
dongreenberg, line 33: Why dryrun?

dongreenberg: Ditto for all below

tests/test_env.py
```diff
@@ -143,7 +143,7 @@ def test_conda_env_from_name_local():
     local_env = rh.env(name=env_name, conda_env=conda_env).save()
     del local_env
 
-    remote_env = rh.env(name=env_name)
+    remote_env = rh.env(name=env_name, dryrun=True)
```
dongreenberg, line 146: Why dryrun?

tests/test_folder.py
```diff
@@ -292,7 +292,7 @@ def test_s3_sharing(s3_folder):
         headers=headers,
     )
 
-    my_folder = rh.Folder.from_name("@/my-s3-shared-folder")
+    my_folder = rh.folder(name="@/my-s3-shared-folder", dryrun=True)
```
dongreenberg, line 295: Dryrun

tests/test_function.py
```diff
@@ -41,7 +41,7 @@ def test_create_function_from_name_local(cpu_cluster):
     local_sum = rh.function(summer).to(cpu_cluster).save(local_name)
     del local_sum
 
-    remote_sum = rh.Function.from_name(local_name)
+    remote_sum = rh.function(name="local_function", dryrun=True)
```
dongreenberg, line 44: dryrun

tests/test_package.py
```diff
@@ -85,7 +85,7 @@ def test_share_git_package():
 
 @pytest.mark.rnstest
 def test_load_shared_git_package():
-    git_package = rh.Package.from_name(name="@/shared_git_package")
+    git_package = rh.package(name="@/shared_git_package", dryrun=True)
```
dongreenberg, line 88: dryrun

tests/test_table.py
```diff
@@ -42,7 +42,7 @@ def test_create_and_reload_file_locally(tmp_path):
         .save()
     )
 
-    reloaded_table = rh.Table.from_name(name)
+    reloaded_table = rh.table(name=name, dryrun=True)
```
dongreenberg, line 45: dryrun

dongreenberg (commented):
Sorry to be repetitive, these changes are sweeping and wanted to make sure the comments propagate everywhere

PR #74, opened 2023-06-20
Add docstring examples and private some methods

carolineechen:


runhouse/rns/hardware/cluster.py
```diff
@@ -628,8 +669,15 @@ def run_python(
         return return_codes
 
     def send_secrets(self, providers: Optional[List[str]] = None):
```
carolineechen, line 671: would `get/sync/retrieve_secrets` make more sense? `cluster.send_secrets` makes it sound like the secrets on the cluster are being sent, rather than sending from local to the cluster.

runhouse/rns/blob.py
```diff
@@ -146,20 +167,35 @@ def write(self):
...
+        """Check whether the blob exists in the file system
+
+        Example:
+            >>> blob = rh.blob(serialized_data, path="saved/path")
+            >>> blob.exists_in_system()
+        """
         return self._folder.fsspec_fs.exists(self.fsspec_url)
 
     # TODO [DG] get rid of this in favor of just "sync_down(path, system)" ?
     def sync_from_cluster(self, cluster, path: Optional[str] = None):
```
jlewitt1, line 188: are we still using this?

PR #78, opened 2023-06-23
In memory blob object

dongreenberg:
- [x] Introduce in-memory blob, replace pinning in obj store tests with blob interface. Tests all pass (pinning fixed so all Ray processes can access the KV store).
- [x] Introduce Env Servlets
  - [x] Reach parity with existing server, including streaming logs
- [x] Add run_module_method to server and servlet. Add new test to call a method on a blob in the blob store.
- [x] Add put_resource RpC
- [x] Support streaming results from generator function or property through call_module_method
  - [x] Add `rh.Queue` to support streaming across Ray boundary
  - [ ] Support `.get` without popping value (subsequent PR)
  - [ ] Support persisting queue (subsequent PR)
- [x] Change env.to to use cluster.put_resource(env), then cluster.call_module_method(env.name, "install")
  - [ ] Update conda_env to do this too
  - [ ] Refactor package to always accept env in _install and use env.run instead of many subprocess.calls
  - [x] Introduce simple env install caching
- [x] Set up saving / streaming logs for calls
  - [x] Merge in Josh's branch which streams to file without using ray logs
  - [x] Fix capturing stdout and capturing logs from generators
  - [x] Support capturing logs for call call_module_method calls, not just function
- [x] Change obj_store to handle local servlet kvs (in Python dict to enable proper pinning) and global kvs
  - [x] Introduce `rh.KVStore`
- [x] Change fn.to to use cluster.put_resource(fn), then cluster.call_module_method(env.name, "__call__")
  - [x] calling `fn.save(new_name)` needs to rename the resource in the cluster's kvstore.
  - [x] Make `.remote` return an actual remote Module, and `.run` return the runkey async
  - [ ] Allow `.remote` for stream to return a Queue while results are still being generated so user can `.get` each new result
  - [ ] Make like-clusters share clients so we don't need to recheck the cluster when working with objects returned by `.remote` (which will contain a new cluster object)
  - [ ] Fix cancelling
- [x] Add simple numpy-based test to evaluate whether pinning keeps object in python memory (based on https://docs.ray.io/en/latest/ray-core/objects/serialization.html#fixing-assignment-destination-is-read-only)
- [x] Add rh.Module parent for all classes which can hold a system and have their methods execute remotely on the system, include user-defined classes
  - [x] Move .run and .remote to there instead of function
  - [ ] Change existing remotable resources to just describe local execution log and rely on Module to handle rpc
    - [x] KVStore
    - [x] Queue
    - [x] Blob
    - [ ] Folder
    - [ ] Table (subsequent PR)
- [ ] Change Run from being a resource to a package for provenance inside Resources
   - [x] Add .provenance to Resource and various constructors
   - [x] update Run's config_from_rns and from_config to contain provenance data directly rather than relying on filesystem config
- [ ] Change system to handle _current_cluster case for `system.run` and RpCs
  - [ ] Handle _current_cluster case for is_up, up_if_not, etc.
- [ ] Change Run save/load behavior with new object store
  - [ ] Make run an optional property of resources instead of a Resource itself 
- [ ] Test generator with small LM
- [ ] Change blob.to to handle to&from same cluster case (no bouncing off laptop)
- [ ] Deprecate old RPCs which are no longer needed.
- [ ] Document more clearly
- [ ] Update tutorials


runhouse/rns/blobs/blob.py
```diff
@@ -0,0 +1,291 @@
...
+        Example:
+            >>> blob = rh.blob(data)
+            >>> blob.rm()
+        """
+        if self.system is None:
+            self._data = None
+        if self.system.on_this_cluster():
+            obj_store.delete(self.name)
+        else:
+            self.system.delete(self.name)
```
jlewitt1, line 192: Is this supposed to delete the blob from the cluster's local file system? We don't have a `delete` method for the cluster which does this (i think previously this went through the blob's `folderz, how can we do this now that a blob doesn't live in a folder object?

dongreenberg:
Merging this into main for now, but still more cleanup to do.

PR #89, opened 2023-07-30
Sagemaker integration POC

jlewitt1:
Creates a Runhouse abstraction for connecting to SageMaker compute via the `SageMakerCluster` class. 

Start by supporting two core usage paths: 
- Compute backend:  Use SageMaker as a compute backend, just as you would an on-demand cluster or on-prem/byo cluster 
- Dedicated training jobs: Launch a training job by providing a SageMaker Estimator object

Core functionality:
- [x] Request compute via the SageMaker API
- [x] Start HTTP server + Ray on the compute
- [x] rsync support
- [x] Autostop support
- [x] Launch dedicated SageMaker training jobs  
- [x] Use SageMaker as a pure "backend" (similar to BYO Clusters or On-Demand clusters)
- [x] Tearing down / stopping the instance
- [x] Logging support (same as other cluster types)
- [x] Launch instance with custom image 
- [x] Fix GPU related errors in cluster's post installation script 
- [x] Use SageMaker role ARN configured in aws credentials file
- [x] Reliably importing functions to run on the cluster from any module 
- [x] Performance: Do not re-generate ssh keys when re-connecting to an existing cluster
- [x] Create reliable SSH connection to the compute (via https://github.com/aws-samples/sagemaker-ssh-helper)
- [x] Managing SSH connections with multiple active SageMaker clusters concurrently (currently unreliable)
- [ ] Handling various SageMaker specific dependencies: awscli, boto, sagemaker, and removing fsspec & s3fs dependencies
- [ ] Portability: any SageMaker cluster resource saved to Runhouse Den should be enough to recreate or reconnect to from other environments
- [x] AWS role / profile flexibility: specify either a specific Role ARN or a profile name with necessary SageMaker access credentials
- [ ] Add more robust SageMaker cluster tests, add dedicated pytest marker

Additional features:
- [ ] SageMaker inference endpoints
- [ ] Support for multi-instance clusters (untested)
- [ ] Secrets: Adding ARN role portability (required for SageMaker APIs) 
- [ ] Support for creating / re-loading in collab / notebook environments


dongreenberg:
I was poking around with the Sagemaker connection flow to see if we can do it faster by reusing the keys that get created. Turns out the answer is a big yes, but reveals an important can of worms we need to deal with.

Essentially, when you run the `sm-local-start-ssh` command with the tunnel args, sagemaker-ssh is calling another script `sm-connect-ssh-proxy` (https://github.com/aws-samples/sagemaker-ssh-helper/blob/main/sagemaker_ssh_helper/sm-connect-ssh-proxy). If you look inside that script, you can see that at the very end after all the key setup (essentially creating a keypair and copying it to s3 to then scp it onto the cluster's authorized_keys), it makes a normal ssh call using an `aws ssm` command as the ProxyCommand. If you look in the readme, they actually specify to call a similar command, `aws ssm start-session --target mi-1234567890abcdef0` to essentially manually connect to the cluster using ssm which creates a secure session with the desired instance.

So first I tried just running that `aws ssm` command, and it connects an ssh session instantly, great. I assumed that was because I already had the keypair, but no, even if I rename the key it connects instantly through ssm (but doesn't create the tunnel). So I ran the full tunnel command (including the `aws ssm` command inside the ProxyCommand) at the end of the `sm-connect-ssh-proxy` script, and it too connects instantly. Great. I could also run `ssh rh-sagemaker` (through the lines automatically added to my `.ssh/config`) as I wanted and connect in. But then I killed it and for some reason tried running `ssh rh-sagemaker` again just to confirm that the tunnel closed, but it connected. Strange. I thought maybe the test I just ran in PyCharm might have not freed everything properly, so I killed all of PyCharm. The ssh command still went through. Keep in mind this ssh command is going through localhost, because the host configured in my `.ssh/config` behind rh-sagemaker is localhost. I turned off my wifi for a minute to kill any ssh sessions that might be up, turned it back on and tried again, and it connected. Yikes. Then I commented out the ssh tunnel pieces of the sagemaker_cluster.py and reran my sagemaker cpu cluster test I was running, and it passed instantly with zero initial connection time, faster than even on_demand_cluster tests. That's great news but a bit wild.

So ssm creates a Session which is quite long lived. There's a command to see all of your active ssm sessions (`aws ssm describe-sessions --state Active`) and I have dozens just from the last few days of sagemaker work. I went to the AWS console and I see a shitload (about 300) that all say "connected" created by me and Josh. It's supposed to timeout after 20 minutes of inactivity but even now more than 20 minutes later I can still rerun tests with the tunnel logic commented out and they pass. There are aws cli commands to delete sessions.

So I think there are a few important takeaways.
1. The reason we can't create tunnels to more than one sagemaker cluster at a time is that sagemaker-ssh is creating a long-lived tunnel through a specific port. Even now, the only way I can make it so I can't connect with `ssh rh-sagemaker`  is to `ps | grep ssh` and then `kill -9` the pid for the `sm-local-start-ssh` command that was run the first time I ran the test. We need to change that port to have multiple connections open.
2. We probably don't need to create a new tunnel at all after we start the session the first time. We do need to figure out the exact flow though.
3. We probably need to somehow be terminating these sessions after the compute goes down to clean up after ourselves. Maybe we want to set them to auto-terminate in the same inactivity period as the cluster autostop, and just recreate the session if we need to later.
4. We can bypass sagemaker-ssh entirely if we want, by creating the key pair ahead of time or using an existing key pair, putting the key-pair onto the cluster in the estimator script at launch time, and then creating the ssm session ourselves.
5. In general I think we'd be happier if we supported longer lived cluster connections like this, including for on_demand. We've talked about that a lot but now's a good time to introduce it with the secrets work.

PR #109, opened 2023-09-22
feat(k8s): adds kubernetes support to Runhouse

RohanSreerama5:
Adds Kubernetes support to Runhouse. 
Utilizes Skypilot API. 

Example usage: 
```
cluster = rh.cluster(
              name="cpu-cluster-39",
              instance_type="CPU:1",
              provider="kubernetes",      
          )
```

Function setup and function calls work with this new support. Further testing and explorations will be done soon.

### DIFF
--- a/runhouse/main.py
+++ b/runhouse/main.py
@@ -103,7 +103,27 @@ def _start_server(restart, restart_ray, screen, create_logfile=True):
 
     try:
         # We do these one by one so it's more obvious where the error is if there is one
-        for cmd in cmds:
+        for i, cmd in enumerate(cmds):
+
+            last_cmd = i == len(cmds)-1
+
+            # need to clean up this execution logic 
+            if last_cmd:
+                try:
+                    # consider cleaning up this command to parameterize it more
+                    new_str = "nohup /opt/conda/bin/python -m runhouse.servers.http.http_server >> /home/sky/.rh/server.log 2>&1 &"
+                    console.print(f"Executing `{new_str}`")
+                    output = subprocess.run(new_str, shell=True, check=True)
+                    if output.returncode != 0:
+                        reg_output = subprocess.run(shlex.split(cmd), text=True)
+                        if reg_output.returncode != 0:
+                            console.print(f"Error while executing `{cmd}`")
+                            raise typer.Exit(1)
+                        return 
+                    return
+                except subprocess.CalledProcessError as e:
+                    print(f"Error: {e}")    
+
             console.print(f"Executing `{cmd}`")
             result = subprocess.run(shlex.split(cmd), text=True)
             # We don't want to raise an error if the server kill fails, as it may simply not be running
--- a/runhouse/resources/hardware/cluster.py
+++ b/runhouse/resources/hardware/cluster.py
@@ -98,6 +98,10 @@ class Cluster(Resource):
             from .sagemaker_cluster import SageMakerCluster
 
             return SageMakerCluster(**config, dryrun=dryrun)
+        elif resource_subtype == "KubernetesCluster":
+            from .kubernetes_cluster import KubernetesCluster
+
+            return KubernetesCluster(**config, dryrun=dryrun)
         else:
             raise ValueError(f"Unknown cluster type {resource_subtype}")
 
--- a/runhouse/resources/hardware/kubernetes_cluster.py
+++ b/runhouse/resources/hardware/kubernetes_cluster.py
@@ -1,0 +1,127 @@
+import sky
+from pathlib import Path
+import subprocess
+import logging
+import time
+
+from runhouse.resources.hardware.on_demand_cluster import OnDemandCluster
+
+logger = logging.getLogger(__name__)
+
+class KubernetesCluster(OnDemandCluster):
+
+    def up(self):
+        """Up the cluster.
+
+        Example:
+            >>> rh.kubernetes_cluster(
+            >>>     name="cpu-cluster-test",
+            >>>     instance_type="CPU:1",
+            >>>     provider="kubernetes",      
+            >>> )
+        """
+        if self.on_this_cluster():
+            return self
+
+        if self.provider in ["kubernetes"]:
+            task = sky.Task(
+                num_nodes=self.num_instances
+                if self.instance_type and ":" not in self.instance_type
+                else None,
+                # docker_image=image,  # Zongheng: this is experimental, don't use it
+                # envs=None,
+            )
+            cloud_provider = (
+                sky.clouds.CLOUD_REGISTRY.from_str(self.provider)
+                if self.provider != "cheapest"
+                else None
+            )
+            task.set_resources(
+                sky.Resources(
+                    cloud=cloud_provider,
+                    instance_type=self.instance_type
+                    if self.instance_type
+                    and ":" not in self.instance_type
+                    and "CPU" not in self.instance_type
+                    else None,
+                    accelerators=self.instance_type
+                    if self.instance_type
+                    and ":" in self.instance_type
+                    and "CPU" not in self.instance_type
+                    else None,
+                    cpus=self.instance_type.rsplit(":", 1)[1]
+                    if self.instance_type
+                    and ":" in self.instance_type
+                    and "CPU" in self.instance_type
+                    else None,
+                    region=self.region,
+                    image_id=self.image_id,
+                    use_spot=self.use_spot,
+                )
+            )
+            if Path("~/.rh").expanduser().exists():
+                task.set_file_mounts(
+                    {
+                        "~/.rh": "~/.rh",
+                    }
+                )
+            # If we choose to reduce collisions of cluster names:
+            # cluster_name = self.rns_address.strip('~/').replace("/", "-")
+            sky.launch(
+                task,
+                cluster_name=self.name,
+                idle_minutes_to_autostop=self.autostop_mins,
+                down=True,
+            )
+        else:
+            raise ValueError(f"Cluster provider {self.provider} not supported.")
+
+        self._update_from_sky_status()
+        
+        self.restart_server(restart_ray=True)
+
+        return self
+
+    def restart_server(
+        self,
+        _rh_install_url: str = None,
+        resync_rh: bool = True,
+        restart_ray: bool = True,
+        env_activate_cmd: str = None,
+    ):
+        """Restart the RPC server.
+
+        Args:
+            resync_rh (bool): Whether to resync runhouse. (Default: ``True``)
+            restart_ray (bool): Whether to restart Ray. (Default: ``True``)
+            env_activate_cmd (str, optional): Command to activate the environment on the server. (Default: ``None``)
+        Example:
+            >>> rh.kubernetes_cluster("rh-cpu").restart_server()
+        """
+        logger.info(f"Restarting HTTP server on {self.name}.")
+
+        if resync_rh:
+
+            # sync_rh_cmd = "kubectl cp ../runhouse/ default/cpu-cluster-42-dc19-ray-head:runhouse"
+
+            # consider modifying this so that the kubernetes namespace is parameterized 
+            # also consider extracting the kubernetes pod name 
+            sync_rh_cmd = f"kubectl cp ../runhouse/ default/{self.name}-dc19-ray-head:runhouse"
+
+            cmd = f"{sync_rh_cmd}"
+            try:
+                subprocess.run(cmd, shell=True, check=True)
+            except subprocess.CalledProcessError as e:
+                print(f"Error: {e}")
+
+        install_rh_cmd = "pip install ./runhouse"
+
+        cmd_1 = self.CLI_RESTART_CMD + (" --no-restart-ray" if not restart_ray else "")
+        cmd_1 = f"{install_rh_cmd} && {cmd_1}"
+
+        status_codes = self.run(commands=[cmd_1])
+        if not status_codes[0][0] == 0:
+            raise ValueError(f"Failed to restart server {self.name}.")
+        # As of 2023-15-May still seems we need this.
+        time.sleep(5)
+        return status_codes
--- a/runhouse/resources/hardware/utils.py
+++ b/runhouse/resources/hardware/utils.py
@@ -65,9 +65,10 @@ class SkySSHRunner(SSHCommandRunner):
         ssh_private_key=None,
         ssh_control_name=None,
         ssh_proxy_command=None,
+        disable_control_master: Optional[bool] = False
     ):
         super().__init__(
-            ip, ssh_user, ssh_private_key, ssh_control_name, ssh_proxy_command
+            ip, ssh_user, ssh_private_key, ssh_control_name, ssh_proxy_command, disable_control_master
         )
         self._tunnel_procs = []
 
--- a/runhouse/__init__.py
+++ b/runhouse/__init__.py
@@ -10,6 +10,8 @@ from runhouse.resources.hardware import (
     OnDemandCluster,
     sagemaker_cluster,
     SageMakerCluster,
+    kubernetes_cluster,
+    KubernetesCluster
 )
 
 # WARNING: Any built-in module that is imported here must be capitalized followed by all lowercase, or we will
--- a/runhouse/resources/function.py
+++ b/runhouse/resources/function.py
@@ -129,12 +129,16 @@ class Function(Module):
 
         logging.info("Setting up Function on cluster.")
         # To up cluster in case it's not yet up
+
         new_function.system.check_server()
         new_function.name = new_function.name or self.fn_pointers[2]
         # TODO
         # env.name = env.name or (new_function.name + "_env")
-        new_env = env.to(new_function.system, force_install=force_install)
-        new_function.env = new_env
+
+        # temporarily commented out the below to enable k8s support 
+
+        # new_env = env.to(new_function.system, force_install=force_install)
+        # new_function.env = new_env
 
         new_function.dryrun = True
         system.put_resource(new_function, dryrun=True)
--- a/runhouse/resources/hardware/__init__.py
+++ b/runhouse/resources/hardware/__init__.py
@@ -1,5 +1,6 @@
 from .cluster import Cluster
-from .cluster_factory import cluster, ondemand_cluster, sagemaker_cluster
+from .cluster_factory import cluster, ondemand_cluster, sagemaker_cluster, kubernetes_cluster
 from .on_demand_cluster import OnDemandCluster
 from .sagemaker_cluster import SageMakerCluster
+from .kubernetes_cluster import KubernetesCluster
 from .utils import _current_cluster, _get_cluster_from, RESERVED_SYSTEM_NAMES
--- a/runhouse/resources/hardware/cluster_factory.py
+++ b/runhouse/resources/hardware/cluster_factory.py
@@ -6,6 +6,7 @@ from runhouse.resources.hardware.utils import RESERVED_SYSTEM_NAMES
 from .cluster import Cluster
 from .on_demand_cluster import OnDemandCluster
 from .sagemaker_cluster import SageMakerCluster
+from .kubernetes_cluster import KubernetesCluster
 
 
 # Cluster factory method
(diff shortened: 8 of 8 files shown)


runhouse/resources/hardware/kubernetes_cluster.py
```diff
@@ -0,0 +1,127 @@
...
+            >>> rh.kubernetes_cluster(
+            >>>     name="cpu-cluster-test",
+            >>>     instance_type="CPU:1",
+            >>>     provider="kubernetes",      
+            >>> )
+        """
+        if self.on_this_cluster():
+            return self
+
+        if self.provider in ["kubernetes"]:
```
carolineechen, line 26: should we just default the provider for `kubernetes_cluster()` to "kubernetes" and throw an error in the factory function if they input something else?

runhouse/resources/hardware/kubernetes_cluster.py
```diff
@@ -0,0 +1,127 @@
...
+            task = sky.Task(
```
carolineechen, line 27: I think this section is the same as ondemandcluster's `up()`, we can extract the part inside the if statement to a helper function and reuse it directly to avoid duplicate code

runhouse/scripts/kubernetes_cluster/main.tf
```diff
@@ -0,0 +1,94 @@
+# Script to spin up a quick EKS cluster that can be used for testing Runhouse Kubernetes support
```
carolineechen, line 1: nice!

runhouse/main.py
```diff
@@ -103,7 +103,27 @@ def _start_server(restart, restart_ray, screen, create_logfile=True):
...
-        for cmd in cmds:
+        for i, cmd in enumerate(cmds):
+
+            last_cmd = i == len(cmds)-1
+
+            # need to clean up this execution logic 
+            if last_cmd:
+                try:
+                    # consider cleaning up this command to parameterize it more
+                    new_str = "nohup /opt/conda/bin/python -m runhouse.servers.http.http_server >> /home/sky/.rh/server.log 2>&1 &"
```
carolineechen, line 114: curious what's the reason for modifying the last command?

jlewitt1: there were issues with running screen commands inside the pod, is that still ongoing @RohanSreerama5 ? Also can we generalize parts of the command (ex: using `sys.executable` instead of the hard coded path, relative path to the `server.log` file) so its extensible to the other cluster types?

RohanSreerama5: Yeah, it wasn't super trivial to install screen on the pod, so I ended up using `nohup` for the purpose of this POC. 

In a future commit, I can be sure to parameterize and generalize the command better.

runhouse/resources/hardware/kubernetes_cluster.py
```diff
@@ -0,0 +1,127 @@
...
+import subprocess
+import logging
+import time
+
+from runhouse.resources.hardware.on_demand_cluster import OnDemandCluster
+
+logger = logging.getLogger(__name__)
+
+class KubernetesCluster(OnDemandCluster):
+
```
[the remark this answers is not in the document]
RohanSreerama5: It was not necessary to add any new parameters to the parent constructor, so I didn't initialize it here.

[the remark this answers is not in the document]
RohanSreerama5: Of course, I can see why it could be best practice to initialize it regardless so I can add it in a future commit.

runhouse/resources/hardware/kubernetes_cluster.py
RohanSreerama5: Yeah, I think that would be good.

KubernetesTest.ipynb
```diff
@@ -0,0 +1,869 @@
+{
```
carolineechen, line 1: just checking that we'll be removing this in favor of unit tests when we merge, correct? we should avoid using ipynbs for tests, and theyre also quite large/slow to send to the cluster which is done by default

runhouse/resources/hardware/kubernetes_cluster.py
```diff
@@ -0,0 +1,108 @@
+import copy
+import logging
+import os
+import subprocess
+import warnings
```
jlewitt1, line 5: looks like there are some unused imports here

if you run the linter as well, will make clear which can be removed

`pre-commit run --all-files`

runhouse/resources/hardware/kubernetes_cluster.py
```diff
@@ -0,0 +1,108 @@
...
+
+            output = subprocess.check_output(command, shell=True, text=True)
+
+            lines = output.strip().split("\n")
+            if lines:
+                pod_name = lines[0].split()[0]
+            else:
+                logger.info("No matching pods found.")
+        except subprocess.CalledProcessError as e:
+            logger.info(f"Error: {e}")
```
jlewitt1, line 105: raise or return here since theres no pod name to ssh into?

runhouse/scripts/kubernetes_cluster/main.tf
```diff
@@ -0,0 +1,93 @@
...
```
jlewitt1, line 1: can we add a README to this folder explainijng this sript, what needs to be changed, etc.?

tests/test_resources/test_clusters/test_kubernetes_cluster/test_kubernetes.py
```diff
@@ -0,0 +1,62 @@
+# import os
+# import unittest
+
+# import pytest
+
+# import runhouse as rh
+
+# from tests.test_resources.test_clusters.cluster_tests import np_array, sd_generate_image
+
+# @pytest.mark.clustertest
```
jlewitt1, line 10: can we remove these?

runhouse/main.py
```diff
@@ -129,14 +130,44 @@ def _start_server(
...
+                        console.print(f"Executing with nohup because screen is not available: `{nohup_cmd}`")
+                        output = subprocess.run(nohup_cmd, shell=True, check=True)
+                        if output.returncode != 0:
+                            console.print(f"Error while executing `{cmd}`")
+                            raise typer.Exit(1)
+                        
+                        break
+       
+                except subprocess.CalledProcessError as e:
+                    print(f"Error: {e}")
```
jlewitt1, line 169:
```suggestion
                    console.print(f"Error: {e}")
```

KubernetesTest.ipynb
jlewitt1: in terms of examples, we can also add some more info to the core [cluster API docs](https://github.com/run-house/runhouse/blob/main/docs/api/python/cluster.rst)

carolineechen: yup, I think it makes sense to remove the notebook and split the content out into
* unit tests in tests/ folder
* examples in rh docs, either within docstrings or explicit k8s section
* tutorial for k8s cluster. the compute tutorial under clusters probably makes the most sense for now

runhouse/resources/hardware/kubernetes_cluster.py
```diff
@@ -0,0 +1,108 @@
...
+        return KubernetesCluster(**config, dryrun=dryrun)
+
+    @property
+    def config_for_rns(self):
+        config = super().config_for_rns
+
+        # Also store the ssh keys for the cluster in RNS
+        config.update(
+            {
+                "instance_type": self.instance_type,
```
carolineechen, line 66: since kubernetes cluster already subclasses ondemandcluster, we don't need to include overlapping variables in this update, like instance_type, num_instances, etc.

runhouse/resources/hardware/kubernetes_cluster.py
```diff
@@ -0,0 +1,108 @@
...
+                "context": self.context,
+            }
+        )
+        return config
+
+    def ssh(self):
+        """SSH into the cluster.
+
+        Example:
+            >>> cluster = rh.kubernetes_cluster (
```
carolineechen, line 86:
```suggestion
            >>> cluster = rh.kubernetes_cluster(
```

runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -374,8 +374,8 @@ def up(self):
                 idle_minutes_to_autostop=self.autostop_mins,
                 down=True,
             )
-        elif self.provider == "k8s":
-            raise NotImplementedError("Kubernetes Cluster provider not yet supported")
+        # elif self.provider == "k8s":
```
carolineechen, line 377: please clean up any debug/old comments before merging

setup.py
```diff
@@ -66,10 +66,10 @@ def parse_readme(readme: str) -> str:
     "fastapi",
     "fsspec<=2023.5.0",
     "pyarrow",
-    "pyOpenSSL>=21.1.0",
+    "pyOpenSSL>=23.3.0",
```
carolineechen, line 69: curious was the previous version of pyopenssl breaking something in kubernetes? also, if you could also update the version in `requirements.txt` to match this

tests/fixtures/kubernetes_cluster_fixtures.py
```diff
@@ -0,0 +1,33 @@
+import pytest
```
carolineechen, line 1: Should we remove this file (or conftest?). looks like a duplicate

tests/test_resources/test_clusters/cluster_tests.py
```diff
@@ -53,6 +53,29 @@ def test_read_shared_cluster(ondemand_cpu_cluster):
...
+def test_function_on_k8s_cluster(kubernetes_cpu_cluster):
+    def num_cpus():
+        import multiprocessing
+
+        return f"Num cpus: {multiprocessing.cpu_count()}"
+
+    num_cpus_cluster = rh.function(name="num_cpus_cluster", fn=num_cpus).to(
+        system=kubernetes_cpu_cluster, reqs=["./"]
+    )
+    assert num_cpus_cluster() == "Num cpus: 4"
```
carolineechen, line 76: is it guaranteed this is always 4 from `instance_type="1CPU--1GB"`?

runhouse/main.py
```diff
@@ -129,14 +130,54 @@ def _start_server(
...
         if screen and Path(Cluster.SERVER_LOGFILE).exists():
             f = open(Cluster.SERVER_LOGFILE, "r")
             f.readlines()  # Discard these, they're from the previous times the server was started
 
         # We do these one by one so it's more obvious where the error is if there is one
-        for cmd in cmds:
+        for i, cmd in enumerate(cmds):
+
+            last_cmd = i == len(cmds) - 1
+            if last_cmd:
```
dongreenberg, line 144: Iterating over all the commands to pull out the one that you want to manipulate is a bit roundabout. Could you just run the `command -v screen` before `Cluster._start_server_cmds` and then make the `screen` argument to `Cluster._start_server_cmds` accept either `None`, `screen`, or `nohup` to construct the correct command there?

runhouse/resources/hardware/kubernetes_cluster.py
```diff
@@ -0,0 +1,108 @@
...
+                "region": self.region,
+                "live_state": self._get_sky_state(),
+                "namespace": self.namespace,
+                "kube_config_path": self.kube_config_path,
+                "context": self.context,
+            }
+        )
+        return config
+
+    def ssh(self):
```
dongreenberg, line 82: If this and storing the context are the only major fork from on_demand_cluster, I think we should likely use OnDemandCluster and branch in the `ssh` method. Theoretically we could store the context, namespace, and kubeconfig in the ssh_creds as long as SkyPilot makes them settable (we should chat with them about that). At some point when we introduce ClusterSecrets we can have variants for SSH and Kubeconfig secrets, which can also hold the connection logic. For now we can leave it in a separate class and chat with the skypilot team further. The gap where a user creates an regular OnDemandCluster and SkyPilot gives them a k8s pod back isn't a big deal, because the only gap will be this ssh method.

dongreenberg (commented):
Awesome work!!

runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -83,6 +86,11 @@ def __init__(
         self.address = None
         self.client = None
 
+        # K8s related fields
```
carolineechen, line 89: what was the reasoning for fully integrating kubernetes into ondemand cluster? code snippets like this/in the config_for_rns with k8s specific variables or k8s specific branches make it seem like it makes more sense and also pretty straightforward to have KubernetesCluster be a separate class but subclass OnDemandCluster (but could still reuse the same ondemand_cluster factory function).

would not really be user facing though so we could always make this refactor in a separate following PR if this makes sense.

tests/test_resources/test_clusters/test_kubernetes_cluster.py
```diff
@@ -0,0 +1,31 @@
...
+
+@pytest.mark.kubernetestest
+class TestKubernetesCluster(tests.test_resources.test_clusters.test_cluster.TestCluster):
+
+    MAP_FIXTURES = {"resource": "cluster"}
+
+    UNIT = {"cluster": []}
+    LOCAL = {"cluster": []}
+    MINIMAL = {"cluster": ["kubernetes_cpu_cluster"]}
+    THOROUGH = {"cluster": []}
```
carolineechen, line 20: let's include kubernetes_cpu_cluster in thorough and maximal. setting it to [] overrides it and excludes it from those

carolineechen (commented):
a couple of comments, mostly just making sure that existing on-demand behavior hasn't been changed. minimal tests pass for me locally using your kubeconfig.

could you also update the PR description (some portions are out-of-date), and also rebase and run pre-commit and check that the CI tests are passing? thanks!

runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -347,22 +366,26 @@ def up(self):
             )
             task.set_resources(
                 sky.Resources(
-                    cloud=cloud_provider,
+                    cloud=cloud_provider,  # TODO: confirm if passing instance type in old way (without --) works when provider is k8s
+
                     instance_type=self.instance_type
                     if self.instance_type
                     and ":" not in self.instance_type
-                    and "CPU" not in self.instance_type
```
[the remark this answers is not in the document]
dongreenberg: +1, I think this will break passing `instance_type="CPU:2". What we want is to just check if `--` is in instance type (or self.instance_type), in which case we just pass it as is, and if not, use the existing behavior. We should probably just do this (compute the `instance_type` `accelerators` and `cpus`) above instead of inline here.

runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -481,14 +502,24 @@ def ssh(self, node: str = None):
...
-            if not sky_key.exists():
-                raise FileNotFoundError(f"Expected default sky key in path: {sky_key}")
+        if self.provider == "kubernetes":
+            command = f"kubectl get pods | grep {self.name}"
+
+            try:
+
+                output = subprocess.check_output(command, shell=True, text=True)
 
-            subprocess.run(["ssh", "-i", str(sky_key), f"ubuntu@{node}"])
```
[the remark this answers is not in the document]
dongreenberg: +1, we dropped node support here. I would leave the existing logic alone for now in the case where `self.provider != "kubernetes"`, and we should update it in a separate PR anyway because skypilot now has nicer support for sshing into nodes (you can just do `ssh my-cluster-worker1` which is pretty neat, so we could accept an int for node in addition to an ip).

dongreenberg (commented):
Pretty close! Agree with Caroline we need to just make sure this doesn't disrupt the ondemand_cluster flow.

carolineechen:
I think we should also include the extra requirements for kubernetes here https://github.com/run-house/runhouse/blob/d993d33cac80fa1a80b4c937718c72fc2b0d6e55/setup.py#L89

runhouse/resources/hardware/cluster.py
```diff
@@ -56,7 +56,8 @@ class Cluster(Resource):
     SERVER_STOP_CMD = f'pkill -f "{SERVER_START_CMD}"'
     # 2>&1 redirects stderr to stdout
     START_SCREEN_CMD = f"screen -dm bash -c \"{SERVER_START_CMD} 2>&1 | tee -a '{SERVER_LOGFILE}' 2>&1\""
-    RAY_START_CMD = f"ray start --head --port {DEFAULT_RAY_PORT}"
+    START_NOHUP_CMD = f"nohup {SERVER_START_CMD} >> {SERVER_LOGFILE} 2>&1 &"
+    RAY_START_CMD = "ray start --head --port 6379"
```
carolineechen, line 60: can we leave this as `DEFAULT_RAY_PORT`? makes it more clear and also in case this ever gets changed

runhouse/main.py
```diff
@@ -139,9 +139,15 @@ def _start_server(
             f.readlines()  # Discard these, they're from the previous times the server was started
 
         # We do these one by one so it's more obvious where the error is if there is one
-        for cmd in cmds:
+        for i, cmd in enumerate(cmds):
+            last_cmd = i == len(cmds) - 1
             console.print(f"Executing `{cmd}`")
-            result = subprocess.run(shlex.split(cmd), text=True)
+            if (
+                last_cmd
```
carolineechen, line 146: nit: can just do `if i == len(cmds) - 1` and avoid setting the `last_cmd` variable, I think it's pretty self explanatory what the expression means esp with the comment

runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -120,6 +120,16 @@ def config_for_rns(self):
                 "live_state": self._get_sky_state(),
             }
         )
+
+        # if self.provider == "kubernetes":
+        #     config.update(
+        #         {
+        #             "namespace": self.namespace,
```
carolineechen, line 127: is this meant to be removed or kept in?

runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -356,6 +366,34 @@ def _update_from_sky_status(self, dryrun: bool = False):
...
+        if (
+            self.instance_type
+            and ":" in self.instance_type
+            and "CPU" not in self.instance_type
+        ):
+            return self.instance_type
+
+        return None
+
+    def compute_cpus(self):
```
carolineechen, line 387: nit
```suggestion
    def num_cpus(self):
```

runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -356,6 +366,34 @@ def _update_from_sky_status(self, dryrun: bool = False):
...
 
+    def compute_instance_type(self):
+        if "--" in self.instance_type:  # K8s specific syntax
+            return self.instance_type
+        elif ":" not in self.instance_type and "CPU" not in self.instance_type:
+            return self.instance_type
+
+        return None
+
+    def compute_accelerators(self):
```
carolineechen, line 377: nit
```suggestion
    def accelerators(self):
```

runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -356,6 +366,34 @@ def _update_from_sky_status(self, dryrun: bool = False):
         cluster_dict = self.status(refresh=not dryrun)
         self._populate_connection_from_status_dict(cluster_dict)
 
+    def compute_instance_type(self):
```
carolineechen, line 369:
```suggestion
    def get_instance_type(self):
```

runhouse/resources/hardware/utils.py
```diff
@@ -571,4 +571,4 @@ def ssh_tunnel(
 
     # ssh_tunnel should certainlly be non-None at this point.
     cache_open_tunnel(address, ssh_port, ssh_tunnel)
-    return ssh_tunnel
+    return ssh_tunnel, local_port
```
carolineechen, line 574: regarding the current testing errors, although you change it to `return ssh_tunnel, local_port` here, there are other break out points above where only `ssh_tunnel` is being returned, and it is likely running into one of those paths and unable to split up `ssh_tunnel` as a tuple

runhouse/resources/hardware/on_demand_cluster.py
dongreenberg: Let's delete it

dongreenberg (commented):
Couple nits but looking very close, I think we can merge as soon as the tunnel stuff is worked out.

jlewitt1:
test

jlewitt1:
test

jlewitt1:
testing

jlewitt1:
test

PR #110, opened 2023-09-27
Support an optional `host` parameter for the start and restart commands.

DenisYay:
Example usages:
runhouse start --custom_server_host 0.0.0.0
runhouse restart --custom_server_host 0.0.0.0
python -m runhouse.servers.http.http_server --host 0.0.0.0

### DIFF
--- a/runhouse/main.py
+++ b/runhouse/main.py
@@ -91,9 +91,7 @@ def load_cluster(cluster_name: str):
         c._update_from_sky_status(dryrun=True)
 
 
-def _start_server(
-    restart, restart_ray, screen, create_logfile=True, custom_server_host=None
-):
+def _start_server(restart, restart_ray, screen, create_logfile=True, host=None):
     from runhouse.resources.hardware.cluster import Cluster
 
     cmds = Cluster._start_server_cmds(
@@ -101,7 +99,7 @@ def _start_server(
         restart_ray=restart_ray,
         screen=screen,
         create_logfile=create_logfile,
-        custom_server_host=custom_server_host,
+        host=host,
     )
 
     try:
@@ -124,9 +122,9 @@ def _start_server(
 def start(
     restart_ray: bool = typer.Option(False, help="Restart the Ray runtime"),
     screen: bool = typer.Option(False, help="Start the server in a screen"),
-    custom_server_host: Optional[str] = typer.Option(
+    host: Optional[str] = typer.Option(
         None,
-        help="Custom server host address e.g. 0.0.0.0. Default is `None` and the server would start on 127.0.0.1.",
+        help="Custom server host address e.g. 0.0.0.0. Default is `None` and the server would start on 127.0.0.1",
     ),
 ):
     """Start the HTTP server on the cluster."""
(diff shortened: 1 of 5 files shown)


runhouse/rns/defaults.py
```diff
@@ -26,7 +26,7 @@ class Defaults:
         "default_autostop": -1,
         "use_spot": False,
         "use_local_configs": True,
-        "disable_data_collection": False,
+        "disable_data_collection": True,
```
jlewitt1, line 29: thoughts about leaving it on by default? otherwise we would not be getting the sentry error reports / cluster usage data

DenisYay: Makes sense! I was using it for local Docker testing.

runhouse/main.py
```diff
@@ -121,10 +124,18 @@ def _start_server(restart, restart_ray, screen, create_logfile=True):
 def start(
     restart_ray: bool = typer.Option(False, help="Restart the Ray runtime"),
     screen: bool = typer.Option(False, help="Start the server in a screen"),
+    custom_server_host: Optional[str] = typer.Option(
```
dongreenberg, line 127: This is code golf but I think I'd prefer for this to just be `host`

runhouse/main.py
```diff
@@ -140,6 +151,10 @@ def restart(
         False,
         help="Resync the Runhouse package. Only relevant when restarting remotely.",
     ),
+    custom_server_host: Optional[str] = typer.Option(
```
dongreenberg, line 154: Ditto `host` comment above.

runhouse/resources/hardware/cluster.py
```diff
@@ -480,7 +482,13 @@ def _start_server_cmds(cls, restart, restart_ray, screen, create_logfile):
                 Path(cls.SERVER_LOGFILE).touch()
             cmds.append(cls.START_SCREEN_CMD)
         else:
-            cmds.append(cls.SERVER_START_CMD)
+            server_start_cmd = cls.SERVER_START_CMD
+            if custom_server_host is not None:
+                server_start_cmd += f" --host {custom_server_host}"
```
dongreenberg, line 487: I'm a bit confused - above it looks like this would need to be `--custom_server_host` but here it's just `--host`?

dongreenberg: Oh oh http_server.py takes `--host`, not the CLI, got it.

dongreenberg (approved):
Looks great, only note is changing the arg in the CLI to `--host`.

runhouse/main.py
DenisYay: Done

runhouse/resources/hardware/cluster.py
DenisYay: Yep! Renamed all to host.

runhouse/main.py
DenisYay: Done

DenisYay:
Apologies @dongreenberg , it looks like I flipped you back to 'Requested review' after you've already approved.

dongreenberg (approved):

jlewitt1 (approved):

PR #111, opened 2023-09-27
Fix file .to bugs

carolineechen:
Fixes two bugs in the current `.to(cluster, dest_path)` function for File:
* we set the folder of the `dest_path` to have the path of the `dest_path`, rather than the parent of `dest_path`
* we send the folder to `cluster`, which includes all contents of the file's parent folder, rather than the file alone

The intended goal of `.to(cluster, dest_path)` is to write the contents of the local file to the `dest_path` file.

This can further be improved in the future by directly syncing over files without needing to hold the contents in Python memory.

### DIFF
--- a/runhouse/resources/blobs/file.py
+++ b/runhouse/resources/blobs/file.py
@@ -131,22 +131,25 @@ class File(Blob):
             return new_blob
 
         new_file = copy.copy(self)
-        new_file.local._folder = self._folder.to(
-            system=system, path=path, data_config=data_config
+        new_file.local._folder = folder(
+            path=str(Path(path).parent) if path is not None else path,
+            system=system,
+            data_config=data_config,
         )
+        new_file.write(self.fetch(mode="r"), serialize=False)
         return new_file
 
-    def resolved_state(self, deserialize: bool = True):
+    def resolved_state(self, deserialize: bool = True, mode: str = "rb"):
         """Return the data for the user to deserialize. Primarily used to define the behavior of the ``fetch`` method.
 
         Example:
             >>> data = file.fetch()
         """
-        self.local._cached_data = self._folder.get(self._filename)
+        self.local._cached_data = self._folder.get(self._filename, mode=mode)
         if deserialize:
             try:
                 deserialized_data = pickle.loads(self._cached_data)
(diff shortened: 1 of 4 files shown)


runhouse/resources/blobs/file.py
```diff
@@ -131,9 +131,12 @@ def to(
...
 
         new_file = copy.copy(self)
-        new_file.local._folder = self._folder.to(
-            system=system, path=path, data_config=data_config
+        new_file.local._folder = folder(
+            path=str(Path(path).parent) if path is not None else path,
+            system=system,
+            data_config=data_config,
         )
+        new_file.write(self.fetch())
```
carolineechen, line 139: @dongreenberg calling `self.fetch()` returns the file contents in byte form rather than the original raw text, what do you think is the best way to achieve copying over the file contents or even the file itself?

dongreenberg (approved):

carolineechen:
still running into a bug when using secrets with this change, gonna hold off on merging til I figure that out + do some unit testing

runhouse/resources/blobs/file.py
```diff
@@ -130,23 +129,26 @@ def to(
...
             return new_blob
 
-        new_file = copy.copy(self)
-        new_file.local._folder = self._folder.to(
-            system=system, path=path, data_config=data_config
-        )
+        new_file = file(path=path, system=system, data_config=data_config)
+        try:
+            self._folder.get(self._filename, mode="r")
+            new_file.write(self.fetch(mode="r"), serialize=False, mode="w")
```
dongreenberg, line 135: You don't need to pass `serialize=False` into the fetch too?

runhouse/resources/blobs/file.py
```diff
@@ -130,23 +129,26 @@ def to(
             new_blob.data = data_backup
             return new_blob
 
-        new_file = copy.copy(self)
-        new_file.local._folder = self._folder.to(
-            system=system, path=path, data_config=data_config
-        )
+        new_file = file(path=path, system=system, data_config=data_config)
+        try:
+            self._folder.get(self._filename, mode="r")
```
dongreenberg, line 134: What does this get do?

PR #116, opened 2023-10-03
Check and warn if local and remote server versions don't match

carolineechen:
Throw a warning if runhouse versions on remote and local don't match. Keep track of and update remote runhouse version every time the server is (re)started, as a simple, BC friendly approach (checking for client rh version is different because it is not being restarted/updated; and checking on the server is finicky and potentially BC-breaking on the server side)

### DIFF
--- a/runhouse/resources/hardware/cluster.py
+++ b/runhouse/resources/hardware/cluster.py
@@ -69,6 +69,7 @@ class Cluster(Resource):
         self.ips = ips
         self._rpc_tunnel = None
         self.client = None
+        self._rh_version = None
 
     def save_config_to_cluster(self):
         import json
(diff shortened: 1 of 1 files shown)


runhouse/resources/hardware/cluster.py
```diff
@@ -385,6 +385,18 @@ def check_server(self, restart_server=True):
                 logger.info(f"Checking server {self.name}")
                 self.client.check_server(cluster_config=cluster_config)
                 logger.info(f"Server {self.name} is up.")
+
+                import runhouse
```
carolineechen, line 389: ideas for checking version are `import runhouse; runhouse.__version__` and `pip freeze | grep runhouse`, I think the first is faster?

a bit concerned on latency though, since `check_server` is called quite frequently (most cluster calls will pass through here), does it make sense to just call this when we instantiate a cluster and if it's already up, then check?

jlewitt1: What if we run the check whenever the HTTP server is started / restarted (in the constructor)? Or could the mismatch still exist in other spots that we would miss?

carolineechen: from discussion today: could consider caching the runhouse version as a global variable on remote server when server is started, and compare local runhouse version with that. local import and version calls should be quick if runhouse has already been imported.

dongreenberg (approved):

PR #117, opened 2023-10-04
Secrets as resource

carolineechen:
Remove existing secrets class, instead implementing secrets as a resource, except one that is saved into Vault (instead of RNS) for security purposes. Includes subresource implementation for provider secret type. (Cluster and Env secrets are new concepts addressed in a separate PR)

### Secrets Workflow
* User constructs a Secret resource, generally in one of the following ways. The `.values` property of the resource conveniently extract the values from the object, file, or env (python or .env file).
```
# custom secret -- instantiate with secrets values or path
custom_secret = rh.secret('my_secret', values={"key": "val"})
custom_secret = rh.secret('my_secret', path="values.json")

# passing in provider name; will load values from default path
aws_secret = rh.provider_secret(‘aws’)
aws_secret.values   # {'access_key': 'XXXX', 'secret_key': 'xxxxx'}

# passing in provider name and values, in-memory resource
lamdba_secret = rh.provider("lambda", values={"api_key": "xxxxx"})
```
* To save the secret config (to Vault/local). By default, secret values are saved, but can also save just the metadata by setting values=False.
```
custom_secret.save()
aws_secret.save(values=True/False)
```
* To write down a secret, if it is an in-memory secret. By default, file overriding is not supported, as we don't want to permanently delete/override existing secrets on user's dev environment. Instead, recommend them to manually update the path+secret, or can set overwrite param to True. If a path different from the secret path is given, will return a copy of the original secret, but with the updated path.
```
lamdba_secret.write()  # writes down to default lambda path ~/.lambda_cloud/lambda_keys
lambda_secret.write("~/.lambda_cloud/lambda_keys_2")
```
* To send the secret to a cluster, including optionally writing it down to a remote file (True by default).
```
remote_secret = aws_secret.to(cluster, path=aws_secret.path)
# or, remote_secret = aws_secret.to(cluster).write()
remote_secret.path   # remote File resource where the secrets are saved
remote_secret.values  # {'access_key': 'XXXX', 'secret_key': 'xxxxx'}
```
* To login and write down the secrets (requires having saved the values, otherwise can only extract the metadata). This will also store the path the secrets are stored in the runhouse config.
```
rh.login(download_secrets=True)
```
* To logout. This goes through the local (?) and previously downloaded secrets (with paths stored in runhouse config), and prompts the user through each one to confirm deletion.
```
rh.logout(delete_loaded_secrets=True)
```

Separate PR -- Env + Cluster Secrets and Workflows
```
pipeline_secrets = ["aws", "openai", "env"]
rh.env(reqs=[...], secrets=pipeline_secrets)  # will set up secrets on the servelet code is run on; replace "env_vars" key

cluster_secret = rh.cluster_secret(values=config_dict)
rh.cluster("rh-a10x", secrets=cluster_secret)
rh.cluster.share("teammate@gmail.com")  # also shares the cluster secrets with them to be able to access cluster
```

### TODO for this PR
- Some refactoring/cleanup
- Solidifying Secret class methods / utils
- Fix test_login

### Separate/Futures PRs
- rns_server repo: Update Vault saving stuff
- Extend “path” to paths – ex/ aws may require multiple files containing different info
- Env + Cluster Secrets
- Secrets as part of rh.cluster or cluster.share()
- Adding secrets to rh.env, e.g. env_var secrets
- [?] Is_enabled/from_provider support; using provider APIs to check if keys/secrets are valid
- Additional providers support – WandB, OpenAI, Databricks, Anthropic

### takeaways from discussion Oct 4
* secrets param is confusing, change to values (and by default save these values)
* don't override any secrets files, throw error
* additional_files param, along w/ mapping in .to
* cluster + ssh secrets discussion, design tbd
* things to consider in the future -- versioning (vault supports 10 latest), sharing retention/invalidation


runhouse/resources/hardware/cluster.py
```diff
@@ -914,21 +914,43 @@ def run_python(
...
 
-    def sync_secrets(self, providers: Optional[List[str]] = None):
+    def sync_secrets(
+        self, secrets: Optional[List[str or "Secret"]] = None, file: bool = True
+    ):
         """Send secrets for the given providers.
 
         Args:
-            providers(List[str] or None): List of providers to send secrets for.
+            providers(List[str] or None): List of providers to send and write down secrets for.
```
DenisYay, line 923: /s/providers/secrets?

runhouse/resources/hardware/cluster.py
```diff
@@ -914,21 +914,43 @@ def run_python(
 
         return return_codes
 
-    def sync_secrets(self, providers: Optional[List[str]] = None):
+    def sync_secrets(
+        self, secrets: Optional[List[str or "Secret"]] = None, file: bool = True
+    ):
         """Send secrets for the given providers.
 
         Args:
```
DenisYay, line 922: Should we add a comment about the `file` argument?

runhouse/resources/secrets/provider_secrets/ssh_secret.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def delete_file(
+        self,
+        path: str = None,
+    ):
+        logger.warning(
+            "Deleting SSH files is not currently supported. Please manually delete them."
+        )
+
+        # priv_key_file = path or self.path
```
DenisYay, line 99: Should we remove this section?

carolineechen:
Closing in favor of #135 and splitting up Secrets work into more iterative components

PR #125, opened 2023-10-26
Auth Cluster Cache

jlewitt1:
**Move the den auth / resource validation out of the HTTP server and into the object store**

Make `AuthCache` into an actor on the cluster which holds a basic cache. 

The `AuthCache` class itself is the actor: when a new request comes in with a token, pass in to the actor's `add_user` method which loads down resources it has access to and stores it in the cache.

When functions are called on the cluster, validate token (if den auth is enabled) and first check cache to validate the user's access. 

------

Continue using a validator decorator (`validate_cluster_access`) to validate the user's access to the cluster if den auth is enabled before continuing with API logic. If the user has write access to the cluster will be given access to all resources on the cluster by default. 

For function calls must have read or write access to the resource.

### DIFF
--- a/runhouse/servers/http/http_server.py
+++ b/runhouse/servers/http/http_server.py
@@ -16,18 +16,16 @@ from fastapi.responses import JSONResponse, StreamingResponse
 
 from sky.skylet.autostop_lib import set_last_active_time_to_now
 
-from runhouse.globals import configs, env_servlets, obj_store, rns_client
-from runhouse.resources.hardware.utils import (
-    _current_cluster,
-    _get_cluster_from,
-    _load_cluster_config,
-)
-from runhouse.rns.utils.api import resolve_absolute_path, ResourceAccess
+from runhouse.globals import configs, env_servlets, rns_client
+from runhouse.resources.hardware.utils import _load_cluster_config
+from runhouse.rns.utils.api import resolve_absolute_path
 from runhouse.rns.utils.names import _generate_default_name
+from runhouse.servers.http.auth import verify_cluster_access
 from runhouse.servers.http.certs import TLSCertConfig
 from runhouse.servers.http.http_utils import (
     b64_unpickle,
     get_token_from_request,
+    load_current_cluster,
     Message,
     OutputType,
     pickle_b64,
@@ -44,8 +42,9 @@ app = FastAPI()
 global den_auth
 
 
-def validate_user(func):
-    """If using Den auth, validate the user's Runhouse token and access to the resource before continuing."""
+def validate_cluster_access(func):
+    """If using Den auth, validate the user's Runhouse token and access to the cluster before continuing.
+    if calling a function, ensure user has read or write access to the cluster."""
 
     @wraps(func)
     async def wrapper(*args, **kwargs):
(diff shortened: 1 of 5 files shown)


runhouse/servers/http/http_server.py
```diff
@@ -659,9 +575,10 @@ async def call(
     ):
         kwargs = args.get("kwargs", {})
         args = args.get("args", [])
+        token = get_token_from_request(request)
```
dongreenberg, line 578: We should hash the token here so the only thing that's passed around / through the servlet is the hash, not the token itself. The AuthCache's lookup should accept the hash, not the token.

runhouse/servers/servlet.py
```diff
@@ -570,9 +594,19 @@ def add_secrets(self, message: Message):
...
+        module,
+        method=None,
+        args=None,
+        kwargs=None,
+        serialization="json",
+        token=None,
+    ):
         self.register_activity()
         module = obj_store.get(module)
+        self.validate_resource_access(module, token)
```
dongreenberg, line 608: I think it's fine to do this here for now, but most likely I'd move it into the obj_store's `get` method later. Also, it'd probably be cleaner for most of the auth logic to just live in the object store, rather than split it between the obj store and servlet.

runhouse/servers/obj_store.py
```diff
@@ -58,6 +84,19 @@ def call_kv_method(store, method, *args, **kwargs):
         else:
             return getattr(store, method)(*args, **kwargs)
 
+    def get_access_level(self, token: str, resource_uri: str):
+        return self.call_auth_cache_method(
+            self._auth_cache, "lookup_access_level", token, resource_uri
+        )
+
+    def add_user(self, token: str):
```
dongreenberg, line 92: I think we want to call add_user from inside the http_server, so we don't need to pass the token through. You can load the AuthCache actor by name and call add_user on it directly with ray.remote and ray.get (you don't need the helper method or anything like that).

runhouse/servers/obj_store.py
```diff
@@ -46,6 +48,30 @@ def set_name(self, servlet_name: str):
...
+                name="auth_cache",
+                get_if_exists=True,
+                lifetime="detached",
+                namespace="runhouse",
+            )
+            .remote()
+        )
+
+    @staticmethod
+    def call_auth_cache_method(store, method, *args, **kwargs):
```
dongreenberg, line 64: Probably not necessary - I used a helper method like this for the servlet Actors because at one time some were actors and some weren't (and actually now they're all actors). In this case the AuthCache is always an actor, so we can just call methods on it with get and remote directly.

runhouse/servers/http/http_utils.py
```diff
@@ -54,6 +55,23 @@ def b64_unpickle(b64_pickled):
     return pickle.loads(codecs.decode(b64_pickled.encode(), "base64"))
 
 
+def get_token_from_request(request):
+    auth_headers = request.headers.get("Authorization", "")
+    return auth_headers.split("Bearer ")[-1] if auth_headers else None
+
+
+def hash_token(token: str) -> str:
```
dongreenberg, line 63: Move into auth.py?

runhouse/servers/http/http_server.py
jlewitt1: Hmm any ideas then on how we can refresh the cache? Bc we need to pass a token into the request to den

PR #130, opened 2023-11-02
Introduce test levels

dongreenberg:
First major step in testing overhaul. Introduces test levels which allow combinations of fixtures to be set for different levels of testing rigor in different test suites. Also unblocks local docker clusters for testing.

- Install and start runhouse outside dockerfile to copy cluster config and test setup more thoroughly
- Add screen to slim dockerfiles to avoid having to use "--no-screen"
- Add default fixture levels to conftest, and module levels override
- Give keypair and pwd docker images different tags
- Migrate test_module, test_function, and test_obj_store to use `cluster` fixture. Obj store and module tests mostly pass with keypair, and fail for pwd (but pass with pexpect changes in a separate PR)


tests/conftest.py
```diff
@@ -17,24 +18,167 @@
...
 
-dotenv.load_dotenv()
+SSH_USER = "rh-docker-user"
+
+
+class TestLevels(str, enum.Enum):
+    UNIT = "unit"
+    LOCAL = "local"
+    MINIMAL = "minimal"
+    THOROUGH = "thorough"
```
carolineechen, line 28: qq: what are you thinking for the difference of thorough vs maximal?

dongreenberg: Thorough means the complete set of fixtures we'd want to run to have thoroughly tested this code, e.g. what we'd run before release if we changed something significantly. Maximal is essentially every possible fixture that could be run, even superfluous ones (e.g. all the GPU ones). I don't foresee us running maximal practically ever, but it's there just in case. I'll put this into a testing README in another PR (with a few other testing features I'd like to add).

tests/conftest.py
```diff
@@ -47,10 +49,116 @@ def pytest_generate_tests(metafunc):
...
+                else f"RUNHOUSE_VERSION={rh_version}",
+                "-t",
+                f"runhouse:{image_name}",
+                ".",
+            ]
+        else:
+            raise ValueError("No keypath or password file path provided")
+
+        print(shlex.join(build_cmd))
+        # run_shell_command(subprocess, build_cmd, cwd=str(rh_parent_path.parent))
```
dongreenberg: @jlewitt1 is this a prank

tests/conftest.py
```diff
@@ -86,24 +198,65 @@ def test_account():
         configs.set("default_folder", f"/{current_username}")
 
 
-def load_and_share_resources(username_to_share):
+def load_and_share_resources(username_to_share, test_level):
     # Create the shared cluster using the test account
-    dotenv.load_dotenv()
+    if test_level in [TestLevels.UNIT, TestLevels.LOCAL]:
```
dongreenberg, line 203: I'm not sure there's a reason to do this on a remote cluster for higher test levels. In general we can test sharing and access from the test account in a local container, and if we find there are edge cases where the behavior is different from a cluster we can change it then. Also, I think this code should be going in a `test_account_loggedin_cluster` fixture which can be used by many tests. Actually I think there are a few rearrangements I would make here, let's chat about it on Mon.

dongreenberg (commented):
@jlewitt1 changes make sense, but I think I would have broken out the test account cluster stuff into a diff PR so we can talk about them separately, I think there are a few options here.

PR #131, opened 2023-11-03
Use Pexpect for password clusters instead of paramiko and fsspec

dongreenberg:
Uses pexepct to direct passwords to the console for clusters which require them, eliminating fsspec and paramiko from the connection flow, and also giving us a printable string to debug password cluster connection issues with. Everything now flows through the SkyPilot SSHCommandRunner, which I've also updated our fork of.

Also adds an attempt at a simple password test suite. Tests pass for docker and cpu password clusters.

...I have no idea what's going on with the commits. Help.

### DIFF
--- a/runhouse/resources/hardware/cluster.py
+++ b/runhouse/resources/hardware/cluster.py
@@ -27,6 +27,7 @@ from runhouse.resources.hardware.utils import (
     _current_cluster,
     ServerConnectionType,
     SkySSHRunner,
+    SshMode,
 )
 from runhouse.resources.resource import Resource
 
@@ -916,7 +917,10 @@ class Cluster(Resource):
 
             if up:
                 ssh_command = runner.run(
-                    ["mkdir", "-p", dest], stream_logs=False, return_cmd=True
+                    ["mkdir", "-p", dest],
+                    stream_logs=False,
+                    return_cmd=True,
+                    ssh_mode=SshMode.INTERACTIVE,
                 )
                 ssh = pexpect.spawn(ssh_command, encoding="utf-8")
                 ssh.logfile_read = sys.stdout
@@ -928,6 +932,7 @@ class Cluster(Resource):
                 ssh.close()
             else:
                 Path(dest).expanduser().parent.mkdir(parents=True, exist_ok=True)
+
             rsync_cmd = runner.rsync(
                 source,
                 dest,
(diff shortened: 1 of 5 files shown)


runhouse/resources/hardware/cluster.py
```diff
@@ -948,24 +898,61 @@ def _rsync(
...
+                ssh_command = runner.run(
+                    ["mkdir", "-p", dest],
+                    stream_logs=False,
+                    return_cmd=True,
+                    ssh_mode=SshMode.INTERACTIVE,
+                )
+                ssh = pexpect.spawn(ssh_command, encoding="utf-8")
+                ssh.logfile_read = sys.stdout
+                # If CommandRunner uses the control path, the password may not be requested
+                next_line = ssh.expect(["assword:", pexpect.EOF])
```
carolineechen, line 929: not sure if typo or meant for some regex match

runhouse/resources/hardware/utils.py
```diff
@@ -238,22 +294,110 @@ def rsync(
...
-        if filter_options:
-            cr.RSYNC_FILTER_OPTION += f" --filter='{filter_options}'"
-        super().rsync(
-            source,
-            target,
-            up=up,
-            log_path=log_path,
-            stream_logs=stream_logs,
-            max_retry=max_retry,
+        """Uses 'rsync' to sync 'source' to 'target'.
```
carolineechen, line 303: I think there's functionality in .run() that makes it hard to directly reuse skypilot code without also porting it over, but think for this one if it's just filter options we can maybe leave it as is? If skypilot modifies this function on their end, I assume we'd generally want to automatically inherit those changes too and not have to reduplicate updates/have two sources of truth.

carolineechen (commented):
super cool!

runhouse/resources/hardware/cluster.py
dongreenberg: Intended, matches both "Password:" and "password:"

runhouse/resources/hardware/utils.py
dongreenberg: Even though we don't use the filter options anymore, we still modify this one with the return_cmd argument, which we need for pexpect. It's not ideal but I don't see a way around it. We can also try just contributing this option directly up to SkyPilot.

PR #133, opened 2023-11-04
Remove deprecated pytest decorator

jlewitt1:
Use `default_fixtures` based on the provided test level instead of the `cpu_clusters` decorator

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -557,9 +557,20 @@ def sm_source_dir(sm_entry_point, pytorch_training_script):
 
 @pytest.fixture(scope="session")
 def cluster(request):
-    """Parametrize over multiple fixtures - useful for running the same test on multiple hardware types."""
-    # Example: @pytest.mark.parametrize("cluster", [v100_gpu_cluster, k80_gpu_cluster], indirect=True)"""
     return request.getfixturevalue(request.param.__name__)
+
+
+@pytest.fixture(
+    params=[
+        "ondemand_cpu_cluster",
+        "v100_gpu_cluster",
+        "k80_gpu_cluster",
+        "a10g_gpu_cluster",
+    ],
+    ids=["cpu", "v100", "k80", "a10g"],
+)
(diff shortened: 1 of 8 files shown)


tests/test_package.py
```diff
@@ -277,11 +271,6 @@ def test_torch_install_command_generator():
 
 
 @pytest.mark.gputest
-@pytest.mark.parametrize(
-    "cluster",
-    ["ondemand_cpu_cluster", "v100_gpu_cluster", "k80_gpu_cluster", "a10g_gpu_cluster"],
-    indirect=True,
-)
```
dongreenberg, line 284: I don't think we should be removing these here?

jlewitt1: added a new on demand fixture for these package tests, issue was we had duplicate `cluster` fixtures which were causing pytest errors

jlewitt1: One open question is how to incorporate the gpus used here into the default fixtures. @dongreenberg Wdyt about not adding them to the default fixtures, but leaving them as part of the gputest marker where they would only be provisioned if tests are explicitly run with the `gputest` marker?

tests/conftest.py
```diff
@@ -557,11 +557,22 @@ def sm_source_dir(sm_entry_point, pytorch_training_script):
...
+    params=[
+        "ondemand_cpu_cluster",
+        "v100_gpu_cluster",
+        "k80_gpu_cluster",
+        "a10g_gpu_cluster",
+    ],
+    ids=["cpu", "v100", "k80", "a10g"],
+)
+def ondemand_cluster(request):
+    return request.getfixturevalue(request.param)
```
dongreenberg, line 573: I don't think we need this, ideally I'd like to get rid of having to launch all these just to test the torch installs. We should try to break that test up to be more unity (or use mocks) so we can get rid of this, it's too crazy to launch all these gpus.

dongreenberg (approved):

PR #135, opened 2023-11-06
Convert Secrets to RH resource

carolineechen:
Convert Secret(s) class to a RH resource, to make saving, sending, sharing easier and more intuitive with the rest of RH API.

This PR aims to reproduce the current functionality of secrets, but as a resource type. This includes
* Base/Custom secrets class
* Provider secrets class, and support for current builtin providers (aws, azure, gcp, lambda, ssh/sky)
* Login/Logout using secret resources
* Docs and accessibility tutorial updates with new APIs

BC-breaking changes
* is_enabled
* Secrets.put --> resource.save()
* Secrets.download_into_env --> resource.write()


DenisYay (approved):

tests/test_login.py
```diff
@@ -5,34 +5,35 @@
...
+    ).save(headers=headers)
+
+    rh.provider_secret(
+        name="/snowflake",
+        provider="snowflake",
+        values={"token": "ABCD1234"},
+    ).save(headers=headers)
 
 
 @pytest.mark.logintest
```
dongreenberg, line 39: At some point I'd love to move these tests inside a container so we can run them every time as normal unit tests without worrying about what's going on in the runner's filesystem.

dongreenberg (approved):
This is a monster PR, amazing work. I think we should land it but in a followup PR, we should really isolate everything so it has no side effects on the user's filesystem, and doesn't require any special files to be present. For example, the tests expected a file called "file_secret.json" to be in my cwd, but it should be in a test_secrets/assets folder. Also, my aws secrets were overwritten with "abc123"! But the tests passed, so that's good 😄.

runhouse/resources/secrets/secret.py
```diff
@@ -47,6 +47,17 @@ def __init__(
     def values(self):
         return self._values
 
+    @property
+    def config_for_rns(self):
```
jlewitt1, line 51: @carolineechen do we still need this? if we are removing it from the config later on before sending the request to Den

carolineechen: I added this back to better handle in-memory secrets, which aren't associated with a file or env vars. (1) for reconstructing a resource, the values will be lost and won't allow for `Secret.from_config(secret.config_for_rns)` (2) simplifies the logic for `put_resource` on a cluster, works directly rather than needing to pull secrets from vault

jlewitt1: ah ok makes sense!

carolineechen: (2) is easily addressable, but (1) seems like a property of resource that ideally we can maintain. the resource test `test_from_config` will fail if we don't have "values" in the config

PR #137, opened 2023-11-07
Major test refactor

dongreenberg:
- Reorganize tests into directories, including fixtures
- Allow level fixture to be overriden by a pytest class, in addition to module
- Add test_resource (including TestResource) to begin Test class hierarchy for resources
- Add logged out and test account logged in docker clusters, rearrange sharing and docker fixtures a bit

test_obj_store.py and test_module.py both pass, test_resource.py runs properly (but various failures) - basics seem to work.


runhouse/__init__.py
```diff
@@ -30,7 +30,6 @@
     ipython,
     load,
     locate,
-    resources,
```
dongreenberg, line 33: FYI: Not used and collides with rh.resources to import Resource.

runhouse/servers/http/http_server.py
```diff
@@ -204,6 +204,8 @@ def get_cert():
     def check_server():
         try:
             HTTPServer.register_activity()
+            if not ray.is_initialized():
+                raise Exception("Ray is not initialized, restart the server.")
```
dongreenberg, line 208: FYI: was observing failures where Ray didn't start properly which were not being caught to trigger restart_server. Just doing this is enough to make sure it's triggered (and takes practically no time at all).

PR #139, opened 2023-11-09
AWS Lambda initial POC

BelSasha:
What is supported?

1.  Create and execute a lambda (one or multiple .py files), including installing python libraries.
2. Function call methods: map, starmap, save, share, from_name, for_config
3. A factory method, which creates AWSLambdaFunction instance, and sends it to AWS.
4. Document (rst file) was created. 

Tests which were executed and passed:
1. AWS Lambda Function tests (all testing suite)
2. Function tests (minimal level)
3. Module tests (minimal level)


dongreenberg (commented):
Overall this is amazing work and a solid PoC. Next I think we should discuss the API surface and how to make it smoother with the core Function API, the access details, and some of the packaging mechanics.

runhouse/__init__.py
```diff
@@ -20,6 +20,7 @@
 from runhouse.resources.provenance import capture_stdout, Run, run, RunStatus, RunType
 from runhouse.resources.queues import Queue
 from runhouse.resources.resource import Resource
+from runhouse.resources.serverless import aws_lambda_function, AWSLambdaFunction
```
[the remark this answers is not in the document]
BelSasha: done

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,845 @@
...
+    ]
+    MAX_WAIT_TIME = 60  # seconds, max time that can pass before we raise an exception that AWS update takes too long.
+    DEFAULT_REGION = "us-east-1"
+    DEFAULT_RETENTION = 30  # one month, for lambdas log groups.
+    DEFAULT_TIMEOUT = 900  # sec. meaning 15 min.
+    DEFAULT_MEMORY_SIZE = 1024  # MB
+    DEFAULT_TMP_SIZE = 3072  # MB, meaning 3G
+
+    # setup needs to be here and not in __init__ for the reload option.
+    if Path(CRED_PATH).is_file():
```
jlewitt1, line 58: Doesn't this mean that whenever we reload a func it's region will always be the default? what if instead of relying on the `LAMBDA_CLIENT` for reloading we used Den?

tests/test_resources/test_modules/test_functions/test_aws_lambda/test_aws_lambda.py
```diff
@@ -0,0 +1,554 @@
...
+    res2 = config_lambda(8, 12)
+    res3 = config_lambda(14, 17)
+
+    assert res1 == "3"
+    assert res2 == "20"
+    assert res3 == "31"
+    LAMBDAS_NAMES.add(config_lambda.name)
+
+
+def test_share_lambda():
```
jlewitt1, line 529: We can make this test more meaningful by creating and saving the lambda with the test account, then reloading and running with the user's account

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,845 @@
...
+                reqs=[], env_vars={"HOME": "/tmp"}, name=Env.DEFAULT_NAME
+            )
+        else:
+            config["env"] = Env.from_config(config["env"])
+
+        return AWSLambdaFunction(**config, dryrun=dryrun).to()
+
+    @classmethod
+    def from_name(cls, name, dryrun=False, alt_options=None):
+        funcs = cls.LAMBDA_CLIENT.list_functions()
```
jlewitt1, line 141: Expanding here on my previous comment, but another reason to have this load from Den and not the lambda client is to be able to support reloading shared functions.

This also means that we need to parse `name` here, since it can represent an rns address

tests/test_resources/test_modules/test_functions/test_aws_lambda/test_aws_lambda.py
```diff
@@ -0,0 +1,554 @@
...
+    basic_func = rh.aws_lambda_function(name="test_lambda_create_and_run")
+    users = ["josh@run.house"]
+    added_users, new_users = basic_func.share(
+        users=users, notify_users=True, access_type="write"
+    )
+    assert added_users == {}
+    assert new_users == {}
+
+
+def test_remove_resources():
```
jlewitt1, line 539: do we have functionality / testing around deleting the function? (similar to what we do with clusters with `teardown_and_delete`, where we delete the resources themselves from lambdas along with the config from RNS

tests/test_resources/test_modules/test_functions/test_aws_lambda/test_aws_lambda.py
```diff
@@ -0,0 +1,554 @@
...
+
+CRED_PATH = f"{Path.home()}/.aws/credentials"
+DEFAULT_REGION = "us-east-1"
+
+if Path(CRED_PATH).is_file():
+    LAMBDA_CLIENT = boto3.client("lambda")
+else:
+    LAMBDA_CLIENT = boto3.client("lambda", region_name=DEFAULT_REGION)
+IAM_CLIENT = boto3.client("iam")
+LAMBDAS_NAMES = set()
```
jlewitt1, line 23: Do we still need this? We should allow for running each test standalone, and not rely on maintaining a global set of active functions. We could automatically teardown the function at end of a test (or add a flag for that, similar to what we do with running docker containers in detached mode)

tests/test_resources/test_modules/test_functions/test_aws_lambda/test_aws_lambda.py
```diff
@@ -0,0 +1,554 @@
...
+
+
+def test_create_and_run_no_layers():
```
jlewitt1, line 26: I needed to add `env=["httpx"]` to prevent a`ModuleNotFoundError`

BelSasha: Doesn't reproduce on my local machine, maybe should check if it reproduces for others.

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,894 @@
...
+        if not config:
+            raise ValueError(f"Could not find a Lambda called {name}.")
+        return cls.from_config(config)
+
+    # --------------------------------------
+    # Private helping methods
+    # --------------------------------------
+
+    @classmethod
+    def _check_for_child_configs(cls, config):
```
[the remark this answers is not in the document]
BelSasha: indeed redundant after from_name refactor. Removing.

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,889 @@
...
+    tmp_size: Optional[int] = None,
+    retention_time: Optional[int] = None,
+    dryrun: bool = False,
+):
+    """Builds an instance of :class:`AWSLambdaFunction`.
+
+    Args:
+        fn (Optional[Callable]): The Lambda function to be executed.
+        paths_to_code: (Optional[list[str]]): List of the FULL paths to the python code file(s) that should be sent to
+            AWS Lambda. First path in the list should be the path to the handler file which contaitns the main
```
jlewitt1, line 687:
```suggestion
            AWS Lambda. First path in the list should be the path to the handler file which contains the main
```

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,889 @@
...
+            (handler) function. If ``fn`` is provided, this argument is ignored.
+        handler_function_name: str: The name of the function in the handler file that will be executed by the Lambda.
+        runtime: str: The coding language of the fuction. Should be one of the following:
```
jlewitt1, line 690:
```suggestion
        runtime: str: The coding language of the function. Should be one of the following:
```

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,889 @@
...
+            python3.7, python3.8, python3.9, python3.10, python 3.11. (Default: ``python3.9``)
```
jlewitt1, line 691:
```suggestion
            python3.7, python3.8, python3.9, python3.10, python3.11. (Default: ``python3.9``)
```

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,889 @@
...
+        args_names: (Optional[list[str]]): List of the function's accepted parameters, which will be passed to the
+            Lambda Function. If ``fn`` is provided, this argument is ignored.
+            If your function doesn't accept arguments, please provide an empty list.
+        name (Optional[str]): Name of the Lambda Function to create or retrieve.
+            This can be either from a local config or from the RNS.
+        env (Optional[Dict or List[str] or Env]): Specifies the requirments that will be installed, and the enviourment
+            vars that should be attached to the Lambda. Excepts three possible types:
```
jlewitt1, line 698:
```suggestion
            vars that should be attached to the Lambda. Accepts three possible types:
```

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,889 @@
...
+            3. An insrantce of Runhouse Env class.
+            By default, ``runhouse`` package will be installed, and env_vars will include ``{HOME: /tmp/home}``
+        timeout: Optional[int]: The maximum amount of time (in secods) during which the Lambda will run in AWS
+            without timing-out. (Default: ``900``, Min: ``3``, Max: ``900``)
+        memory_size: Optional[int], The amount of memeory, im MB, that will be aloocatied to the lambda.
+             (Default: ``1024``, Min: ``128``, Max: ``10240``)
+        tmp_size: Optional[int], This size of the /tmp folder in the aws lambda file system.
+             (Default: ``3072``, Min: ``512``, Max: ``10240``).
+        region: The AWS region where the lambda will be created. In order to make sure that any packages will be
+            installed proparly, please provide the closest region to your workplace. (Default: ``us-east-1``)
```
jlewitt1, line 712:
```suggestion
            installed properly, please provide the closest region to your workplace. (Default: ``us-east-1``)
```

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,889 @@
...
+        retention_time: Optional[int] The time (in days) that sepcific lambda exection logs will be saved in AWS
```
jlewitt1, line 713:
```suggestion
        retention_time: Optional[int] The time (in days) the lambda execution logs will be saved in AWS
```

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,889 @@
...
+        if not config:
+            raise ValueError(f"Could not find a Lambda called {name}.")
+        return cls.from_config(config)
+
+    # --------------------------------------
+    # Private helping methods
+    # --------------------------------------
+
+    @classmethod
+    def _reqs_to_list(cls, env):
```
jlewitt1, line 166: Maybe add a note that this should no longer be needed once we send the `Env` object to the cluster?

BelSasha: removed this function, since we are installing requirements.txt during runtime.

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,887 @@
...
+def _paths_to_code_from_fn_pointers(fn_pointers):
+    """creates path to code from fn_pinters"""
+    root_dir = rns_client.locate_working_dir()
+    file_path = fn_pointers[1].replace(".", "/") + ".py"
+    paths_to_code = [os.path.join(root_dir, file_path)]
+    return paths_to_code
+
+
+def lambda_function(
+    fn: Optional[str] = None,
```
jlewitt1, line 658: Should be a callable

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,887 @@
...
+        ) as lambda_config:
+            logger.info(
+                f'{lambda_config["Configuration"]["FunctionName"]} was created successfully.'
+            )
+            Path(zip_file_name).absolute().unlink()
+            return lambda_config
+
+    def to(
+        self,
+        # Variables below are deprecated
```
jlewitt1, line 429: `cloud` and `env` are not deprecated

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,887 @@
...
+    """creates path to code from fn_pinters"""
+    root_dir = rns_client.locate_working_dir()
+    file_path = fn_pointers[1].replace(".", "/") + ".py"
+    paths_to_code = [os.path.join(root_dir, file_path)]
+    return paths_to_code
+
+
+def lambda_function(
+    fn: Optional[str] = None,
+    paths_to_code: Optional[str] = None,
```
jlewitt1, line 659: should be a list?

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,887 @@
...
+
+    # --------------------------------------
+    # Private helping methods
+    # --------------------------------------
+
+    def _lambda_exist(self, name):
+        """Checks if a Lambda with the name given during init is already exists in AWS"""
+        func_names = [
+            func["FunctionName"]
+            for func in self.lambda_client.list_functions()["Functions"]
```
jlewitt1, line 169: Instead of iterating over all the existing functions could we query this one specifically? E.g.
https://docs.aws.amazon.com/lambda/latest/dg/example_lambda_GetFunction_section.html

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,887 @@
...
+
+logger = logging.getLogger(__name__)
+
+CRED_PATH = f"{Path.home()}/.aws/credentials"
+SUPPORTED_RUNTIMES = [
+    "python3.7",
+    "python3.8",
+    "python3.9",
+    "python3.10",
+    "python 3.11",
```
jlewitt1, line 30: this is space here intentional?

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,887 @@
...
+        self.args_names = args_names
+        self.runtime = runtime
+        self.timeout = timeout
+        self.memory_size = memory_size
+        self.tmp_size = tmp_size
+        self.aws_lambda_config = (
+            None  # Lambda config and role arn from aws will be saved here
+        )
+        self.retention_time = retention_time
+        self.function_arn = function_arn
```
jlewitt1, line 105: In what scenario would `function_arn` be passed in as an argument?

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,887 @@
...
+                self.lambda_client.delete_function(FunctionName=lambda_name)
+
+            return True
+        except botocore.exceptions.ClientError as aws_exception:
+            raise RuntimeError(
+                f"Could nor delete an AWS resource, got {aws_exception.response['Error']['Message']}"
+            )
+
+
+def _paths_to_code_from_fn_pointers(fn_pointers):
```
jlewitt1, line 649: small nit - can this either be moved to a utils file or the lambdas class? feels a little out of place here with the factory and parent class

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,887 @@
...
+                )
+            time_passed += 1
+            time.sleep(1)
+            config = self._deploy_lambda_to_aws_helper(role_res, zipped_code, env_vars)
+        yield config
+
+    def _create_role_in_aws(self):
+        """create a new role for the lambda function. If already exists - returns it."""
+
+        iam_client = boto3.client("iam")
```
jlewitt1, line 354: initialize this elsewhere so we can re-use it?

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,887 @@
...
+                    + " Please check the resource in AWS console, delete relevant resource(s) "
+                    + "if necessary, and re-run your Runhouse code."
+                )
+            time.sleep(1)
+            lambdas_log_groups = self._get_log_group_names(log_group_prefix)
+            time_passed += 1
+        yield True
+
+    def _invoke(self, *args, **kwargs) -> Any:
+        log_group_prefix = "/aws/lambda/"
```
jlewitt1, line 515: Should this be configurable?

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,887 @@
...
+    ):
+        """
+        Set up a function on AWS as a Lambda function.
+
+        See the args of the factory method :func:`lambda_function` for more information.
+
+        Example:
+            >>> my_lambda = rh.lambda_function(path_to_codes=["full/path/to/model_a_handler.py"],
+            >>> handler_function_name='main_func',
+            >>> runtime='python_3_9',
```
jlewitt1, line 444:
```suggestion
            >>> runtime='python3.9',
```

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,887 @@
...
+            raise RuntimeError("Please provide a path to the lambda handler file.")
+        if isinstance(env, str) and "requirements.txt" in env:
+            paths_to_code.append(Path(env).absolute())
+        if handler_function_name is None or len(handler_function_name) == 0:
+            raise RuntimeError(
+                "Please provide the name of the function that should be executed by the lambda."
+            )
+        if args_names is None:
+            # Parsing the handler file and extracting the arguments names of the handler function.
+            file_path = paths_to_code[0]
```
jlewitt1, line 797: If we can would be cleaner to use the inspect library to handle this extraction for us - something along the lines of: 

```
            spec = importlib.util.spec_from_file_location(module_name, file_path)
            module = importlib.util.module_from_spec(spec)
            spec.loader.exec_module(module)

            # Extract the function and its arguments
            func = getattr(module, func_name, None)
            args_names = inspect.getfullargspec(func).args
```

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,887 @@
...
+            >>> name="my_summer").to()
+            >>> output = my_lambda.map([1, 2], [1, 4], [2, 3])
+            >>> # output = ["4", "9"] (It returns str type because of AWS API)
+
+        """
+
+        return [self._invoke(*args, **kwargs) for args in zip(*args)]
+
+    #
+    def starmap(self, args_lists, **kwargs):
```
jlewitt1, line 577: is `map` and `starmap` working again? otherwise should exclude those in the docs

runhouse/resources/functions/aws_lambda.py
BelSasha: It is saved in DEN, and used for sharing a lambda: if user A in region X shares a lambda with user B in region Y, using the saved arn, user B will be able to load and invoke the function.

runhouse/resources/functions/aws_lambda.py
BelSasha: They always worked, Lambda implementation overrides function implementation.

docs/api/python.rst
```diff
@@ -17,7 +17,7 @@ data abstractions (blobs, folders, tables).
 
 Compute Abstractions
 ------------------------------------
-The Function, Cluster, Env, Package, Run, and Module APIs allow a seamless flow of code and execution across local and remote compute.
+The Function, Lambda Function, Cluster, Env, Package, Run, and Module APIs allow a seamless flow of code and execution across local and remote compute.
```
jlewitt1, line 20: In this section would be worth highlighting the two core usage paths - passing in a function callable, or the typical lambda flow where you provide paths to code and a handler function name

docs/api/python/lambda_function.rst
```diff
@@ -0,0 +1,25 @@
+Lambda Function
+====================================
+
+.. note::
+
+    Lambda Function support is an alpha and under active development.
+    Please report any bugs or let us know of any feature requests.
+
+An AWS Lambda Function is a serverless provided by AWS. It allows to run applications and backend services without
```
jlewitt1, line 9: In this section would be worth highlighting the two core usage paths - passing in a function callable, or the typical lambda flow where you provide paths to code and a handler function name

jlewitt1, line 9:
```suggestion
A `Lambda Function <https://aws.amazon.com/lambda/>`_ is a serverless compute service by AWS. It allows you to run applications and backend services without
```

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,901 @@
...
+        reqs: Optional[List[str]] = None,
+        setup_cmds: Optional[List[str]] = [],
+        force_install: Optional[bool] = False,
+    ):
+        """
+        Set up a function on AWS as a Lambda function.
+
+        See the args of the factory method :func:`lambda_function` for more information.
+
+        Example:
```
jlewitt1, line 451: Isn't the more relevant (and suggested) way of using this method something like:

`rh.function(fn=my_fn).to("aws_lambda")`

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,901 @@
...
+    def map(self, *args, **kwargs):
+        """Map a function over a list of arguments.
+
+        Example:
+            >>> # The my_lambda_handler.py file
+            >>> def my_summer(arg1, arg2, arg3):
+            >>>     return arg1 + arg2 + arg3
+            >>>
+            >>> # your 'main' python file, where you are using runhouse
+            >>> summer_lambda = rh.lambda_function(
```
jlewitt1, line 575: similar to my other comment, but the empty `.to()` call is a little confusing. Aren't we calling it automatically anyway (unless its a dryrun)? Instead maybe we can highlight the usage path of calling `.to("aws_lambda")`?

runhouse/resources/functions/aws_lambda.py
```diff
@@ -0,0 +1,901 @@
...
+            >>> # invokes the Lambda function twice, once with args (1, 2, 3) and once with args (3, 4, 5)
+            >>> output = summer_lambda.starmap(arg_list) # output = ["6", "12"]
+        """
+
+        return [self._invoke(*args, **kwargs) for args in args_lists]
+
+    # --------------------------------------
+    # Lambda Function delete methods
+    # --------------------------------------
+    def delete(self):
```
jlewitt1, line 603: For clusters we have `teardown` and `teardown_and_delete`, the first to delete the infra and the second to infra +  delete the configs from RNS. wdyt about using a similar concept here?

docs/api/python/lambda_function.rst
```diff
@@ -0,0 +1,25 @@
+Lambda Function
```
jlewitt1, line 1: Is it worth including a setup section (e.g. what roles / permissions to configure with AWS), similar to SageMaker (https://www.run.house/docs/api/python/cluster#sagemaker-hardware-setup) or On-demand (https://www.run.house/docs/api/python/cluster#ondemandcluster-hardware-setup)?

PR #142, opened 2023-11-13
Add an end-to-end test for a local docker cluster with telemetry enabled

DenisYay:


DenisYay:
The test can be run using:

pytest -v -s tests/test_http_server.py --force-rebuild

DenisYay:
`test_telemetry.py` tests pass

tests/test_http_server.py
```diff
@@ -3,12 +3,12 @@
 import pytest
 
 
-# @pytest.mark.dockertest
-# @pytest.mark.httpservertest
-# def test_password_cluster_is_up(local_docker_cluster_passwd):
-#     cluster = local_docker_cluster_passwd
-#     cluster.check_server()
-#     assert cluster.is_up()  # Should be true for a Cluster object
+@pytest.mark.dockertest
```
jlewitt1, line 6: thoughts on having a single test that can be paramaterized across clusters? (presumably we'd want to check if the server is up not just on `local_docker_cluster_passwd` and `local_docker_cluster_public_key`?

runhouse/resources/hardware/cluster.py
```diff
@@ -591,6 +594,10 @@ def _use_custom_cert(self):
     def _use_custom_key(self):
         return Path(self.cert_config.key_path).exists()
 
+    @property
+    def _use_local_telemetry(self) -> bool:
```
jlewitt1, line 598: can remove this property getter since its just a boolean?

jlewitt1 (approved):

tests/test_http_server.py
DenisYay: Absolutely, sounds good! This could probably go in as a separate small PR. I can quickly iterate on this tomorrow.

runhouse/resources/hardware/cluster.py
DenisYay: Done.

tests/test_resources/test_clusters/conftest.py
```diff
@@ -358,21 +415,21 @@ def local_docker_cluster_public_key(request):
...
         host="localhost",
         server_host="0.0.0.0",
         ssh_port=local_ssh_port,
         server_connection_type="ssh",
-        den_auth=den_auth,
         ssh_creds={
             "ssh_user": SSH_USER,
             "ssh_private_key": keypath,
         },
+        use_local_telemetry=True,
```
dongreenberg, line 427: It feels very heavy to build and run a whole new container in parallel for a setting change. The same can certainly be said about the different server connection types. In a subsequent PR, could you look into just two fixtures for the containers (one for pwd and one for keypair), and then reuse either the images or containers for various behavior specific features (e.g. enabling telemetry)? Obviously the cleanup would need to be decent. It may also need to be more than just those two, but pretty sure that'd be sufficient for now.

dongreenberg: (Also, in all honesty, I don't see why teletry can't simply be enabled for the existing keypair cluster fixture, unless we think it would interfere with its behavior in some other way).

tests/test_resources/test_clusters/conftest.py
```diff
@@ -343,9 +342,67 @@ def local_docker_cluster_public_key(request):
...
     local_ssh_port = BASE_LOCAL_SSH_PORT + 2
 
+    client, rh_parent_path = build_and_run_image(
+        image_name=image_name,
+        container_name=container_name,
+        dir_name=dir_name,
+        detached=detached,
+        keypath=keypath,
+        force_rebuild=request.config.getoption("--force-rebuild"),
+        port_fwds=[f"{local_ssh_port}:22", "32300:32300"],
```
dongreenberg, line 352: When I ran this I got "RuntimeError: Failed to run docker image keypair-telemetry: docker: Error response from daemon: driver failed programming external connectivity on endpoint rh-slim-keypair-telemetry (f8b9d98d5c96cd3e812949aa8cc850817ec607ebf49b4240a67e1f8263a9c010): Bind for 0.0.0.0:32300 failed: port is already allocated." Another container was up so it may have been a port conflict. You may need to choose a different port on the local side for the mapping. Once I took down the other container the test passed!

dongreenberg (approved):
LGTM! Pretty sick to see the spans showing up!

tests/test_resources/test_clusters/conftest.py
DenisYay: Great points!

In general, there are two schools of thought here:
1. Smart, implicit testing code - e.g. shared fixtures, polymorphic classes inherited in test setups, high degree of reuse of variables and constants
2. Dumb yet explicit testing code - copy-pasta of fixtures and test methods, many testing methods are very similar to each other with maybe 1-2 lines of diff, more code overall

The Google school of thought went from extreme 1 to extreme 2 over the course of the last 8 years or so.

Some of the reasoning behind it were:
* New tests can be added quickly by following existing examples. There's no additional domain knowledge that is required for it so new-hires and folks who are not familiar with the existing intricacies can do it in no time.
* If something breaks, it usually impacts only an isolated number of tests that should be relatively easy to fix.
* Maintenance costs of the existing testing code are usually lower - refactoring efforts usually impact only an isolated number of methods and fixtures and the chance of breaking the rest of the tests depending on the changes is lower
* Improved readability - less in-method/in-class branching code

Some of the downsides of this approach are:
* More code
* Resource management - if run on remote GPUs / other expensive hardware, it is better to share and start less containers, for instance. Same logic applied to CPU clusters is debatable.

Personally, I feel like in our case, a moderate 2 approach would be preferred: dumb testing code optimizing costs. So if the code leads to more GPU clusters' resources that are used, which results in more money we pay out of the pocket - no bueno.

Overall, we absolutely don't need to follow Google's practices on this, but I wanted to give this as a context and hear your thoughts on this. Thanks!

DenisYay: > (Also, in all honesty, I don't see why teletry can't simply be enabled for the existing keypair cluster fixture, unless we think it would interfere with its behavior in some other way).

Sure, we can have only one fixture with the telemetry enabled. One difference would be the memory usage footprint, but if we keep the cluster usage moderate and don't reuse the same running instance in hundreds of tests, I think it should be fine.

tests/test_resources/test_clusters/conftest.py
DenisYay: Good catch! I had the same error early on in testing, then started running only one test until it passed. 
I think Rohin's PR should fix this, happy to take another look if it isn't.

tests/test_resources/test_clusters/conftest.py
dongreenberg: Right, I see the tradeoff, but in this case I think the primary driver is simply the memory consumption of having 4 container running simultaneously on a laptop, while only 1 is actually doing anything at a time. I think we can surgical about optimizing consumption where needed without overengineering the test suite. I agree that a test suite that needs tests is an antipattern.

tests/test_resources/test_clusters/conftest.py
dongreenberg: @rohinb2 FYI

tests/test_http_server.py
DenisYay: Absolutely, thanks for catching this.

tests/test_resources/test_clusters/conftest.py
DenisYay: Yep, agree.

In addition to optimizing memory consumption, hopefully the GitHub Actions approach (or the general CI/CD ramp up approach) should alleviate some of these concern.

PR #143, opened 2023-11-14
[WIP] send rh function instance to AWS Lambda

BelSasha:
Added the ability to send a Runhouse function instance to AWS Lambda. 
Now want to merge it to my "main" lambdas_poc branch in order to work on a single unified branch.

### DIFF
--- a/runhouse/resources/function.py
+++ b/runhouse/resources/function.py
@@ -147,13 +147,6 @@ class Function(Module):
 
         return new_function
 
-    @property
-    def __name__(self):
-        if self.name:
-            return self.name
-        else:
-            return self.fn_pointers[2]
-
     # ----------------- Function call methods -----------------
 
     def __call__(self, *args, **kwargs) -> Any:
--- a/runhouse/resources/function.py
+++ b/runhouse/resources/function.py
@@ -29,6 +29,7 @@ class Function(Module):
         env: Optional[Env] = None,
         dryrun: bool = False,
         access: Optional[str] = None,
+        func_args: Optional[str] = None,
         **kwargs,  # We have this here to ignore extra arguments when calling from from_config
     ):
         """
@@ -40,6 +41,7 @@ class Function(Module):
         """
         self.fn_pointers = fn_pointers
         self.access = access or self.DEFAULT_ACCESS
+        self.func_args = func_args
         super().__init__(name=name, dryrun=dryrun, system=system, env=env, **kwargs)
 
     # ----------------- Constructor helper methods -----------------
@@ -87,10 +89,16 @@ class Function(Module):
             >>> rh.function(fn=local_fn).to(gpu_cluster)
             >>> rh.function(fn=local_fn).to(system=gpu_cluster, env=my_conda_env)
         """
+        from runhouse.resources.serverless import aws_lambda_function
+
         if setup_cmds:
             warnings.warn(
                 "``setup_cmds`` argument has been deprecated. "
                 "Please pass in setup commands to the ``Env`` class corresponding to the function instead."
+            )
+        if system == "AWS_LAMBDA":
+            return aws_lambda_function(
+                fn_pointers=self.fn_pointers, args_names=self.func_args
             )
 
         # to retain backwards compatibility
@@ -442,6 +450,7 @@ def function(
         env = _get_env_from(env) or Env(working_dir="./", name=Env.DEFAULT_NAME)
 
     fn_pointers = None
+    func_args = None
     if callable(fn):
         fn_pointers = Function._extract_pointers(fn, reqs=env.reqs)
         if fn_pointers[1] == "notebook":
@@ -451,6 +460,7 @@ def function(
                 serialize_notebook_fn=serialize_notebook_fn,
                 name=fn_pointers[2] or name,
             )
+        func_args = [param.name for param in inspect.signature(fn).parameters.values()]
     elif isinstance(fn, str):
         # Url must match a regex of the form
         # 'https://github.com/username/repo_name/blob/branch_name/path/to/file.py:func_name'
@@ -489,6 +499,7 @@ def function(
         access=Function.DEFAULT_ACCESS,
         name=name,
         dryrun=dryrun,
+        func_args=func_args,
     ).to(system=system, env=env)
 
     if load_secrets and not dryrun:
--- a/runhouse/resources/function.py
+++ b/runhouse/resources/function.py
@@ -29,7 +29,6 @@ class Function(Module):
         env: Optional[Env] = None,
         dryrun: bool = False,
         access: Optional[str] = None,
-        func_args: Optional[str] = None,
         **kwargs,  # We have this here to ignore extra arguments when calling from from_config
     ):
         """
@@ -41,7 +40,6 @@ class Function(Module):
         """
         self.fn_pointers = fn_pointers
         self.access = access or self.DEFAULT_ACCESS
-        self.func_args = func_args
         super().__init__(name=name, dryrun=dryrun, system=system, env=env, **kwargs)
 
     # ----------------- Constructor helper methods -----------------
--- a/runhouse/resources/function.py
+++ b/runhouse/resources/function.py
@@ -89,7 +89,6 @@ class Function(Module):
             >>> rh.function(fn=local_fn).to(gpu_cluster)
             >>> rh.function(fn=local_fn).to(system=gpu_cluster, env=my_conda_env)
         """
-        from runhouse.resources.serverless import aws_lambda_function
 
         if setup_cmds:
             warnings.warn(
@@ -97,9 +96,10 @@ class Function(Module):
                 "Please pass in setup commands to the ``Env`` class corresponding to the function instead."
             )
         if system == "AWS_LAMBDA":
-            return aws_lambda_function(
-                fn_pointers=self.fn_pointers, args_names=self.func_args
-            )
+            from runhouse.resources.serverless import aws_lambda_function
+
+            args = self.fn_pointers[-1]
+            return aws_lambda_function(fn_pointers=self.fn_pointers, args_names=args)
 
         # to retain backwards compatibility
         if reqs or setup_cmds:
@@ -450,7 +450,6 @@ def function(
         env = _get_env_from(env) or Env(working_dir="./", name=Env.DEFAULT_NAME)
 
     fn_pointers = None
-    func_args = None
     if callable(fn):
         fn_pointers = Function._extract_pointers(fn, reqs=env.reqs)
         if fn_pointers[1] == "notebook":
@@ -460,7 +459,6 @@ def function(
                 serialize_notebook_fn=serialize_notebook_fn,
                 name=fn_pointers[2] or name,
             )
-        func_args = [param.name for param in inspect.signature(fn).parameters.values()]
     elif isinstance(fn, str):
         # Url must match a regex of the form
         # 'https://github.com/username/repo_name/blob/branch_name/path/to/file.py:func_name'
@@ -499,7 +497,6 @@ def function(
         access=Function.DEFAULT_ACCESS,
         name=name,
         dryrun=dryrun,
-        func_args=func_args,
     ).to(system=system, env=env)
 
     if load_secrets and not dryrun:
--- a/runhouse/resources/function.py
+++ b/runhouse/resources/function.py
@@ -98,8 +98,9 @@ class Function(Module):
         if system == "AWS_LAMBDA":
             from runhouse.resources.serverless import aws_lambda_function
 
-            args = self.fn_pointers[-1]
-            return aws_lambda_function(fn_pointers=self.fn_pointers, args_names=args)
+            return aws_lambda_function(
+                fn=self._get_obj_from_pointers(*self.fn_pointers), env=env
+            )
 
         # to retain backwards compatibility
         if reqs or setup_cmds:
@@ -147,6 +148,13 @@ class Function(Module):
         logging.info("Function setup complete.")
 
         return new_function
+
+    @property
+    def __name__(self):
+        if self.name:
+            return self.name
+        else:
+            return self.fn_pointers[2]
 
     # ----------------- Function call methods -----------------
 
--- a/runhouse/resources/function.py
+++ b/runhouse/resources/function.py
@@ -93,12 +93,6 @@ class Function(Module):
                 "``setup_cmds`` argument has been deprecated. "
                 "Please pass in setup commands to the ``Env`` class corresponding to the function instead."
             )
-        if system == "AWS_LAMBDA":
-            from runhouse.resources.serverless import aws_lambda_function
-
-            return aws_lambda_function(
-                fn=self._get_obj_from_pointers(*self.fn_pointers), env=env
-            )
 
         # to retain backwards compatibility
         if reqs or setup_cmds:
@@ -111,6 +105,13 @@ class Function(Module):
         else:
             env = env or self.env or Env(name=Env.DEFAULT_NAME)
             env = _get_env_from(env)
+
+        if system == "AWS_LAMBDA":
+            from runhouse.resources.serverless import aws_lambda_function
+
+            return aws_lambda_function(
+                fn=self._get_obj_from_pointers(*self.fn_pointers), env=env
+            )
 
         if (
             self.dryrun
--- a/runhouse/resources/serverless/aws_lambda_function.py
+++ b/runhouse/resources/serverless/aws_lambda_function.py
@@ -810,7 +810,6 @@ def aws_lambda_function(
 
     if fn is None:
         # ------- arguments validation -------
-        # TODO: use error msg inside the exception
         if paths_to_code is None or len(paths_to_code) == 0:
             raise RuntimeError("Please provide a path to the lambda handler file.")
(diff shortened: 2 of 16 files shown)


runhouse/resources/function.py
```diff
@@ -40,6 +41,7 @@ def __init__(
         """
         self.fn_pointers = fn_pointers
         self.access = access or self.DEFAULT_ACCESS
+        self.func_args = func_args
```
jlewitt1, line 44: can't we extract these ourselves without introducing a new argument? We do this elsewhere using: `func_args = [param.name for param in inspect.signature(fn).parameters.values()]`

runhouse/resources/function.py
```diff
@@ -87,11 +89,17 @@ def to(
             >>> rh.function(fn=local_fn).to(gpu_cluster)
             >>> rh.function(fn=local_fn).to(system=gpu_cluster, env=my_conda_env)
         """
+        from runhouse.resources.serverless import aws_lambda_function
```
jlewitt1, line 92: move inside the if statement where it would be called?

runhouse/resources/serverless/aws_lambda_function.py
```diff
@@ -709,9 +709,14 @@ def config_for_rns(self):
...
         return config
 
+    @property
+    def handler_path(self):
+        return self.local_path_to_code[0]
+
 
 def aws_lambda_function(
     fn: Callable = None,
+    fn_pointers: Optional[Tuple] = None,
```
jlewitt1, line 719: We should be able to extract these ourselves instead of having the user provide them (assuming we are given the `fn`)

runhouse/resources/function.py
BelSasha: makes sense, moved.

runhouse/resources/function.py
BelSasha: Made a change so we will not have to introduce a new argument.

runhouse/resources/function.py
```diff
@@ -140,6 +149,13 @@ def to(
 
         return new_function
 
+    @property
+    def __name__(self):
```
jlewitt1, line 153: I dont think we should overwrite the behavior of name (this is initialized in parent classes), and we also do want to allow anonymous functions

BelSasha: Yeah agreed, deleted it, it was part of the debugging I was doing.

runhouse/resources/function.py
```diff
@@ -87,11 +87,18 @@ def to(
             >>> rh.function(fn=local_fn).to(gpu_cluster)
             >>> rh.function(fn=local_fn).to(system=gpu_cluster, env=my_conda_env)
         """
+
         if setup_cmds:
             warnings.warn(
                 "``setup_cmds`` argument has been deprecated. "
                 "Please pass in setup commands to the ``Env`` class corresponding to the function instead."
             )
+        if system == "AWS_LAMBDA":
```
jlewitt1, line 96: this logic should prob be moved below where we build the Env object, since that will also be passed in to the constructor

BelSasha: makes sense, moving it after creating the env.

tests/test_resources/test_modules/test_functions/test_aws_lambda/test_lambdas_with_fn.py
```diff
@@ -1,10 +1,56 @@
...
+            str(Path(aws_func.handler_path).parent / f"rh_handler_{aws_func.name}.py")
+        )
+    except FileNotFoundError:
+        assert True is True
+    try:
+        os.remove(
+            str(Path(aws_func.handler_path).parent / f"{aws_func.name}_code_files.zip")
+        )
+    except FileNotFoundError:
+        assert True is True
```
jlewitt1, line 30: if we expect an exception to be raised, would be better to do something like: 
```
    with pytest.raises(Exception): 
        rh.aws_lambda_function()
```

runhouse/resources/serverless/aws_lambda_function.py
BelSasha: Yeah, makes sense. Removed.

tests/test_resources/test_modules/test_functions/test_aws_lambda/test_lambdas_with_fn.py
BelSasha: Done

runhouse/resources/serverless/aws_lambda_function.py
```diff
@@ -71,7 +72,7 @@ def __init__(
         fn: Optional[Callable] = None,
         fn_pointers: Optional[tuple] = None,
         name: Optional[str] = None,
-        env: Optional[list[str] or str] = None,
+        reqs: Optional[list[str]] = None,
         env_vars: Optional[dict] = None,
```
jlewitt1, line 76: can we extract `env_vars` from the env object instead of breaking it out as a separate input to the factory

runhouse/resources/serverless/aws_lambda_function.py
```diff
@@ -784,47 +772,76 @@ def aws_lambda_function(
...
 
+    if fn is None:
+        # ------- arguments validation -------
+        if paths_to_code is None or len(paths_to_code) == 0:
+            raise RuntimeError("Please provide a path to the lambda handler file.")
+        if handler_function_name is None or len(handler_function_name) == 0:
+            raise RuntimeError(
+                "Please provide the name of the function that should be executed by the lambda."
+            )
+        if runtime is not None and runtime not in SUPPORTED_RUNTIMES:
```
jlewitt1, line 819: this check should happen regardless of whether `fn` is provided or not

runhouse/resources/serverless/aws_lambda_function.py
```diff
@@ -784,47 +772,76 @@ def aws_lambda_function(
...
+                "Please provide a supported lambda runtime, should be one of the"
+                + f" following: {SUPPORTED_RUNTIMES}"
+            )
+        if args_names is None:
+            raise RuntimeError(
+                "Please provide the names of the arguments provided to handler function, "
+                + "in the order they are passed to the lambda function."
+            )
+
+        if timeout > 900:
```
jlewitt1, line 830: these checks can also be broken out of the `fn` check

runhouse/resources/serverless/aws_lambda_function.py
```diff
@@ -71,7 +72,7 @@ def __init__(
         fn: Optional[Callable] = None,
```
jlewitt1, line 72: the class should only accept `fn_pointers`, can remove `fn`. By saving `fn_pointers` in Den that's enough to reload / share the function

runhouse/resources/serverless/aws_lambda_function.py
```diff
@@ -87,7 +88,7 @@ def __init__(
         """
         if isinstance(fn, Callable):
```
jlewitt1, line 89: can remove all these checks here since all this logic is in the factory

runhouse/resources/serverless/aws_lambda_function.py
```diff
@@ -784,47 +772,76 @@ def aws_lambda_function(
...
+        elif isinstance(env, dict):
+            reqs = env["reqs"]
+            env_vars = env["env_vars"]
+        else:
+            raise Exception(
+                "Env's type could be runhouse Env or dictionary. Please provide env in the correct type"
+                + " and rerun."
+            )
+        if isinstance(reqs, str):
+            reqs = AWSLambdaFunction._reqs_to_list(reqs)
```
jlewitt1, line 801: this is meant to happen when reqs == "requirements.txt"?

runhouse/resources/serverless/aws_lambda_function.py
```diff
@@ -784,47 +772,76 @@ def aws_lambda_function(
...
+    if fn is None:
+        # ------- arguments validation -------
+        if paths_to_code is None or len(paths_to_code) == 0:
+            raise RuntimeError("Please provide a path to the lambda handler file.")
+        if handler_function_name is None or len(handler_function_name) == 0:
+            raise RuntimeError(
+                "Please provide the name of the function that should be executed by the lambda."
+            )
+        if runtime is not None and runtime not in SUPPORTED_RUNTIMES:
+            raise RuntimeError(
```
jlewitt1, line 820: we should also set a default runtime here if one is not provided

runhouse/resources/serverless/aws_lambda_function.py
```diff
@@ -784,47 +772,76 @@ def aws_lambda_function(
...
+                "Please provide a supported lambda runtime, should be one of the"
+                + f" following: {SUPPORTED_RUNTIMES}"
+            )
+        if args_names is None:
```
jlewitt1, line 824: can't we also extract these if the `fn` is passed in? seems like it will always be passed into the class with `None`

runhouse/resources/serverless/aws_lambda_function.py
BelSasha: yep, fixed indentation.

runhouse/resources/serverless/aws_lambda_function.py
BelSasha: done, also for memory_size validation.

runhouse/resources/serverless/aws_lambda_function.py
BelSasha: makes sense, moved the fn parsing logic from constructor to factory method.

runhouse/resources/serverless/aws_lambda_function.py
BelSasha: did some refactor, since not all logic was implemented in the factory method.

runhouse/resources/serverless/aws_lambda_function.py
BelSasha: yes

runhouse/resources/serverless/aws_lambda_function.py
BelSasha: already done in constructor. Will raise a warning instead of an exception, since we are setting a default runtime.

runhouse/resources/serverless/aws_lambda_function.py
BelSasha: If the fn is provided, we are indeed extracting the arguments names. 
But if fn is not provided, we expect the user to provide the args names. If the arguments will not be passed, I've added a piece of code which parses the hander and extracts the arguments names.

runhouse/resources/serverless/aws_lambda_function.py
```diff
@@ -58,24 +55,33 @@ class AWSLambdaFunction(Function):
...
-    DEFAULT_REGION = "us-east-1"
-    DEFAULT_PY_VERSION = "python3.9"
     EMPTY_ZIP = -1
+    MAX_WAIT_TIME = 60  # seconds, max time that can pass before we raise an exception that AWS update takes too long.
+    DEFAULT_REGION = "us-east-1"
+    DEFAULT_RETENTION = 30  # one month, for lambdas log groups.
+    DEFAULT_TIMEOUT = 30  # seconds
+    DEFAULT_MEMORY_SIZE = 128  # MB
+
+    if Path(CRED_PATH).is_file():
```
jlewitt1, line 65: do these still need to be class attributes? we could also initialize them as part of the constructor

runhouse/resources/serverless/aws_lambda_function.py
```diff
@@ -320,7 +323,6 @@ def _create_layer_zip(self):
             shutil.rmtree(dir_name)
             return self.EMPTY_ZIP
```
jlewitt1, line 324: instead of having a separate class attribute to mange this state, just return None?

tests/test_resources/test_modules/test_functions/test_aws_lambda/test_lambdas_with_fn.py
```diff
@@ -1,10 +1,49 @@
...
+
 def test_create_func():
-    import runhouse as rh
+    with pytest.raises(FileNotFoundError):
+        import runhouse as rh
+
+        aws_func = rh.aws_lambda_function(fn=summer, name="summer_lambdas").save()
+        res = aws_func(1, 3)
+        assert int(res) == 4
+        os.remove(
```
jlewitt1, line 22: small nit - use Pathlib instead (path.unlink)

runhouse/resources/serverless/aws_lambda_function.py
```diff
@@ -784,57 +829,107 @@ def aws_lambda_function(
...
+                raise TypeError("AWS Lambda accepts env_vars of dict type only")
+        elif isinstance(env, dict):
+            reqs = env["reqs"]
+            if isinstance(reqs, str):
+                reqs = [reqs]
+            env_vars = env["env_vars"]
+            f_name = name or fn.__name__ or handler_function_name
+            env = Env(reqs=reqs, name=f"{f_name}_env", env_vars=env_vars)
+        else:
+            received_type = str(type(env)).removeprefix("<class '").removesuffix("'>")
```
jlewitt1, line 855: why would this happen?

runhouse/resources/serverless/aws_lambda_function.py
```diff
@@ -85,76 +91,41 @@ def __init__(
...
-            fn_pointers = None
-            if handler_function_name is None or paths_to_code is None:
-                raise ValueError(
-                    "`handler_function_name` and `paths_to_code` must be provided if `fn` is not provided"
-                )
-
-        if name is None:
-            name = handler_function_name.replace(".", "_")
 
         super().__init__(
```
jlewitt1, line 95: Just noting that the `access` attribute will probably need be included here eventually

runhouse/resources/serverless/aws_lambda_function.py
```diff
@@ -195,14 +172,10 @@ def _check_for_child_configs(cls, config):
...
 
     @staticmethod
-    def _paths_to_code_from_fn_pointers(fn_pointers):
-        root_dir = rns_client.locate_working_dir()
-        file_path = fn_pointers[1].replace(".", "/") + ".py"
-        paths_to_code = [os.path.join(root_dir, file_path)]
-        return paths_to_code
-
-    def _reqs_to_list(self, env):
+    def _reqs_to_list(env):
```
jlewitt1, line 175: do we still need this if we are sending the env to the lambda where the Env object can handle the reqs internally?

runhouse/resources/serverless/aws_lambda_function.py
```diff
@@ -85,76 +91,41 @@ def __init__(
...
-
-        if memory_size < 128:
-            # TODO: send a warning to the user
-            memory_size = 128
-        if memory_size > 10240:
-            # TODO: send a warning to the user
-            memory_size = 10240
         self.memory_size = memory_size
-
+        self.env_vars, self.reqs = None, None
```
jlewitt1, line 110: we can extract these directly from the Env object

runhouse/resources/serverless/aws_lambda_function.py
```diff
@@ -603,9 +624,9 @@ def to(
...
         if self._lambda_exist(self.name):
             # updating the configuration with the initial configuration.
-            # TODO{SB}: enable the user to change the config of the Lambda.
+            # TODO [SB]: enable the user to change the config of the Lambda.
             # lambda_config = self._update_lambda_config(env_vars)
-            lambda_config = LAMBDA_CLIENT.get_function(FunctionName=self.name)
+            lambda_config = self.LAMBDA_CLIENT.get_function(FunctionName=self.name)
 
         else:
             # creating a new Lambda function, since it's not existing in the AWS account which is configured locally.
```
jlewitt1, line 632: similar note as above, we don't need to pass `env_vars` around if we can access them through the env object

runhouse/resources/serverless/aws_lambda_function.py
```diff
@@ -320,7 +323,6 @@ def _create_layer_zip(self):
             shutil.rmtree(dir_name)
             return self.EMPTY_ZIP
         for req in reqs:
```
jlewitt1, line 325: this fails if reqs is set to: `./` (as set by the env)

runhouse/resources/serverless/aws_lambda_function.py
```diff
@@ -241,14 +215,43 @@ def _wait_until_update_is_finished(self, name):
...
+                time_passed += 1
+                response = self.LAMBDA_CLIENT.get_function(FunctionName=name)
                 state = response["Configuration"]["State"] == "Active"
                 last_update_status = (
                     response["Configuration"]["LastUpdateStatus"] != "InProgress"
                 )
         return True
 
+    @contextlib.contextmanager
+    def _wait_until_log_group_update_is_finished(self, log_group_name):
```
jlewitt1, line 234: where do we use this?

runhouse/resources/serverless/aws_lambda_function.py
BelSasha: makes sense, changed.

runhouse/resources/serverless/aws_lambda_function.py
BelSasha: A rare edge case. The probability it'll happen is really low, so removing it for now. Will add later on if it be necessary (will be more clear after users will start using AWS Lambdas).

PR #146, opened 2023-11-15
Fixing test account context manager to correctly update account in global cache.

rohinb2:
There was a bug in the `test_account` context manager / fixture that led to the config not being properly updated globally. This led to Resources being created with the wrong account in test usages.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,12 +1,9 @@
 import contextlib
 import enum
-import os
-
-import dotenv
 
 import pytest
 
-import runhouse as rh
+from runhouse.globals import rns_client
 
 
 class TestLevels(str, enum.Enum):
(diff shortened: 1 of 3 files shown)


rohinb2:
* main
  * **PR #146** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/146?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/146?utm_source=stack-comment).

tests/conftest.py
```diff
@@ -82,39 +79,14 @@ def test_account():
...
-        yield {
-            "test_token": test_account_token,
-            "test_username": test_account_username,
-            "test_folder": test_account_folder,
-        }
-
+        account = rns_client.load_account_from_env()
+        if account is None:
+            pytest.skip("`TEST_TOKEN` or `TEST_USERNAME` not set, skipping test.")
+
```
jlewitt1, line 87: to enable using the fixture as a context manager (ex: `with test_account`) we should also yield the account here

rohinb2:
oh yep forgot that after i updated

jlewitt1 (approved):

PR #153, opened 2023-11-19
Server Tests Github Action

jlewitt1:


.github/workflows/server_tests.yaml
```diff
@@ -0,0 +1,61 @@
...
+      - '*'
+  workflow_dispatch:
+
+jobs:
+  server-test:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+#        python-version: [3.7, 3.8, 3.9, 3.10]
```
jlewitt1, line 18: @dongreenberg you think it's overkill to run these tests across multiple python versions?

dongreenberg:
I'm not sure we really need a server tests action specifically. We can just run `pytest tests/ --level unit` with Rohin's recent changes, and the UNIT server tests will be run within the set (as soon as they're marked as such).

.github/workflows/server_tests.yaml
dongreenberg: Will they launch in parallel (i.e. as separate actions)?

.github/workflows/server_tests.yaml
```diff
@@ -0,0 +1,61 @@
...
+        echo "${{ secrets.DOCKER_SSH_PUBLIC_KEY }}" > $HOME/.ssh/runhouse/docker/id_rsa.pub
+        chmod 600 $HOME/.ssh/runhouse/docker/id_rsa.pub
+
+    - name: Save Runhouse config
+      env:
+        TEST_TOKEN: ${{ secrets.TEST_TOKEN }}
+        TEST_USERNAME: ${{ secrets.TEST_USERNAME }}
+      run: |
+        mkdir -p $HOME/.rh
+        python -c "import yaml; import os; config = {'token': os.environ['TEST_TOKEN'], 'username': os.environ['TEST_USERNAME'], 'default_folder': f\"/{os.environ['TEST_USERNAME']}\", 'default_keypair': '$HOME/.ssh/runhouse/docker/id_rsa'}; yaml.dump(config, open('$HOME/.rh/config.yaml', 'w'))"
```
dongreenberg, line 51: I would actually rather we put these at `~/.ssh/sky-key` so they get picked up by default and we don't need to set a default_keypair, and then we can simply run `runhouse login -y $TEST_TOKEN` here instead of manually dropping in the config yaml, which is flakier and less realistic.

dongreenberg: Also, there's a big problem with this. We test sharing by sharing a resource with test account. In this action the sharer and sharee would be the same. We probably need a second CI test account or something that we use to log in here, so it's sharing with Test account just like we are locally.

PR #149, opened 2023-11-16
Fix global ssh tunnel cache to use port as well.

rohinb2:
We use a global dictionary to cache open ssh connections to particular
addresses. However, for local tests where several connections are running
locally on localhost, the cache would incorrectly hit.

PR #159, opened 2023-11-20
Update Dockerfile for Sudo support + Nginx

jlewitt1:
add separate fixture for local tls with nginx testing

Fix global ssh tunnel cache to use port as well. (#149)

We use a global dictionary to cache open ssh connections to particular
addresses. However, for local tests where several connections are running
locally on localhost, the cache would incorrectly hit.

Add an end-to-end test for a local docker cluster with telemetry enabled (#142)

Disable telemetry by default in the local docker public key cluster

Add ServerModule Tests (#151)

update nginx tests for local docker

Merge remote-tracking branch 'origin/main' into telemetry-nginx-docker-updates

update nginx local tests to use separate containers for http / https

set up client port for nginx fixtures

### DIFF
--- a/runhouse/docker/slim/password-file-auth/Dockerfile
+++ b/runhouse/docker/slim/password-file-auth/Dockerfile
@@ -14,7 +14,7 @@ RUN mkdir -p /app/ssh
 
 # Install the required packages
 RUN apt-get update --allow-insecure-repositories && \
-    apt-get install -y --no-install-recommends gcc python3-dev openssh-server rsync supervisor screen curl && \
+    apt-get install -y --no-install-recommends gcc python3-dev openssh-server rsync supervisor screen curl sudo ufw && \
     apt-get clean && \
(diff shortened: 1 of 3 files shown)


jlewitt1:
* main
  * **PR #159** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/159?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/159?utm_source=stack-comment).

DenisYay (approved):
This looks great, thanks!

There's actually one more Dockerfile that we should update:

https://github.com/run-house/runhouse/blob/main/runhouse/docker/cuda/Dockerfile

We should also likely have some tests around it, but these can wait until we externalize an official Runhouse Docker support.

PR #177, opened 2023-11-26
install docker before running unittests in git (CI)

BelSasha:


dongreenberg:
What necessitated this? The UNIT level tests are supposed to not need Docker (the definition of LOCAL is using containers, whereas UNIT shouldn't need them).

PR #178, opened 2023-11-27
Refactor env tests and add env fixtures

carolineechen:
* added fixtures for various types of base and conda envs
* refactored env tests to subclass resource test and reuse tests across fixtures
* usage: `pytest -v --level "minimal" tests/test_resources/test_envs/test_env.py`

next steps for env testing
* conda docker container working to enable local tests
* cut down on testing time

### DIFF
--- a/tests/test_resources/test_envs/test_env.py
+++ b/tests/test_resources/test_envs/test_env.py
@@ -1,5 +1,4 @@
 import os
-import shutil
 import unittest
 
 from pathlib import Path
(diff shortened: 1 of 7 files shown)


tests/test_resources/test_envs/test_env.py
```diff
@@ -343,25 +37,168 @@ def np_summer(a, b):
...
+        env.to(cluster)
+
+        for req in env.reqs:
+            if not req == "./":
+                res = cluster.run([f"pip freeze | grep {req}"], env=env)
+                assert res[0][0] == 0  # installed properly
+
+        _uninstall_env(env, cluster)
+
+    @unittest.skip("Running into s3 folder issue")
```
carolineechen, line 107: test currently failing, from trying to copy rh.folder to s3. sky added a `_assert_kwargs_builtin_type` check recently that's resulting `AssertionError: kwargs should not contain none built-in types: {'region_name': None}`. needs to be addressed for folder tests to pass as well.

tests/test_resources/test_envs/test_env.py
```diff
@@ -343,25 +37,168 @@ def np_summer(a, b):
...
-@unittest.skip("Map does not work properly following Module refactor.")
-@pytest.mark.clustertest
-def test_conda_map_fn(cluster):
-    conda_dict = _get_conda_env(name="test-map-fn")
-    conda_env = rh.env(conda_env=conda_dict, reqs=["pytest", "numpy"])
-    map_fn = rh.function(np_summer, system=cluster, env=conda_env)
-    inputs = list(zip(range(5), range(4, 9)))
-    results = map_fn.starmap(inputs)
-    assert results == [4, 6, 8, 10, 12]
+    MAP_FIXTURES = {"resource": "env"}
```
carolineechen, line 42: @dongreenberg this is still running all the base resource tests for the resource fixtures in addition to the env fixtures. how should that be resolved?

dongreenberg (approved):
This is great. Now that the existing tests have been migrated to the new structure, I think I would prioritize 1) bringing down the test level for most of these to at least LOCAL, even if that means excluding the conda envs from UNIT and LOCAL initially and 2) Adding UNIT level tests which run without a docker cluster. Our current tests are very integrationy (e.g. we test env.to by checking that reqs were installed, rather than using a local HTTP server fixture and checking that the servlet was created properly). I don't feel super strongly that those are more important than getting conda env working, so if you're in the middle of that no need to put it down.

PR #181, opened 2023-11-28
Remove duplication by using fixtures in HTTP server tests.

rohinb2:
There was extensive code duplication between testing with Den Auth and without. These are now merged into one file, using fixtures to pass different clusters in for different tests. 

The tests are now also added to the top level pytest.

This is as @jlewitt1 and I discussed.

cc @dongreenberg 

For reviewing, I recommend just viewing the output file. I didn't change the content of any tests, really, just the structuring.

## Testing

`pytest -v --level local` (includes unit)

`pytest -v tests/test_servers/ --level local  --ignore-filters`

PR #182, opened 2023-11-28
Avoid ever importing fixtures from conftest.

rohinb2:
When setting up the fixture list within a class, we were doing something along the lines of:

```
class TestClass:
    UNIT = {"cluster": [local_cluster]}
```

This requires an import of `local_cluster` from conftest. Turns out, this causes session scoped fixtures to be recomputed (https://github.com/pytest-dev/pytest/issues/3217). It's bad practice to import from conftest in the first place. So, I changed the fixtures that run `getfixturevalue` to just use the string argument, and we can avoid importing, and rely on pytest to set up the fixtures correctly. Now we set up fixtures like so:

```
class TestClass:
    UNIT = {"cluster": ["local_cluster"]}
```

Moreover... if a fixture is defined in a nested `conftest.py`, and imported in the global `conftest.py`, then there can be conflicts. The way tests work is that they'll look heirarchically upward in conftest.py files, and then grab the first fixture they find that matches the name. Even if the global `conftest.py` imports from a nested `conftest.py` file, it is considered a different fixture object (unintuitive, yes). This will sometimes lead to session scoped fixtures being initialized twice in the same session. More info on nested conftest.py files [here](https://github.com/pytest-dev/pytest/issues/1931).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -5,6 +5,54 @@ import pytest
 
 from runhouse.globals import rns_client
 
+"""
+HOW TO USE FIXTURES IN RUNHOUSE TESTS
+
+You can parameterize a fixture to run on many variations for any test.
+
+Make sure your parameterized fixture is defined in this file (can be imported from some other file)
+like so:
+
+@pytest.fixture(scope="function")
+def cluster(request):
+    return request.getfixturevalue(request.param)
+
+You can define default_fixtures for any parameterized fixture below.
+
+You can use MAP_FIXTURES to map a parameterized fixture to a different name. This
+is useful if you have a subclass that you want to test the parent's parameterized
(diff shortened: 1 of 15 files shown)


rohinb2:
* main
  * **PR #181** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/181?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #182** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/182?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/182?utm_source=stack-comment).

PR #184, opened 2023-11-30
Function tests passing with THOROUGH

dongreenberg:
- Fix renaming bug where rename rpc doesn't lookup env in which to rename remote object properly
- Fix renaming bug where we check if passed name (which can be a name or rns_address) matches resource.name (not address), which would be False even if full rns_addresses match
- Fix multiple bugs with cluster tunnels (not looking up open tunnels properly in pool, not setting client_port to found connected_port after searching)
- Fix cluster factory bug (not instantiating from_name because server_port or connection type was defaulted to not None value)
- Change usage of byo_cpu to static_cpu_cluster
- Test calling function directly over HTTP properly
- Test multiprocessing properly to stop hanging
- Change a bunch of unittest.skips into pytest.mark.skips to avoid creating fixture for no reason
- Remove torch usage in tests, way too slow and OOMs in containers

Function and Module tests all pass with THOROUGH


dongreenberg:
* main
  * **PR #184** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/184?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/184?utm_source=stack-comment).

PR #187, opened 2023-11-30
Follow up updates to README

DenisYay:


README.md
```diff
@@ -42,9 +42,9 @@ power their ML infra, and Runhouse makes its best-in-class features accessible t
 ## 🤨 Who is this for?
 
 * 👩‍🔧 **Engineers, Researchers and Data Scientists** who don't want to spend 3-6 months translating and packaging
-their work to share it, and want to be able to iterate and improve production services, pipelines, and data artifacts
+their work to share it, and want to be able to iterate and improve their experiments, production services, pipelines, and data artifacts
```
dongreenberg, line 45: Good addition, but I would move experiments after pipelines, and wouldn't use "their" (I want to evoke that you can pick up any service, pipeline, etc. and iterate on it, not only your own).

DenisYay: Done

README.md
```diff
@@ -42,9 +42,9 @@ power their ML infra, and Runhouse makes its best-in-class features accessible t
...
 with a Pythonic, debuggable devX.
-* 👩‍🔬 **ML and data teams** who want a versioned, shared, maintainable stack of services used across
+* 👩‍🔬 **ML and data teams** who want a stable, versioned, shareable, maintainable stack of services used across
```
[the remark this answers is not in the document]
DenisYay: Agree about stable, I was trying to convey a stable version (not that the actual service is stable wrt up time), but I can see how it doesn't work in the proposed framing. Removed for now.
Re shared: I see your point, changed back, thanks.

DenisYay:
Did another pass, can you PTAL?

README.md
```diff
@@ -76,12 +75,12 @@ my_s3_folder = my_cluster_folder.to("s3", path="my_bucket/my_folder")
...
 
-You can send a function to the cluster, including the environment in which the function will live, which is actually
+You can send a [function](https://www.run.house/docs/api/python/function#function) to the cluster, including the environment in which the function will live, which is actually
 set up in its own Ray process. You can send it to an existing env, or create a new one on the fly. Like the folder
-above, the function object which is returned from `.to` is a proxy to the remote function. When you call it a
+above, the function object which is returned from `.to` is a proxy to the remote function. When you call it, a
 lightweight request is sent to the cluster's Runhouse HTTP server to execute the function with the given inputs and
 returns the results. Note that the function is not serialized, but rather imported on the cluster after the local
-working directory (`"./"`, by default the git root) is sent up.
+working directory (`"./"`, by default the git root) is sent up. This results in a significant speed reduction vs. the serialized approach.
```
dongreenberg, line 83: This is not correct, it's about package version mismatches. Probably better to just drop this.

README.md
```diff
@@ -179,15 +178,16 @@ various types of infra. You can find our currently support systems and high-leve
...
     - EC2 - **Supported**
     - SageMaker - **Supported**
     - Lambda - **In Progress**
-  - GCP - **Supported**
-  - Azure - **Supported**
+  - Google Cloud Platform (GCP) - **Supported**
+  - Microsoft Azure - **Supported**
   - Lambda Labs - **Supported**
-  - Modal - Planned
+  - Modal Labs - Planned
```
dongreenberg, line 190: Oh way better!

dongreenberg (approved):
LGTM!

README.md
DenisYay: Removed.
Should we mention what are the benefits of the non-serialized approach? Otherwise it's an implementation detail which might be unclear to the reader at this point?

PR #195, opened 2023-12-05
Secrets test refactoring

carolineechen:
* subclass resource tests
* fixtures, including parameterized provider secrets


tests/test_resources/test_secrets/conftest.py
```diff
@@ -0,0 +1,49 @@
...
+
+@pytest.fixture(scope="function")
+def custom_secret():
+    args = {"values": base_secret_values}
+    custom_secret = rh.secret(**args)
+    init_args[id(custom_secret)] = args
+    return custom_secret
+
+
+@pytest.fixture(params=providers)
```
carolineechen, line 38: @rohinb2 could you take a look and see if you have ideas on how to get this working? basically I'm hoping to have a parameterized fixture across the provider secrets to avoid manually creating like 8 provider secrets and expanding them in the pytest levels, and gets a little messy with the overarching `request.getfixturevalue secret` fixture above.

currently, setting the provider params to the `provider_secret` fixture here, and then using `lazy_fixture("provider_secret")` inside the level resource mapping to retrieve fixtures for all providers. but turns out the fixture itself (not the name) is being passed into the `secret` fixture and `request.getfixturevalue` throws an error for not being able to find it.

carolineechen: the error can be found here: https://github.com/run-house/runhouse/actions/runs/7106349953/job/19345567802?pr=195#step:5:382

PR #192, opened 2023-12-05
Add levels to test_module

dongreenberg:

PR #200, opened 2023-12-07
Add signatures to functions and modules

dongreenberg:
Ensures that we can still call a remote module or function even if the local code isn't present.
Runs with LOCAL and MINIMAL for both test_function and test_module.

### DIFF
--- a/runhouse/resources/module.py
+++ b/runhouse/resources/module.py
@@ -19,52 +19,17 @@ from runhouse.rns.utils.names import _generate_default_name
 logger = logging.getLogger(__name__)
 
 # These are methods that the Module's __getattribute__ logic should not intercept to run remotely
-LOCAL_METHODS = [
-    "RESOURCE_TYPE",
-    "__class__",
-    "__delattr__",
-    "__dict__",
-    "__dir__",
-    "__doc__",
-    "__eq__",
-    "__format__",
-    "__ge__",
-    "__gt__",
-    "__hash__",
-    "__init__",
-    "__init_subclass__",
-    "__le__",
-    "__lt__",
-    "__module__",
-    "__ne__",
-    "__new__",
-    "__reduce__",
-    "__reduce_ex__",
-    "__repr__",
-    "__setattr__",
-    "set_async",
-    "__getattribute__",
-    "__sizeof__",
-    "__str__",
-    "__subclasshook__",
-    "__weakref__",
-    "_check_for_child_configs",
-    "_cls_pointers",
+# Should we just be using inspect.getclasstree to get signatures without the Module methods instead?
+LOCAL_METHODS = dir(Resource) + [
+    "_pointers",
+    "env",
     "_env",
     "_extract_pointers",
     "_name",
-    "_resource_string_for_subconfig",
     "_rns_folder",
-    "_save_sub_resources",
+    "system",
     "_system",
-    "config_for_rns",
-    "delete_configs",
     "dryrun",
-    "env",
-    "from_config",
-    "from_name",
-    "history",
-    "is_local",
     "remote",
     "local",
     "resolve",
@@ -72,18 +37,14 @@ LOCAL_METHODS = [
     "resolved_state",
     "fetch",
     "fetch_async",
-    "name",
+    "set_async",
     "rename",
-    "rns_address",
-    "save",
-    "save_attrs_to_config",
-    "share",
-    "system",
     "to",
-    "unname",
     "provenance",
-    "refresh",
     "get_or_to",
+    "signature",
+    "_signature",
+    "method_signature",
 ]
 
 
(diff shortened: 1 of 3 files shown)


dongreenberg:
* main
  * **PR #192** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/192?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #200** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/200?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/200?utm_source=stack-comment).

dongreenberg:
@rohinb2 fix small bug so unit tests now pass

runhouse/resources/module.py
```diff
@@ -19,71 +19,32 @@
...
-    "set_async",
-    "__getattribute__",
-    "__sizeof__",
-    "__str__",
-    "__subclasshook__",
-    "__weakref__",
-    "_check_for_child_configs",
-    "_cls_pointers",
+# Should we just be using inspect.getclasstree to get signatures without the Module methods instead?
+LOCAL_METHODS = dir(Resource) + [
```
rohinb2, line 23: what happens if a user creates a class where they inherit `rh.Module` and have a method `env` on their class?

runhouse/resources/module.py
```diff
@@ -220,11 +199,49 @@ def env(self):
...
+                "property": True,
+                "async": False,
+                "gen": False,
+                "local": False,
+            }
+
+        signature = inspect.signature(method)
+        ppts = {
+            "signature": str(signature),
+            "property": not callable(method),
```
rohinb2, line 224: can set this to `False` to be clearer

runhouse/resources/module.py
```diff
@@ -220,11 +199,49 @@ def env(self):
...
+            return {
+                "signature": None,
+                "property": True,
+                "async": False,
+                "gen": False,
+                "local": False,
+            }
+
+        signature = inspect.signature(method)
+        ppts = {
```
rohinb2, line 222: better variable name than `ppts`

runhouse/resources/module.py
dongreenberg: It overrides the built-in property (which __getattribute__ will handle properly), but is still run locally rather than as an RPC if the system is a remote cluster. You can see this behavior in function.py, which overrides a bunch of things, most notably `.to`.

runhouse/resources/module.py
dongreenberg: Not sure I understand, isn't that what this will do?

runhouse/resources/module.py
```diff
@@ -220,11 +199,49 @@ def env(self):
...
+            "signature": str(signature),
+            "property": not callable(method),
+            "async": inspect.iscoroutinefunction(method)
+            or inspect.isasyncgenfunction(method),
+            "gen": inspect.isgeneratorfunction(method)
+            or inspect.isasyncgenfunction(method),
+            "local": "local" in signature.parameters
```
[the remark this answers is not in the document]
dongreenberg: It means that the method has a kwarg `local=True` which makes it run locally rather than as an RPC by default (unless overridden with `local=False`). Not very clear, gonna change it to `default_local`.

runhouse/resources/module.py
```diff
@@ -448,7 +457,7 @@ def call_wrapper():
                             **kwargs,
                         )
 
-                    if inspect.isasyncgenfunction(attr) or is_async_gen:
+                    if is_gen:
 
                         async def async_gen():
```
rohinb2, line 462: I don't quite understand this logic, and why we don't need to create a thread for this, and why we need it for the solo async call

tests/test_resources/test_modules/test_module.py
```diff
@@ -503,6 +503,134 @@ async def test_set_async(self, cluster, env):
...
+        arr = SlowNumpy(size=5)
+        assert set(arr.signature.keys()) == {
+            "slow_iter",
+            "cpu_count",
+            "size_minus_cpus",
+            "factory_constructor",
+        }
+
+        df = SlowPandas(size=10)
+        print(df.signature)
```
rohinb2, line 556: print?

runhouse/resources/module.py
rohinb2: yeah but I think including logic, aka like `not callable` when we know it is `False` is a bit cleaner, it's just a nit, either way is fine

rohinb2:
In general, looks good, I'm not extremely familiar with some of the nuances here to comment extremely deeply on logic, but good to merge

rohinb2 (approved):

runhouse/resources/module.py
dongreenberg: Let's talk about this flow live

dongreenberg:
### Merge activity

* **Dec 11, 12:07 PM**: @@dongreenberg started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/200).

PR #211, opened 2023-12-11
Fix small bug with reloading module from name

dongreenberg:
We weren't catching ModuleNotFoundErrors to default to the signature in cases where the function's
package is not present, only AttributeErrors, which is the case when a Module's backing code isn't
present.

### DIFF
--- a/runhouse/resources/module.py
+++ b/runhouse/resources/module.py
@@ -426,7 +426,7 @@ class Module(Resource):
             # Handle properties
             if is_prop:
                 return attr
-        except AttributeError as e:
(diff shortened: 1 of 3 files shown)


dongreenberg:
* main
  * **PR #211** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/211?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/211?utm_source=stack-comment).

PR #215, opened 2023-12-12
Remove add_secrets endpoint

carolineechen:
no longer needed after converting secrets to a resource type. can just use general resource endpoint


jlewitt1:
Happy to see this API go :)

rohinb2 (approved):

PR #222, opened 2023-12-13
Resource global visibility updates

jlewitt1:
- Add `resource_visibility` to rns configs for resources
- Allow for sharing globally (`public`) without specifying any users


docs/architecture.rst
```diff
@@ -86,3 +86,41 @@ Runhouse provides tools for visibility and management of resources as long-livin
...
+-------------------
+
+In addition to providing control over the access levels, Runhouse lets you control the visibility you provide
+to your users.
+
+Similar to Google Drive, you can choose how visible you would like each resource to be:
+
+- :code:`private` (default): Only users who have been granted access explictly to the resource can view and access it.
+- :code:`public`: Users can access and search for the resource by its name only (will not be vieweable by default).
+- :code:`searchable`: Resource will be available publicly to all users.
```
carolineechen, line 109: 😮

runhouse/resources/resource.py
```diff
@@ -321,36 +322,47 @@ def is_local(self):
...
-            notify_users (bool): Send email notification to users who have been given access. Defaults to `False`.
-            headers (Optional[Dict]): Request headers to provide for the request to RNS. Contains the user's auth token.
+            users (Union[str, list], optional): Single user or list of user emails and / or runhouse account usernames.
+                If none are provided and ``global_visibility`` is set to ``searchable``, resource will be made publicly
+                available to all users. Defaults to ``read``.
+            access_level (:obj:`ResourceAccess`, optional): Access level to provide for the resource.
+            global_visibility (:obj:`ResourceVisibility`, optional): Type of visibility to provide for the shared
+                resource. Defaults to ``private``.
+            notify_users (bool, optional): Whether to send an email notification to users who have been given access.
+                Defaults to ``False``.
```
carolineechen, line 347: the fn signature has this defaulting to `True` currently

runhouse/resources/resource.py
```diff
@@ -86,6 +86,7 @@ def config_for_rns(self):
             "resource_type": self.RESOURCE_TYPE,
             "resource_subtype": self.__class__.__name__,
             "provenance": self.provenance.config_for_rns if self.provenance else None,
+            "resource_visibility": self.global_visibility,
```
carolineechen, line 89: thoughts on just naming it `visibility`?

runhouse/resources/resource.py
```diff
@@ -321,36 +322,47 @@ def is_local(self):
...
         global_visibility: Optional[ResourceVisibility] = None,
         notify_users: bool = True,
         headers: Optional[Dict] = None,
     ) -> Tuple[Dict[str, ResourceAccess], Dict[str, ResourceAccess]]:
-        """Grant access to the resource for the list of users (or a single user). If a user has a Runhouse account they
+        """Grant access to the resource for a list of users (or a single user). If a user has a Runhouse account they
         will receive an email notifying them of their new access. If the user does not have a Runhouse account they will
-        also receive instructions on creating one, after which they will be able to have access to the Resource.
+        also receive instructions on creating one, after which they will be able to have access to the Resource. If
+        ``global_visibility`` is set to ``searchable``, users will not be notified.
```
carolineechen, line 334: I agree with this statement, is it being set somewhere in this PR or am I missing it?

jlewitt1: yea we set it in the `share` method if specified

carolineechen: are we setting the default `notify_users=True` to False when the visibility is set to public? I can't seem to find where this happens

jlewitt1: we don't need to since den knows how to handle that use case, but do you think we should on the OSS side anyway for clarity?

carolineechen: oh ok, I also didn't see it on the den side but I could've just missed it. but if that's the case, I think just adding a comment in the oss code mentioning should be enough

dongreenberg:
* `main`
  * **PR #222** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/222?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/222?utm_source=stack-comment).

dongreenberg (approved):

PR #231, opened 2023-12-14
Update deprecations and corresponding fn signatures

carolineechen:
* deprecate access_type, with warning, instead of BC-breaking (remove in future release)
* update function signatures w/ deprecated arguments to hide those arguments
* correct funciton_factory file name to function_factory

### DIFF
--- a/runhouse/resources/resource.py
+++ b/runhouse/resources/resource.py
@@ -326,8 +326,11 @@ class Resource:
         global_visibility: Optional[ResourceVisibility] = None,
         notify_users: bool = True,
         headers: Optional[Dict] = None,
+        access_type: Union[ResourceAccess, str] = None,  # deprecated
     ) -> Tuple[Dict[str, ResourceAccess], Dict[str, ResourceAccess]]:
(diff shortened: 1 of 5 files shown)


runhouse/resources/resource.py
```diff
@@ -377,6 +380,12 @@ def share(
                     f"For example: `{self.name}.to(system='s3')`"
                 )
 
+        if access_type:
+            logger.warning(
+                "``access_type`` argument is deprecated and will be removed in . Please use ``access_level`` instead."
```
jlewitt1, line 385:
```suggestion
                "``access_type`` argument is deprecated and will be removed. Please use ``access_level`` instead."
```

jlewitt1 (approved):

runhouse/resources/resource.py
```diff
@@ -377,6 +380,13 @@ def share(
                     f"For example: `{self.name}.to(system='s3')`"
                 )
 
+        if access_type:
+            logger.warning(
```
jlewitt1, line 384: use `warnings` instead of logger here?

PR #239, opened 2023-12-15
Login updates

carolineechen:
* don't prompt for uploading config if user doesn't have one locally
* secrets vault conversion not fully working so catch an exception for now

### DIFF
--- a/runhouse/rns/login.py
+++ b/runhouse/rns/login.py
@@ -154,8 +154,10 @@ def _login_download_secrets(headers: Optional[str] = None):
             if download_path:
                 logger.info(f"Loading down secrets for {name} into {download_path}")
                 secret.write()
-        except ValueError:
-            logger.warning(f"Was not able to load down secrets for {name}.")
+        except ValueError as e:
+            logger.warning(
+                f"Encountered {e}. Was not able to load down secrets for {name}."
+            )
 
 
 def _login_upload_secrets(interactive: bool, headers: Optional[Dict] = None):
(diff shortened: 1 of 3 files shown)


runhouse/rns/login.py
```diff
@@ -113,10 +116,11 @@ def login(
         configs.download_and_save_defaults()
         # We need to fresh the RNSClient to use the newly loaded configs
         rns_client.refresh_defaults()
-    elif upload_config:
+    if upload_config:
```
carolineechen, line 119: should we enable uploading (after downloading and saving differences) for anything that might be set locally that hasn't been reflected in den yet? or should these be mutually exclusive?

dongreenberg: I originally tried to make it so it would offer to upload the new merged config, but I think it's too complicated to explain. I think mutually exclusive is fine for now.

runhouse/rns/login.py
```diff
@@ -204,8 +209,8 @@ def _convert_secrets_resource(names: List[str] = None, headers: Optional[Dict] =
                 values = _load_vault_secrets(name, headers=headers)
                 secret = provider_secret(name, values=values)
                 secret.save()
-
-        except AttributeError:
+                _delete_vault_secrets(name=name, headers=headers)
+        except (AttributeError, Exception):
```
dongreenberg, line 213: We're kind of burrying the error here, should we add it to the warning?

dongreenberg (approved):

PR #244, opened 2023-12-18
Lambdas / Factory API smoothing

BelSasha:
Main changes:

1. Lambda function factory smoothing (removed args_names, paths_to_code, handler_funciton_name)
2. Enabling creating lambda function with a path to code using LambdaFunction.from_handler_file method.
3.  Remove tests that became redundant after the refactor.


runhouse/resources/functions/aws_lambda.py
```diff
@@ -43,13 +47,20 @@ class LambdaFunction(Function):
     DEFAULT_MEMORY_SIZE = 1024  # MB
     DEFAULT_TMP_SIZE = 3072  # MB, meaning 3G
     HOME_DIR = "/tmp/home"
+    SUPPORTED_RUNTIMES = [
```
jlewitt1, line 50: this is already defined in the factory file, maybe remove from there and just use this one?

runhouse/resources/functions/aws_lambda_factory.py
```diff
@@ -26,12 +22,9 @@
 
 def aws_lambda_fn(
     fn: Optional[Callable] = None,
-    paths_to_code: Optional[list[str]] = None,
-    handler_function_name: Optional[str] = None,
-    runtime: Optional[str] = None,
-    args_names: Optional[list[str]] = None,
     name: Optional[str] = None,
     env: Optional[dict or list[str] or Env] = None,
```
jlewitt1, line 26:
```suggestion
    env: Optional[Union[Dict, List[str], Env]] = None,
```

runhouse/resources/functions/aws_lambda.py
```diff
@@ -149,10 +162,219 @@ def from_name(cls, name, dryrun=False, alt_options=None):
...
+            paths_to_code,
+            env,
+            runtime,
+            timeout,
+            memory_size,
+            tmp_size,
+            retention_time,
+        )
+
     def _lambda_exist(self, name):
```
jlewitt1, line 378: instead of returning `True` return the function itself? that way we save an extra call to that same API later on

dongreenberg (approved):
This is awesome work!! Just had one note about the delete and delete_from_den methods - we can get rid of delete_from_den because delete_configs is already inherited from resource, and can we use `teardown` as the method to delete the lambda itself to be consistent with cluster?

PR #245, opened 2023-12-18
Add additional provider secrets

carolineechen:
Add secrets for new providers (anthropic, cohere, langchain, openai, pinecone, wandb). These secrets are all of the type `ApiSecretKey`, which is similar to the base `ProviderSecret` class, with the expected single value dict `{"api_key": "value..."}`, and env var `[PROVIDER]_API_KEY`, and by default we set env vars in `write()` and `to()` functions.

### DIFF
--- a/docs/api/python/secrets.rst
+++ b/docs/api/python/secrets.rst
@@ -42,7 +42,7 @@ AWSSecret Class
    .. autoattribute:: _DEFAULT_ENV_VARS
 
 AzureSecret Class
----------------
+-----------------
 .. autoclass:: runhouse.resources.secrets.provider_secrets.azure_secret.AzureSecret
    :show-inheritance:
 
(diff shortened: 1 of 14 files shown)


docs/api/python/secrets.rst
```diff
@@ -103,3 +103,51 @@ SkySecret Class
...
+
+OpenAISecret Class
+------------------
+.. autoclass:: runhouse.resources.secrets.provider_secrets.openai_secret.OpenAISecret
+   :show-inheritance:
+
+   .. autoattribute:: _PROVIDER
+   .. autoattribute:: _DEFAULT_ENV_VARS
+
+PineConeSecret Class
```
dongreenberg, line 139: Nit: I don't think the C is capitalized

dongreenberg (approved):
This is a sick PR, so extensible. Awesome work!!

PR #248, opened 2023-12-19
Support secrets in env

carolineechen:
* add secrets field into env, these secrets are sent to env/cluster alongside env.to or fn.to(env) calls
* add env field to cluster.sync_secrets/add_secrets
* load down env vars into env for login. write down env vars written down through runhouse
* some small bug fixes with paths and loading from name

### DIFF
--- a/runhouse/resources/envs/env_factory.py
+++ b/runhouse/resources/envs/env_factory.py
@@ -3,6 +3,8 @@ from pathlib import Path
 from typing import Dict, List, Optional, Union
 
 from runhouse.resources.packages import Package
+
+# from runhouse.resources.secrets.secret import Secret
 from .conda_env import CondaEnv
 
 from .env import Env
@@ -17,6 +19,7 @@ def env(
     setup_cmds: List[str] = None,
     env_vars: Union[Dict, str] = {},
     working_dir: Optional[Union[str, Path]] = "./",
+    secrets: Optional[Union[str, "Secret"]] = [],
     dryrun: bool = False,
 ):
     """Builds an instance of :class:`Env`.
@@ -51,7 +54,7 @@ def env(
         >>> conda_env = rh.env(conda_env="local-conda-env-name")     # from a existing local conda env
         >>> conda_env = rh.env(conda_env="conda_env.yaml", reqs=["pip:/accelerate"])   # with additional reqs
     """
-    if name and not any([reqs, conda_env, setup_cmds, env_vars]):
+    if name and not any([reqs, conda_env, setup_cmds, env_vars, secrets]):
         return Env.from_name(name, dryrun)
 
     reqs = _process_reqs(reqs or [])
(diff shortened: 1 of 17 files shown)


runhouse/resources/envs/env_factory.py
```diff
@@ -3,6 +3,8 @@
 from typing import Dict, List, Optional, Union
 
 from runhouse.resources.packages import Package
+
+# from runhouse.resources.secrets.secret import Secret
```
dongreenberg, line 7: cut?

runhouse/rns/login.py
```diff
@@ -149,11 +150,11 @@ def _login_download_secrets(headers: Optional[str] = None):
...
-
-            download_path = secret.path or secret._DEFAULT_CREDENTIALS_PATH
-            if download_path:
-                logger.info(f"Loading down secrets for {name} into {download_path}")
-                secret.write()
+            if hasattr(secret, "path"):
+                download_path = secret.path or secret._DEFAULT_CREDENTIALS_PATH
+                if download_path:
+                    logger.info(f"Loading down secrets for {name} into {download_path}")
+                    secret.write()
```
dongreenberg, line 157: If the secret doesn't have a path but has a default place for env vars, do we drop the loaded values into env vars, or ignore them?

tests/test_resources/test_envs/test_env.py
```diff
@@ -199,3 +199,35 @@ def test_working_dir_env(self, env, cluster, tmp_path):
...
+            .write(path="~/hf_token")
+            .save()
+        )
+        secrets = [path_secret, api_key_secret, named_secret.provider]
+
+        env.secrets = secrets
+        get_env_var_cpu = rh.function(_get_env_var_value).to(
+            system=cluster, env=env, force_install=True
+        )
+        env = env.to(cluster)
```
dongreenberg, line 222: This extra env.to is redunant, it's called from within fn.to above

tests/test_resources/test_envs/test_env.py
```diff
@@ -199,3 +199,35 @@ def test_working_dir_env(self, env, cluster, tmp_path):
...
+        ).write(path="~/lambda_keys")
+        api_key_secret = rh.provider_secret(
+            "openai", values={"api_key": "test_openai_key"}
+        )
+        named_secret = (
+            rh.provider_secret("huggingface", values={"token": "test_hf_token"})
+            .write(path="~/hf_token")
+            .save()
+        )
+        secrets = [path_secret, api_key_secret, named_secret.provider]
```
dongreenberg, line 216: Can we add one which is just the string name for a builtin provider?

carolineechen: that's what I'm trying to achieve with the named secret, just passing in `named_secret.provider` as a string. constructed a resource for it for the test to stay consistent, rather than needing to depend on a user's env setup or overriding any of the actual default env vars or paths.

dongreenberg: Oh oh I didn't get that, cool!

runhouse/rns/login.py
carolineechen: per discussion, will save into os.environ if called from python rh.login(). if called through runhouse login app command, will save the env vars into ~/.rh/secrets/login.env and ask the user to source env vars themselves.

PR #257, opened 2023-12-21
Install Lambda requirements using env.install()

BelSasha:


runhouse/resources/functions/aws_lambda.py
```diff
@@ -305,8 +311,17 @@ def arguments_validation(
...
+            "./" in env.reqs
+            and env.reqs[0] != "./"
+            and env.reqs[0].split("/")[0].split()[0] != env.reqs[0].split()[0]
+        ):
+            reqs = env.reqs
+            path_to_txt = str(
+                Path(reqs[0].split(":")[1] + "requirements.txt").absolute()
+            )
+            paths_to_code.append(path_to_txt)
+            env.reqs = [path_to_txt] + env.reqs[1:]
```
dongreenberg, line 324: Not totally clear on what's going on here, but looks pretty specific to one structure/use case. I think it would be a lot more robust if we could just send the packages to the lambda into the home directory, as we do with regular clusters. We can't rsync into a lambda, but can either 1) send the necessary files over the zip interface or 2) zip them ourselves and send them via a call into the lambda as we do with the env. I don't think that should be in this PR but feel it's quite critical.

runhouse/resources/functions/aws_lambda.py
```diff
@@ -433,32 +448,18 @@ def _rh_wrapper(self):
         f.write("def lambda_handler(event, context):\n")
         f.write(f"\tif not os.path.isdir('{self.HOME_DIR}'):\n")
         f.write(f"\t\tos.mkdir('{self.HOME_DIR}')\n")
+        f.write(
+            f"\tif not os.path.isdir('{self.HOME_DIR}/runhouse'):\n"
+            f"\t\tsubprocess.call(['pip', 'install', 'runhouse', '-t', '{self.HOME_DIR}/'])\n"
+            f"\tsys.path.insert(1, '{self.HOME_DIR}/')\n\n"
+        )
```
dongreenberg, line 455: Separate PR, but I continue to feel all this would be a lot easier to manage if we just put it into a file in servers/ rather than write it out like this. We can talk more about how that would look offline.

dongreenberg (commented):
Great work, glad to see you got this working!

PR #278, opened 2023-12-28
sync rh to all workers in addition to head node

jlewitt1:


jlewitt1:
* `main`
  * **PR #278** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/278?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/278?utm_source=stack-comment).

runhouse/resources/hardware/cluster.py
```diff
@@ -764,10 +780,23 @@ def restart_server(
...
         if resync_rh:
-            self._sync_runhouse_to_cluster(_install_url=_rh_install_url)
+            if not self.address:
+                raise ValueError(f"No address set for cluster <{self.name}>. Is it up?")
+
+            sync_function = partial(self._sync_to_nodes, _rh_install_url)
 
-        # Update the cluster config on the cluster
-        self.save_config_to_cluster()
+            with ProcessPoolExecutor() as executor:
```
jlewitt1, line 788: @dongreenberg you think this is overkill?

runhouse/resources/hardware/cluster.py
```diff
@@ -703,10 +717,12 @@ def restart_server(
         logger.info(f"Restarting Runhouse API server on {self.name}.")
 
         if resync_rh:
-            self._sync_runhouse_to_cluster(_install_url=_rh_install_url)
+            if not self.address:
+                raise ValueError(f"No address set for cluster <{self.name}>. Is it up?")
 
-        # Update the cluster config on the cluster
-        self.save_config_to_cluster()
+            for node in self.ips:
```
jlewitt1, line 723: For now leaving it like this until it can be unblocked - so far have tried: 
- `ProcessPoolExecutor`: Not running properly in a multiprocessing context
- `ThreadPoolExecutor`: Not running properly in multi-threaded approach, since the Live display in Rich is not designed to be used concurrently from multiple threads: 
```
        with self._lock:
            if self._live is not None:
>               raise errors.LiveError("Only one live display may be active at once")
E               rich.errors.LiveError: Only one live display may be active at once
```

runhouse/resources/hardware/sagemaker_cluster.py
```diff
@@ -1221,7 +1222,7 @@ def _stop_instance(self, delete_configs=True):
             rns_client.delete_configs(resource=self)
             logger.info(f"Deleted {self.name} from configs")
 
-    def _sync_runhouse_to_cluster(self, _install_url=None, env=None):
+    def _sync_runhouse_to_cluster(self, node: str, _install_url=None, env=None):
```
DenisYay, line 1225: Do we need to update the code below to use `node` ?

runhouse/resources/hardware/cluster.py
DenisYay: Have you tried combining `ProcessPoolExecutor` with an async event loop?

Something along the lines of:

```
        loop = asyncio.get_event_loop()
        result = await loop.run_in_executor(
            processPoolExecutor, ...
        )
```

runhouse/resources/hardware/sagemaker_cluster.py
jlewitt1: yep good catch, was missing from the case where we install from site-packages

runhouse/resources/hardware/cluster.py
```diff
@@ -1074,10 +1124,11 @@ def run_python(
         commands: List[str],
         env: Union["Env", str] = None,
         stream_logs: bool = True,
+        node: str = None,
         port_forward: Optional[int] = None,
         run_name: Optional[str] = None,
     ):
-        """Run a list of python commands on the cluster.
+        """Run a list of python commands on the cluster, or a specific cluster node if its IP is provided.
```
DenisYay, line 1131: Should we add an example below for running on a specific cluster node?

runhouse/resources/hardware/cluster.py
```diff
@@ -842,6 +858,25 @@ def disconnect(self):
         if self._rpc_tunnel:
             self._rpc_tunnel.stop()
 
+    def _sync_across_all_nodes(self, _rh_install_url):
+        loop = asyncio.new_event_loop()
+        executor = ProcessPoolExecutor()
```
DenisYay, line 863: Do we need to also call: `asyncio.set_event_loop(loop)`?

According to the documentation:
```
If there’s need to set this loop as the event loop for the current context, [set_event_loop()](https://python.readthedocs.io/en/latest/library/asyncio-eventloops.html#asyncio.set_event_loop) must be called explicitly.
```
Source:
https://python.readthedocs.io/en/latest/library/asyncio-eventloops.html#asyncio.AbstractEventLoopPolicy.new_event_loop

If the existing code works, we can probably skip it for now, but it would be good to be aware of it in case of any exceptions or weird behavior that we may face in the future.

runhouse/resources/hardware/cluster.py
```diff
@@ -842,6 +858,25 @@ def disconnect(self):
...
+                    executor, self._sync_to_nodes, _rh_install_url, address
+                )
+                for address in self.ips
+            ]
+            loop.run_until_complete(asyncio.gather(*tasks))
+
+        except Exception as e:
+            raise e
+
+        finally:
```
DenisYay, line 877: We could consider rewriting the `finally` block as follows:

```
finally:
        # Cancel all tasks running on the loop
        for task in asyncio.all_tasks(loop):
            task.cancel()
        
        # Wait for all tasks to be cancelled
        loop.run_until_complete(asyncio.gather(*asyncio.all_tasks(loop), return_exceptions=True))
        
        # Shutdown the executor
        executor.shutdown(wait=True)
        
        # Close the loop
        loop.close()
```

runhouse/resources/hardware/cluster.py
```diff
@@ -703,10 +717,12 @@ def restart_server(
         logger.info(f"Restarting Runhouse API server on {self.name}.")
 
         if resync_rh:
-            self._sync_runhouse_to_cluster(_install_url=_rh_install_url)
+            if not self.address:
```
DenisYay, line 720: Should we add a comment on what is the logical condition here e.g. in what cases would self.address be None?

runhouse/resources/hardware/cluster.py
```diff
@@ -872,16 +908,18 @@ def _rsync(
...
             source = source + "/" if not source.endswith("/") else source
             dest = dest + "/" if not dest.endswith("/") else dest
 
         ssh_credentials = copy.copy(self.ssh_creds) or {}
-        ssh_credentials.pop("ssh_host", self.address)
+        ssh_credentials.pop("ssh_host", node)
         pwd = ssh_credentials.pop("password", None)
 
-        runner = SkySSHRunner(self.address, **ssh_credentials, port=self.ssh_port)
+        runner = SkySSHRunner(node, **ssh_credentials, port=self.ssh_port)
```
DenisYay, line 922: Confirming that all worker nodes would always have `ssh_port` open?

runhouse/resources/hardware/cluster.py
```diff
@@ -994,13 +1033,23 @@ def run(
         if not run_name:
             # If not creating a Run then just run the commands via SSH and return
             return self._run_commands_with_ssh(
-                commands, cmd_prefix, stream_logs, port_forward, require_outputs
+                commands,
```
DenisYay, line 1036: nit: should `node` be first?

runhouse/resources/hardware/sagemaker_cluster.py
DenisYay, line 1225: Should it `node` explicitly default to None?

DenisYay (approved):

runhouse/resources/hardware/cluster.py
jlewitt1: good point, my understanding is that since this loop's usage is confined to this function we don't necessarily need to set it, but if we do then we should also close the loop as you added in your other comment. 

Seems like it's the recommended way to do it like this, so I'll update it!

runhouse/resources/hardware/cluster.py
jlewitt1: made it optional since we can always default to using the head node's address, do you think that API should be more explicit?

DenisYay: Optional sounds great to me, thanks

PR #280, opened 2023-12-28
Introduce traces telemetry to call_server_module() methods

DenisYay:
Example trace for /Calculator/summer:
Screenshot 2023-12-28 at 12 35 22 PM

### DIFF
--- a/runhouse/servers/http/http_server.py
+++ b/runhouse/servers/http/http_server.py
@@ -33,6 +33,7 @@ from runhouse.servers.http.http_utils import (
     pickle_b64,
     Response,
     ServerSettings,
+    telemetry_rewrite_span_name,
 )
 from runhouse.servers.nginx.config import NginxConfig
 from runhouse.servers.servlet import EnvServlet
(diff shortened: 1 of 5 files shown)


runhouse/servers/http/http_server.py
```diff
@@ -367,9 +369,14 @@ def put_resource(request: Request, message: Message):
     @staticmethod
     @app.post("/{module}/{method}")
     @validate_cluster_access
+    @tracer.start_as_current_span("call_module_method")
     def call_module_method(
         request: Request, module, method=None, message: dict = Body(default=None)
     ):
+        current_span = trace.get_current_span()
+        http_target = current_span.attributes.get("http.target", "")
```
jlewitt1, line 377: would we ever fail to load the current span here? Would cause the whole request to fail if so

DenisYay: Yes, it's possible due to various reasons, for example if the tracer object is not present or isn't initialized.
The span returned can also be of a read-only type on some machines.
These are unlikely scenarios but I think it's still good to have a validation here, thanks.

DenisYay:
This PR is not needed.

What would actually be desirable is to rewrite the root span name, not the child one in call_server_module().
But there doesn't seem to be a way to access and modify the root span in the aftermath of its creation.

So for now, dropping this PR. There may be a way to rewrite the HTTP call name on the FastAPI level, but that's out of scope for now.

PR #281, opened 2023-12-29
Fix two servlet bugs in LOCAL tests

dongreenberg:
- Auth was only flowing for named resources, corrected so a user with proper cluster access also has access to unnamed resources or non-resources via call_module_method
- Renaming resources in obj_store wasn't actually changing the internal resource.name or rns_address, which was causing auth to use the old rns_address for authentication.
- The call endpoint did not support visibility, now it does

### DIFF
--- a/runhouse/resources/module.py
+++ b/runhouse/resources/module.py
@@ -858,9 +858,9 @@ class Module(Resource):
             and isinstance(self.system, Cluster)
             and self.system.on_this_cluster()
         ):
-            obj_store.rename(old_key=old_name, new_key=self.name)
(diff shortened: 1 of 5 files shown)


dongreenberg:
* `main`
  * **PR #281** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/281?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/281?utm_source=stack-comment).

PR #294, opened 2024-01-03
Fix Sharing Behavior of Local Cluster in Unit

dongreenberg:
1. Make `setup_cluster_config` much simpler, just save down a cluster config without sharing anything. Previously, the test_account was creating the cluster and sharing it with Den Tester, which is incorrect because then we were checking if test_account would be blocked from cluster access. Also, in nearly all other logged-in tests, the Den Tester account creates the cluster and shares with Test Account, so the reverse happening here was confusing.
1. Revert the incorrect resource.revoke that was added before.
1. Rename `local_docker_cluster_...` fixtures to `docker_cluster_...`
1. Prevent configs cache from being writen as ~/.rh/config.yaml if empty
1. Skip local/no docker den_auth tests if no token is set.
1. Increase httpx latency timeout for local_client fixture, Den latency spike was causing failure when auth cache needs to be refreshed (should be addressed separately).

Still TODO (in subsequent PRs):
- [x] Rename test_account fixture to friend_test_account
- [x] Introduce logged_in_test_account fixture
- [ ] Fix all usages of test_account which should really be logged_in_test_account
- [ ] Introduce corrupted_token() and logged_out() test utils to test those behaviors more comprehensively
- [x] Allow rns_client.load_account_from_env to accept .env path so Pycharm testers don't have .env silently missed (and pass runhouse/.env by default)
- [ ] Introduce RH_DISABLE_USAGE_COLLECTION

### DIFF
--- a/tests/test_resources/test_resource.py
+++ b/tests/test_resources/test_resource.py
@@ -128,12 +128,9 @@ class TestResource:
         )  # To deal with tuples and non-json types
         for k, v in history[0]["data"].items():
             assert config[k] == v
-        resource.delete_configs()
 
     @pytest.mark.level("local")
-    def test_sharing(
-        self, resource, local_docker_cluster_pk_ssh_test_account_logged_in
-    ):
+    def test_sharing(self, resource, docker_cluster_pk_ssh_test_account_logged_in):
 
         # Unnamed resources can't even be saved, let alone shared
         if resource.name is None:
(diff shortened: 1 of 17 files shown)


dongreenberg:
* `main`
  * **PR #294** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/294?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/294?utm_source=stack-comment).

tests/test_resources/test_resource.py
```diff
@@ -157,6 +157,8 @@ def test_sharing(
                 resource.rns_address
             )
 
+        resource.revoke(users=["info@run.housee"], notify_users=False)
```
jlewitt1, line 160:
```suggestion
        resource.revoke(users=["info@run.housee"], notify_users=False)
```
```suggestion
        resource.revoke(users=["info@run.house"], notify_users=False)
```

PR #305, opened 2024-01-04
Start Ray on the head and the worker nodes

DenisYay:
The algorithm:
* We connect to the existing Ray instance started in `_start_server_cmds` and initialized inside `http_server.py`. The instance listens on port 6379 (note: the Skypilot ray instance listens on port 6380).
* We start the required number of Ray workers as part of the restart login in `cluster.py`. The workers start on the number of nodes that is derived from cluster.ips (-1 for the head node)
* We await for all workers to join the Ray cluster in `http_server.py` (verified by a call to `ray.nodes()` )
* Once the above completes successfully, we proceed with the restart flow, otherwise an error is thrown

### DIFF
--- a/runhouse/resources/hardware/cluster.py
+++ b/runhouse/resources/hardware/cluster.py
@@ -772,6 +772,8 @@ class Cluster(Resource):
             self.client.use_https = https_flag
             self.client.cert_path = self.cert_config.cert_path
 
+        # Send the ssh restart command to each of the workers
+        
         return status_codes
 
     @contextlib.contextmanager
--- a/runhouse/resources/hardware/cluster.py
+++ b/runhouse/resources/hardware/cluster.py
@@ -1,3 +1,4 @@
+import asyncio
 import contextlib
 import copy
 import json
@@ -10,6 +11,7 @@ import sys
 import threading
 import time
 import warnings
+from concurrent.futures import ProcessPoolExecutor
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple, Union
 
@@ -118,7 +120,7 @@ class Cluster(Resource):
         self.ips = self.ips or [None]
         self.ips[0] = addr
 
-    def save_config_to_cluster(self):
+    def save_config_to_cluster(self, node: str = None):
         config = self.config_for_rns
         if "live_state" in config.keys():
             # a bunch of setup commands that mess up dumping
@@ -128,7 +130,8 @@ class Cluster(Resource):
         self.run(
             [
                 f"mkdir -p ~/.rh; touch {CLUSTER_CONFIG_PATH}; echo '{json_config}' > {CLUSTER_CONFIG_PATH}"
-            ]
+            ],
+            node=node,
         )
 
     @staticmethod
@@ -253,11 +256,22 @@ class Cluster(Resource):
         )
         return self
 
-    def _sync_runhouse_to_cluster(self, _install_url=None, env=None):
-        if not self.address:
-            raise ValueError(f"No address set for cluster <{self.name}>. Is it up?")
+    def _sync_to_nodes(self, _rh_install_url: str, node: str):
+        """Sync to all cluster nodes (head node + workers where relevant)"""
+        try:
+            # Assuming _sync_runhouse_to_cluster and save_config_to_cluster are process-safe
+            self._sync_runhouse_to_cluster(_install_url=_rh_install_url, node=node)
 
+            # TODO: Deprecate once config is stored via Ray
+            # Update the cluster config on the cluster
+            self.save_config_to_cluster(node=node)
+
+        except Exception as e:
+            raise e
+
+    def _sync_runhouse_to_cluster(self, node: str = None, _install_url=None, env=None):
         local_rh_package_path = Path(pkgutil.get_loader("runhouse").path).parent
+        node = node or self.address
 
         # Check if runhouse is installed from source and has setup.py
         if (
@@ -272,6 +286,7 @@ class Cluster(Resource):
             self._rsync(
                 source=str(local_rh_package_path),
                 dest=dest_path,
+                node=node,
                 up=True,
                 contents=True,
                 filter_options="- docs/",
@@ -291,7 +306,7 @@ class Cluster(Resource):
 
         install_cmd = f"{env._run_cmd} {rh_install_cmd}" if env else rh_install_cmd
 
-        status_codes = self.run([install_cmd], stream_logs=True)
+        status_codes = self.run([install_cmd], node=node, stream_logs=True)
 
         if status_codes[0][0] != 0:
             raise ValueError(f"Error installing runhouse on cluster <{self.name}>")
@@ -760,10 +775,9 @@ class Cluster(Resource):
         logger.info(f"Restarting Runhouse API server on {self.name}.")
 
         if resync_rh:
-            self._sync_runhouse_to_cluster(_install_url=_rh_install_url)
-
-        # Update the cluster config on the cluster
-        self.save_config_to_cluster()
+            # Sync Runhouse & configs across all cluster nodes in parallel
+            self._sync_across_all_nodes(_rh_install_url)
+            logger.info("Finished syncing Runhouse to cluster nodes.")
 
         use_custom_cert = self._use_custom_cert
         if use_custom_cert:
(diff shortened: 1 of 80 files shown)


tests/fixtures/on_demand_cluster_fixtures.py
```diff
@@ -79,13 +79,17 @@ def ondemand_https_cluster_with_auth():
...
+    }
     c = rh.ondemand_cluster(**args)
     init_args[id(c)] = args
 
     c.up_if_not()
 
     c.save()
 
-    c.install_packages(["pytest"])
+    c.install_packages(["pytest", "ray"])
```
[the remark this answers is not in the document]
DenisYay: Done

tests/test_resources/test_clusters/test_multinode_cluster.py
```diff
@@ -27,3 +27,14 @@ def test_rsync_and_ssh_onto_worker_node(self, multinode_cpu_cluster):
...
 
         assert "runhouse" in status_codes[0][1]
+
+    @pytest.mark.level("thorough")
+    def test_ray_started_on_worker_node_after_cluster_restart(
+        self, multinode_cpu_cluster
+    ):
+        worker_node = multinode_cpu_cluster.ips[-1]
+
+        multinode_cpu_cluster.restart_server(restart_ray=True)
```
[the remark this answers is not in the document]
DenisYay: Agree, removed

runhouse/resources/hardware/cluster.py
```diff
@@ -698,6 +703,54 @@ def _start_server_cmds(
 
         return cmds
 
+    def _start_ray(self, host, master_host, n_hosts, ray_port):
+        if host == master_host:
+            # Head node
+            if ray.is_initialized():
```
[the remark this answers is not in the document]
DenisYay: Removed the current check.

We could run
`nmap -sV --reason -p 6381 127.0.0.1`

and check that we get

PORT     STATE SERVICE REASON         VERSION
6381/tcp open  redis?  syn-ack

Problem: we need to install nmap.

or

`nc -vv -z 127.0.0.1 6381`

Connection to domain_name 6381 port [tcp/*] succeeded!

But this result is too coarse imho (not Ray specific).

Please let me know if you have a preference between the two, but I don't think it's a blocker for a first merge.

runhouse/resources/hardware/cluster.py
```diff
@@ -698,6 +703,54 @@ def _start_server_cmds(
...
+                logger.info(
+                    f"There is a Ray cluster already running on the head node {master_host}. Shutting it down."
+                )
+                ray.shutdown()
```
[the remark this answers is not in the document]
DenisYay: Yep, removed the coarse code.

runhouse/resources/hardware/cluster.py
```diff
@@ -787,6 +840,17 @@ def restart_server(
...
+            # Restart Ray on the head node and each of the workers
+            # TODO: kill ray on all nodes first. Need to think more of the
+            # multiple multi-node clusters running on the same set of machines case.
+            master_host = self.address
+            n_hosts = len(self.ips)
+            for host in self.ips:
+                self._start_ray(host, master_host, n_hosts, self.DEFAULT_RAY_PORT)
+
+            # logger.info("🎉 All workers present and accounted for 🎉")
+            # logger.info(ray.cluster_resources())
```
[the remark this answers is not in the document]
DenisYay: +1

runhouse/resources/hardware/cluster.py
```diff
@@ -698,6 +703,54 @@ def _start_server_cmds(
...
+            worker_thread = threading.Thread(
+                target=self._wait_for_workers, args=(self, n_hosts)
+            )
+            worker_thread.start()
+
+        else:
+            # Worker node
+            self.run(
+                commands=[
+                    f"sleep 10 && ray start --address={master_host}:{ray_port} --block",
```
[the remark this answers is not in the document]
DenisYay: did you mean the on the worker nodes? I'm not sure that ray start supports that? `ray.init(address="ray://123.45.67.89:6381", namespace="runhouse")` does, but I'm not sure that that's the recommended way to connect to the cluster: https://docs.ray.io/en/latest/ray-core/configure.html#cluster-resources

runhouse/resources/hardware/cluster.py
```diff
@@ -787,6 +840,17 @@ def restart_server(
...
             self.client.cert_path = self.cert_config.cert_path
 
+        if restart_ray:
+            # Restart Ray on the head node and each of the workers
+            # TODO: kill ray on all nodes first. Need to think more of the
+            # multiple multi-node clusters running on the same set of machines case.
+            master_host = self.address
+            n_hosts = len(self.ips)
+            for host in self.ips:
+                self._start_ray(host, master_host, n_hosts, self.DEFAULT_RAY_PORT)
```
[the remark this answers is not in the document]
DenisYay: Yes, I've updated the code to start remotely on the head node and remotely on the workers.

Without the explicit start on the head node, we have two ray servers: 6380 by SkyPilot and 6379 default.

Do you mean we should use the existing 6379 default one and connect the workers to it?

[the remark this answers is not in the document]
DenisYay: As based on an offline convo, we are reusing the existing Ray instance on port 6379

runhouse/resources/hardware/cluster.py
```diff
@@ -698,6 +699,33 @@ def _start_server_cmds(
...
+        public_head_ip = master_host
+
+        # Find the internal IP corresponding to the public_head_ip
+        internal_head_ip = None
+        for internal, external in self.live_state["handle"][
+            "stable_internal_external_ips"
+        ]:
+            if external == public_head_ip:
+                internal_head_ip = internal
+                break
```
dongreenberg, line 713: Nice!

runhouse/resources/hardware/cluster.py
```diff
@@ -787,6 +815,15 @@ def restart_server(
             self.client.use_https = https_flag
             self.client.cert_path = self.cert_config.cert_path
 
+        if restart_ray:
+            # Restart Ray on the head node and each of the workers
+            # TODO: kill ray on all nodes first. Need to think more of the
```
dongreenberg, line 820: We should find out if it dies automatically on the workers when we kill it on the head or if we're really borking them when we pkill the head node.

runhouse/resources/hardware/cluster_factory.py
```diff
@@ -307,6 +307,10 @@ def ondemand_cluster(
         else open_ports
     )
 
+    # Add the new Runhouse Ray port to open_ports if it's not already there
+    if Cluster.DEFAULT_RAY_PORT not in open_ports:
+        open_ports.append(Cluster.DEFAULT_RAY_PORT)
+
```
dongreenberg, line 313: 😬 makes me nervous. If we use internal IPs do we really need to open it? Wouldn't this expose the cluster to the outside world?

tests/test_resources/test_clusters/test_multinode_cluster.py
```diff
@@ -27,3 +27,12 @@ def test_rsync_and_ssh_onto_worker_node(self, multinode_cpu_cluster):
...
         assert "runhouse" in status_codes[0][1]
+
+    @pytest.mark.level("thorough")
+    def test_ray_started_on_worker_node_after_cluster_restart(
+        self, multinode_cpu_cluster
+    ):
+        head_node = multinode_cpu_cluster.ips[0]
+
+        status_codes = multinode_cpu_cluster.run(["ray status"], node=head_node)
+        assert status_codes[0][0] == 0
```
dongreenberg, line 38: We should check that two nodes show up in the output, right?

tests/fixtures/on_demand_cluster_fixtures.py
```diff
@@ -79,7 +79,11 @@ def ondemand_https_cluster_with_auth():
 
 @pytest.fixture(scope="session")
 def multinode_cpu_cluster():
-    args = {"name": "rh-cpu-multinode", "num_instances": 2, "instance_type": "CPU:2+"}
+    args = {
+        "name": "rh-cpu-multinode-777",
```
dongreenberg, line 83: Is 777 lucky?

DenisYay: It was! But it's working now, so seems like we don't need luck any more ;)

runhouse/resources/hardware/cluster_factory.py
DenisYay: Works without, removed.

runhouse/resources/hardware/cluster.py
DenisYay: Seems like 'RAY_HEAD_IP=127.0.0.1 RAY_HEAD_PORT=6379 ray stop' should stop the entire cluster with workers when run on the head node.

tests/test_resources/test_clusters/test_multinode_cluster.py
DenisYay: Does .run() return the console output?

runhouse/resources/hardware/cluster.py
DenisYay: Nope, it doesn't seem to work..

Back to pkill for now. 

The workers do not die automatically, they become orphaned. There seems to be a way to carefully send shutdown commands to them, but need to experiment further. We can probably decouple it from the current PR?

DenisYay: Next to try: python3 -m "import ray; ray.init('ray://localhost:6379'); ray.shutdown()"
https://docs.ray.io/en/latest/ray-core/api/doc/ray.shutdown.html

If that doesn't work, maybe this is a more involved and explicit way:
https://stackoverflow.com/questions/69613739/how-to-kill-ray-tasks-when-the-driver-is-dead

PR #332, opened 2024-01-09
fixtures followed by lambda examples

BelSasha:


tests/test_resources/test_modules/test_functions/test_aws_lambda/test_lambdas_with_fn.py
```diff
@@ -14,15 +14,15 @@ def test_create_func():
     aws_func = rh.aws_lambda_fn(fn=summer, name="summer_lambdas")
     aws_func.save()
     res = aws_func(1, 3)
-    assert int(res) == 4
```
jlewitt1, line 17: what changed that we no longer have to cast it back to an int?

PR #356, opened 2024-01-17
Overhaul put_resource logic.

rohinb2:
Introduce a generic format and precedence for calls into `EnvServlet`.

All functions will take in `data` and `serialization`. They can put whatever
they want in the `data` field and unpack them as tuples before and after, as
done in this PR. This will support pickle and json serialization.

The other parameters to `EnvServlet` functions will be explicit.

The `EnvServlet` should also only call `local` ObjStore functions, and the
ObjStore should deal with all the outside-of-env logic and call the right Env.

### DIFF
--- a/runhouse/servers/env_servlet.py
+++ b/runhouse/servers/env_servlet.py
@@ -6,6 +6,7 @@ import signal
 import threading
 import time
 import traceback
+from functools import wraps
 from typing import Any, List, Optional, Union
 
 from sky.skylet.autostop_lib import set_last_active_time_to_now
@@ -19,16 +20,51 @@ from runhouse.resources.queues import Queue
 from runhouse.resources.resource import Resource
 from runhouse.rns.utils.api import ResourceVisibility
 
-from runhouse.rns.utils.names import _generate_default_name
+# from runhouse.rns.utils.names import _generate_default_name
 from runhouse.servers.http.http_utils import (
     b64_unpickle,
+    deserialize_data,
     Message,
     OutputType,
     pickle_b64,
     Response,
+    serialize_data,
 )
 
 logger = logging.getLogger(__name__)
+
+
+def error_handling_decorator(func):
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        EnvServlet.register_activity()
+        serialization = kwargs.pop("serialization", None)
+        serialized_data = kwargs.pop("data", None)
+        deserialized_data = deserialize_data(serialized_data, serialization)
+        kwargs["data"] = deserialized_data
+
+        try:
+            output = func(*args, **kwargs)
+            if output is not None:
+                serialized_data = serialize_data(output, serialization)
+                return Response(
+                    output_type=OutputType.RESULT_SERIALIZED,
+                    data=serialized_data,
+                    serialization=serialization,
+                )
+            else:
+                return Response(
+                    output_type=OutputType.SUCCESS,
+                )
+        except Exception as e:
+            logger.exception(e)
+            return Response(
+                error=pickle_b64(e),
+                traceback=pickle_b64(traceback.format_exc()),
+                output_type=OutputType.EXCEPTION,
+            )
+
+    return wrapper
 
 
 class EnvServlet:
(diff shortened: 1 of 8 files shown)


rohinb2:
* `main`
  * **PR #356** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/356?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/356?utm_source=stack-comment).

runhouse/servers/env_servlet.py
```diff
@@ -19,18 +20,53 @@
...
+                )
+            else:
+                return Response(
+                    output_type=OutputType.SUCCESS,
+                )
+        except Exception as e:
+            logger.exception(e)
+            return Response(
+                error=pickle_b64(e),
+                traceback=pickle_b64(traceback.format_exc()),
```
dongreenberg, line 63: Shouldn't this also use the serialization option passed above?

runhouse/servers/http/http_server.py
```diff
@@ -308,31 +308,23 @@ def update_settings(request: Request, message: ServerSettings) -> Response:
...
-                )
-
-        return HTTPServer.call_in_env_servlet(
-            "put_resource",
-            [message],
-            env=message.env,
-            create=True,
-            lookup_env_for_name=message.key,
-        )
+    def put_resource(request: Request, params: PutResourceParams):
```
dongreenberg, line 311: Oo pydantic validation is this the ritz?

runhouse/servers/http/http_server.py
```diff
@@ -308,31 +308,23 @@ def update_settings(request: Request, message: ServerSettings) -> Response:
...
+                serialized_data=params.serialized_data,
+                serialization=params.serialization,
+                env_name=env_name,
+            )
+        except Exception as e:
+            logger.exception(e)
+            HTTPServer.register_activity()
+            return Response(
+                error=pickle_b64(e),
+                traceback=pickle_b64(traceback.format_exc()),
```
dongreenberg, line 325: I continue to wonder out loud if we need to serialize these. Most likely not. Strings are probably fine.

dongreenberg: In the HTTPServer specifically. I don't know for the EnvServlet and obj store.

runhouse/servers/http/http_utils.py
```diff
@@ -26,6 +27,12 @@ class ServerSettings(BaseModel):
     flush_auth_cache: Optional[bool] = None
 
 
+class PutResourceParams(BaseModel):
+    serialized_data: Any
+    serialization: Optional[str] = None
+    env_name: Optional[str] = None
```
dongreenberg, line 33: Just noting this is quite generic. I'd expect that if we had pydantic types for each route we'd go more specific (e.g. not bundle into serialized_data), or we reuse a type like this for several things which share these inputs.

runhouse/servers/obj_store.py
```diff
@@ -603,3 +603,95 @@ def get_obj_refs_dict(self, d: Dict[Any, Any]):
...
+        serialization: Optional[str] = None,
+        env_name: Optional[str] = None,
+    ) -> "Response":
+        from runhouse.servers.http.http_utils import deserialize_data
+
+        if env_name is None:
+            if self.servlet_name is None:
+                raise ObjStoreError("No env name provided and no servlet name set.")
+            else:
+                env_name = self.servlet_name
```
dongreenberg, line 622: If we're inside this env, should we just call self.put_resource_local directly? No reason to serialize, deserialize, and add a ray call to the servlet if we're already in the right env process?

tests/test_servers/test_http_server.py
```diff
@@ -58,7 +58,9 @@ def test_put_resource(self, http_client, blob_data, cluster):
         resource = rh.blob(data=blob_data, system=cluster)
         data = pickle_b64((resource.config_for_rns, state, resource.dryrun))
         response = http_client.post(
-            "/resource", json={"data": data}, headers=rns_client.request_headers()
+            "/resource",
+            json=PutResourceParams(serialized_data=data, serialization="pickle").dict(),
```
dongreenberg, line 62: 👁️👄👁️

dongreenberg (approved):
Overall excellent, much cleaner!

runhouse/servers/env_servlet.py
rohinb2: currently we deserialize all exceptions the same as using pickle, client side

runhouse/servers/http/http_server.py
rohinb2: I can remove this later

runhouse/servers/http/http_utils.py
rohinb2: It'd be harder to make the serialization arg general then

runhouse/servers/obj_store.py
rohinb2: fixed

rohinb2:
### Merge activity

* **Jan 19, 11:52 AM**: @rohinb2 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/356).

PR #360, opened 2024-01-18
Move scripts and docker out of source folder

carolineechen:
We should reserve the runhouse/ folder for package source code, moving auxiliary files like scripts and docker out into the root branch.


carolineechen:
* `main`
  * **PR #360** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/360?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/360?utm_source=stack-comment).

runhouse/resources/hardware/sagemaker_cluster.py
```diff
@@ -930,7 +930,7 @@ def _refresh_ssm_session_with_cluster(self, num_ports_to_try: int = 5):
             self._create_ssm_session_with_cluster()
 
         # https://github.com/aws-samples/sagemaker-ssh-helper/blob/main/sagemaker_ssh_helper/sm-connect-ssh-proxy
-        full_module_name = "scripts/sagemaker_cluster/refresh-ssm-session.sh"
+        full_module_name = "resources/hardware/scripts/refresh-ssm-session.sh"
```
jlewitt1, line 933: all these scripts are sagemaker specific, should we create another subfolder within `hardware`? will matter more if we add other cluster scripts at some point

carolineechen: yea that makes sense, scripts is a little generic here. just clarifying, are you thinking:

1. `resources/hardware/sagemaker` that contains the scripts + sagemaker_cluster.py
2. `resources/hardware/scripts/sagemaker` with sagemaker scripts, and the sagemaker_cluster.py stays in `hardware/`
3. something else

this might also be subject to change if we do add other cluster/resource scripts but should be easy to move around

jlewitt1: good point, for now feels a little overkill to bundle the scripts with the cluster file, so I think option 1 makes sense

jlewitt1 (approved):

PR #367, opened 2024-01-19
Change Module's endpoint from propery to function

dongreenberg:
This allows us to set external=True or False, in case we want a locally-callable
endpoint. Previously if the endpoint was only locally callable, we would only return
None. This is now also consistent with Cluster's endpoint().

### DIFF
--- a/runhouse/resources/hardware/cluster.py
+++ b/runhouse/resources/hardware/cluster.py
@@ -176,7 +176,7 @@ class Cluster(Resource):
         return config
 
     def endpoint(self, external=False):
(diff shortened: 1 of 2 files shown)


dongreenberg:
* `main`
  * **PR #367** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/367?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/367?utm_source=stack-comment).

PR #368, opened 2024-01-19
Improve `_current_cluster` to initialize object store and handle a local cluster.

rohinb2:
If we are running a Python script on a cluster, the object store may not be
initialized. We used to initialize the object store within `__init__.py`, but
that is now removed to prevent circular dependency issues. 

So now, if a user requests the `_current_cluster` as a part of their script/flow and the object store isn't already initialized, then make sure to initialize it.
Note that in a server or a servlet, `cluster_servlet` would already not be
`None`, so we wouldn't re-initialize it and set the name to `None`.

Also change `_current_cluster` to return `config` by default, and check if `name` is `None`.

### DIFF
--- a/runhouse/resources/hardware/utils.py
+++ b/runhouse/resources/hardware/utils.py
@@ -101,13 +101,22 @@ def load_cluster_config_from_file() -> Dict:
         return {}
 
 
-def _current_cluster(key="name"):
+def _current_cluster(key="config"):
     """Retrive key value from the current cluster config.
     If key is "config", returns entire config."""
     from runhouse.globals import obj_store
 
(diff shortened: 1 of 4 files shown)


rohinb2:
* `main`
  * **PR #368** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/368?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/368?utm_source=stack-comment).

PR #373, opened 2024-01-22
Add call, call_local, and call_for_env_servlet_name to obj store

dongreenberg:
Not yet tested or at parity with env servlet calls. Introduce FutureModules as purpose-built alternatives to saving async or generator results in blobs and queues.

### DIFF
--- a/runhouse/resources/module.py
+++ b/runhouse/resources/module.py
@@ -148,7 +148,7 @@ class Module(Resource):
                     # and this may a "type" module rather than an "instance". The user might instantiate it later, or
                     # it may be populated with attributes by the servlet's put_resource.
                     module_cls = _module_subclass_factory(
-                        module_cls, config.get("pointers"), remote_init=True
+                        module_cls, config.get("pointers")
                     )
             except ModuleNotFoundError:
                 # If we fail to construct the module class from the pointers, we check if the code is not local,
@@ -1014,7 +1014,7 @@ class Module(Resource):
     #     full_name = func_name
 
 
-def _module_subclass_factory(cls, cls_pointers, remote_init=False):
+def _module_subclass_factory(cls, cls_pointers):
     def __init__(
         self,
         *args,
@@ -1041,8 +1041,14 @@ def _module_subclass_factory(cls, cls_pointers, remote_init=False):
         )
         # This allows a class which is already on the cluster to construct an instance of itself with a factory
         # method, e.g. my_module = MyModuleCls.factory_constructor(*args, **kwargs)
-        if self.system and self.system.on_this_cluster() and remote_init:
+        if self.system and self.system.on_this_cluster():
             self._remote_init(*args, **kwargs)
+
+    @classmethod
+    def _module_init_only(cls, *args, **kwargs):
+        new_module = cls.__new__(cls)
+        Module.__init__(new_module, *args, **kwargs)
+        return new_module
 
     def __call__(
         self,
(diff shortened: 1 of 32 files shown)


dongreenberg:
* `main`
  * **PR #373** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/373?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/373?utm_source=stack-comment).

runhouse/resources/module.py
```diff
@@ -603,20 +578,23 @@ def local(self, *args, **kwargs):
...
-        if not self._client() or not self._name:
-            return super().__setattr__(key, value)
-
-        return self._client().call(
-            module_name=self._name,
-            method_name=key,
-            new_value=value,
-            stream_logs=False,
-        )
+    # def __setattr__(self, key, value):
```
rohinb2, line 581: can we land this with this commented out?

runhouse/servers/obj_store.py
```diff
@@ -753,7 +754,212 @@ def rename(self, old_key: Any, new_key: Any):
...
+                data=data,
+                serialization=serialization,
+                run_name=run_name,
+                stream_logs=stream_logs,
+                remote=remote,
+            )
+
+        if remote and isinstance(res, dict):
+            config = res
+            if config["system"] == self.get_cluster_config():
```
rohinb2, line 950: how would the `config["system"]` be a config and not an initialized cluster object? people would only set that up via `.to(rh.here)` or `(..., system=rh.here)` right?

runhouse/servers/env_servlet.py
```diff
@@ -48,26 +41,21 @@ def wrapper(*args, **kwargs):
...
+                    output_type=OutputType.RESULT_SERIALIZED,
+                    data=serialized_data,
+                    serialization=serialization,
+                )
             else:
                 return Response(
                     output_type=OutputType.SUCCESS,
                 )
         except Exception as e:
-            if serialization is None:
```
rohinb2, line 66: why don't we want to `raise e` here directly?

tests/conftest.py
```diff
@@ -300,6 +301,7 @@ def logged_in_account():
 }
```
rohinb2, line 301: <i class='graphite__hidden'>[Re: line 297]<br/><br/></i>

not adding `local_daemon` to unit? to keep up the speed i assume<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/run-house/runhouse/373?utm_source=unchanged-line-comment">Graphite</a>.</span>

runhouse/servers/http/http_server.py
```diff
@@ -66,50 +68,47 @@ async def wrapper(*args, **kwargs):
...
-
-        if token is None:
-            raise HTTPException(
-                status_code=404,
-                detail="No token found in request auth headers. Expected in "
-                f"format: {json.dumps({'Authorization': 'Bearer <token>'})}",
-            )
+        request_id = request.headers.get("X-Request-ID", str(uuid.uuid4()))
+        token_hash = hash_token(token) if den_auth_enabled and token else None
+        _ctx_token = obj_store._set_ctx(request_id=request_id, token_hash=token_hash)
```
rohinb2, line 76: nit: why do we prefix this, why not just call this `ctx_token`

rohinb2, line 76: also `._set_ctx` and `._get_ctx` should not be prefixed with underscores since they're called outside the obj store

runhouse/servers/env_servlet.py
dongreenberg: Inside `handle_exception_response` we have special handling for certain exceptions that we want to convert into HTTPExceptions (e.g. permission error -> 403), and barring that if serialization is None we just raise the exception as you're suggesting.

tests/conftest.py
dongreenberg: The vast majority of tests which apply to `local_daemon` as a proper cluster are currently in LOCAL. Putting it in UNIT would mean running a bunch of UNIT tests on a cluster that aren't really used to running on a cluster. I figured this is less disruptive and we can adjust it later if it works fine.

PR #377, opened 2024-01-23
Fix defaulting behavior of cluster factories and classes

dongreenberg:


dongreenberg:
* `main`
  * **PR #377** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/377?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/377?utm_source=stack-comment).

carolineechen (commented):
general comment about the way we're doing cluster factory functions right now, but could we get rid of kwargs for specific factories like ondemand/kubernetes/sagemaker? I get the point for the generic cluster class if we're passing in those args into more specific cluster factories, but for the others I think we should just include the supported variables as documented fn args and not support random additional kwargs that won't be passed in. this also made us have to be extra careful with parsing kwargs and setting constraints as we've done with the k8s factory

cc @RohanSreerama5

runhouse/resources/hardware/cluster_factory.py
```diff
@@ -85,51 +79,27 @@ def cluster(
...
+            server_connection_type=server_connection_type,
+            ssl_keyfile=ssl_keyfile,
+            ssl_certfile=ssl_certfile,
+            den_auth=den_auth,
+            kwargs=kwargs,
         )
+        # Filter out None/default values
+        alt_options = {k: v for k, v in alt_options.items() if v is not None}
+        try:
+            c = Cluster.from_name(name, dryrun, alt_options=alt_options)
```
rohinb2, line 97: we want to enable this `alt_options` behavior right? this is a changefrom before

rohinb2:
this seems very straightforward

rohinb2 (approved):

runhouse/resources/hardware/cluster_factory.py
dongreenberg: Y

dongreenberg:
This failure is just because I was running the tests locally at the same time so the !rh.exists failed. Merging.

PR #386, opened 2024-01-24
Avoid backgrounding by default when doing `runhouse start`.

rohinb2:
With the addition of `nohup` if `screen` is not available, we were
backgrounding by default when running `runhouse start`. Now we should only do
it if `runhouse start --nohup` was passed.

### DIFF
--- a/runhouse/main.py
+++ b/runhouse/main.py
@@ -100,6 +100,7 @@ def _start_server(
     restart,
     restart_ray,
     screen,
+    nohup,
     create_logfile=True,
     host=None,
     port=None,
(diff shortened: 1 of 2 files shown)


rohinb2:
* `main`
  * **PR #386** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/386?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/386?utm_source=stack-comment).

rohinb2:
if you run `runhouse start` locally, on `main`, right now, it uses `nohup`. 

We want to avoid backgrounding when running `runhouse start`. This PR just makes it optional to background.

PR #390, opened 2024-01-24
Simplify `rh.here` Ray cluster detection logic.

rohinb2:
The logic that we previously had to detect if we are on a cluster was extra
tricky because we would initializing the `rh.here` variable upon `import
runhouse as rh`, so there would be circular situations where we need to avoid
deadlocks in Actor creation.

However, now this is almost always being called in a process where one should
easily be able to connect to Ray and get the `cluster_servlet` if it exists. We
don't need any fancy `list_actors` nonsense or an extra `ray.init`.

This is also part of a more general theme of trying to trim down the excess
`ray.init`s that we have in the code to simplify the `ray start` sequence.

### DIFF
--- a/runhouse/rns/top_level_rns_fns.py
+++ b/runhouse/rns/top_level_rns_fns.py
@@ -2,9 +2,6 @@ import logging
 import sys
 from typing import List
 
-import ray
-from ray.experimental.state.api import list_actors
-
 from runhouse.globals import configs, obj_store, rns_client
 
 from runhouse.logger import LOGGING_CONFIG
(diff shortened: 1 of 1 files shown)


rohinb2:
* `main`
  * **PR #390** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/390?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/390?utm_source=stack-comment).

PR #396, opened 2024-01-25
Adds tqdm and s3fs to CI jobs

RohanSreerama5:
Adds s3fs and tqdm dependancies to CI jobs. This is useful for running Folder tests.

### DIFF
--- a/.github/workflows/setup_runhouse/action.yaml
+++ b/.github/workflows/setup_runhouse/action.yaml
@@ -24,4 +24,4 @@ runs:
     - name: Install python packages & dependencies
       shell: bash
       run:
-        pip install pytest httpx pytest_asyncio docker
+        pip install pytest httpx pytest_asyncio docker s3fs tqdm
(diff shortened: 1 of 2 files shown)


.github/workflows/setup_runhouse/action.yaml
```diff
@@ -24,4 +24,4 @@ runs:
     - name: Install python packages & dependencies
       shell: bash
       run:
-        pip install pytest httpx pytest_asyncio docker
+        pip install pytest httpx pytest_asyncio docker s3fs tqdm
```
jlewitt1, line 27: CI is throwing boto credentials errors, maybe we need to configure aws with an access key and secret access key from secrets? 

could be something like: 
```
          aws configure set aws_access_key_id ${{ secrets.AWS_ACCESS_KEY }}
          aws configure set aws_secret_access_key ${{ secrets.AWS_SECRET_KEY }}
          aws configure set default.region us-east-1
```

PR #398, opened 2024-01-26
Update _get_env_from to handle cases better

carolineechen:
improve 2 cases of _get_env_from, which is expected to return an `Env` type:
1. input env = "base_env" -- falls through all the if statements and returns the string "base_env" instead of the Env object
2. env is not saved in rns_client but exists on the cluster associated with the call. returns the env string instead of Env object

### DIFF
--- a/runhouse/resources/envs/utils.py
+++ b/runhouse/resources/envs/utils.py
@@ -37,7 +37,7 @@ def _process_reqs(reqs):
     return preprocessed_reqs
 
 
-def _get_env_from(env):
+def _get_env_from(env, system: "Cluster" = None):
     if isinstance(env, Resource) or env is None:
         return env
 
@@ -49,12 +49,28 @@ def _get_env_from(env):
         return Env(reqs=env, working_dir="./", name=Env.DEFAULT_NAME)
     elif isinstance(env, Dict):
         return Env.from_config(env)
-    elif isinstance(env, str) and rns_client.exists(env, resource_type="env"):
-        return Env.from_name(env)
-    elif env == Env.DEFAULT_NAME:
-        return Env(name=Env.DEFAULT_NAME)
+    elif isinstance(env, str):
+        if rns_client.exists(env, resource_type="env"):
+            return Env.from_name(env)
+        if system:
+            if system.get(env):
+                return system.get(env)
 
-    raise ValueError("Could not locate Env from value: {env}")
+            # handle case where env is passed in as full rns address
+            name, _ = rns_client.split_rns_name_and_path(
+                rns_client.resolve_rns_path(env)
+            )
+            if system.get(name):
+                return system.get(name)
+
+        if env == Env.DEFAULT_NAME:
+            return Env(name=Env.DEFAULT_NAME)
+
+    warning = f"Could not locate Env from value: {env}"
+    warning = f"{warning} on cluster {system.name}" if system else warning
+    logging.warning(warning)
+
+    return env
 
 
 def _get_conda_yaml(conda_env=None):
--- a/runhouse/resources/envs/utils.py
+++ b/runhouse/resources/envs/utils.py
@@ -38,7 +38,7 @@ def _process_reqs(reqs):
 
 
 def _get_env_from(env):
-    if isinstance(env, Resource):
+    if isinstance(env, Resource) or env is None:
         return env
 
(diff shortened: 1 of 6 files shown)


carolineechen:
* `main`
  * **PR #398** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/398?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/398?utm_source=stack-comment).

runhouse/resources/envs/utils.py
```diff
@@ -49,12 +49,27 @@ def _get_env_from(env):
...
+            )
+            if system.get(name):
+                return system.get(name)
+
+        if env == Env.DEFAULT_NAME:
+            return Env(name=Env.DEFAULT_NAME)
+
+    warning = f"Could not locate Env from value: {env}"
+    warning = f"{warning} on cluster {system.name}" if system else warning
+    logging.warning(warning)
```
carolineechen, line 71: @dongreenberg can you double check my logic here? and is there an instance where it still makes sense to return the input `env` if it doesn't satisfy any of the conditions or to simply throw an error?

most cases we expect an `Env` to be returned which will throw an error elsewhere if it returns a string and we try to call an env property on it. but there could also be cases where the output is simply used to populate a config etc, not sure if it's an issue if a string is returned if the corresponding Env could not be found.

runhouse/resources/envs/utils.py
```diff
@@ -37,8 +37,8 @@ def _process_reqs(reqs):
     return preprocessed_reqs
 
 
-def _get_env_from(env):
-    if isinstance(env, Resource):
+def _get_env_from(env, system: "Cluster" = None):
+    if isinstance(env, Resource) or env is None:
         return env
 
     from runhouse.resources.envs import Env
```
dongreenberg, line 44:
```suggestion
    from runhouse.resources.envs import Env

    if isinstance(env, Env) or env is None:
        return env

```

runhouse/resources/envs/utils.py
```diff
@@ -49,12 +49,27 @@ def _get_env_from(env):
...
+        if system:
+            if system.get(env):
+                return system.get(env)
+
+            # handle case where env is passed in as full rns address
+            name, _ = rns_client.split_rns_name_and_path(
+                rns_client.resolve_rns_path(env)
+            )
+            if system.get(name):
+                return system.get(name)
```
dongreenberg, line 64:
```suggestion
            # handle case where env is passed in as full rns address
            name, _ = rns_client.split_rns_name_and_path(
                rns_client.resolve_rns_path(env)
            ) if "/" in env else (env, None)
            if system.get(name):
                return system.get(name)
```

dongreenberg: Just saves a `get`

runhouse/resources/envs/utils.py
dongreenberg: tbh I'm nervous that passing system here is a little non-deterministic. Den and this system could have different configs for that env name, and the only thing determining which is returned is the order we've implemented here. What case are we solving for where we want to check on the cluster for the env config?

I also don't think I'd raise a warning here. We use this method all the time in instances where we don't necessarily expect and env to be found (but should construct one if it is), and it'll be pretty noisy for the user.

runhouse/resources/hardware/cluster.py
```diff
@@ -317,7 +318,7 @@ def install_packages(
         from runhouse.resources.envs.env import Env
 
         self.check_server()
-        env = _get_env_from(env) or Env(name=env or Env.DEFAULT_NAME)
+        env = _get_env_from(env, system=self) or Env(name=env or Env.DEFAULT_NAME)
```
dongreenberg, line 321: This makes me a bit nervous - if the env contains package objects they will be serialized and sent back pointing to the cluster's local filesystem. Also not totally clear what's happening here, we're getting the env object from the system by name so we can then extract the name out of it?

dongreenberg (commented):
Tbh I think I'm missing some context, but there's some back and forth here with the cluster that seems a bit hairy to me. Let's chat about it live

PR #401, opened 2024-01-26
Add deleting env servlet functionality

carolineechen:
Support deleting and killing and env servlet through `cluster.delete(["env_key"])`.

Clears the keys existing within the env servlet, removes the env servlet from cluster servlet (and anywhere that keeps track of active env servlets), and kills the corresponding actor.

### DIFF
--- a/runhouse/servers/obj_store.py
+++ b/runhouse/servers/obj_store.py
@@ -594,10 +594,6 @@ class ObjStore:
 
             del env_servlets[env_name]
 
-        # delete the local key
-        if self.contains_local(env_name):
-            self.delete_local(env_name)
-
         self.remove_env_servlet_name(env_name)
         return env_servlet_keys
 
--- a/runhouse/servers/obj_store.py
+++ b/runhouse/servers/obj_store.py
@@ -580,34 +580,35 @@ class ObjStore:
     def delete_local(self, key: Any):
         self.pop_local(key)
 
-    def delete_env(self, env_name: Any):
+    def _delete_env_contents(self, env_name: Any):
         from runhouse.globals import env_servlets
 
         # clear keys in the env servlet
-        env_servlet_keys = self.keys_for_env_servlet_name(env_name)
+        deleted_keys = self.keys_for_env_servlet_name(env_name)
         self.clear_for_env_servlet_name(env_name)
 
-        # delete the env servlet actor
+        # delete the env servlet actor and remove its references
         if env_name in env_servlets:
             actor = env_servlets[env_name]
             ray.kill(actor)
 
             del env_servlets[env_name]
-
         self.remove_env_servlet_name(env_name)
-        return env_servlet_keys
+
+        return deleted_keys
 
     def delete(self, key: Union[Any, List[Any]]):
         keys_to_delete = [key] if isinstance(key, str) else key
         deleted_keys = []
 
         for key_to_delete in keys_to_delete:
+            if key_to_delete in self.get_all_initialized_env_servlet_names():
+                deleted_keys += self._delete_env_contents(key_to_delete)
+
             if key_to_delete in deleted_keys:
                 continue
 
-            if key_to_delete in self.get_all_initialized_env_servlet_names():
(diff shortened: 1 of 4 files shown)


carolineechen:
* `main`
  * **PR #401** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/401?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/401?utm_source=stack-comment).

runhouse/servers/obj_store.py
```diff
@@ -572,11 +580,40 @@ def delete_for_env_servlet_name(env_servlet_name: str, key: Any):
     def delete_local(self, key: Any):
         self.pop_local(key)
 
+    def delete_env(self, key: Any):
```
rohinb2, line 583: can we rename key to `env_name` or something

runhouse/servers/obj_store.py
```diff
@@ -572,11 +580,40 @@ def delete_for_env_servlet_name(env_servlet_name: str, key: Any):
...
+        from runhouse.globals import env_servlets
+
+        # clear keys in the env servlet
+        env_servlet_keys = list(self._kv_store.keys()) if self.has_local_storage else []
```
rohinb2, line 587: these are just the keys within this servlet, unrelated to the servlet that we are trying to delete, right? why would we mark these keys as deleted? bit confused

rohinb2:
I think it's also worthwhile to have a test in `test_cluster` that serves more as an integration test. 

You can make an env, do, `cluster.put(env1)` then put a bunch of keys in it, then make another env, do, `cluster.put(env2)`, put a bunch of keys there, put a bunch of keys without an env (which should go to `base`) and then delete one and assert that everything that should have been deleted was deleted.

runhouse/servers/obj_store.py
rohinb2, line 587: maybe this should be `self.keys_for_env_servlet_name(key)`?

runhouse/servers/obj_store.py
```diff
@@ -572,11 +580,40 @@ def delete_for_env_servlet_name(env_servlet_name: str, key: Any):
...
+        self.remove_env_servlet_name(key)
+        return env_servlet_keys
+
     def delete(self, key: Union[Any, List[Any]]):
         keys_to_delete = [key] if isinstance(key, str) else key
+        deleted_keys = []
+
         for key_to_delete in keys_to_delete:
-            if self.contains_local(key_to_delete):
+            if key_to_delete in deleted_keys:
```
rohinb2, line 609: ah i understand, technically you could have already deleted something they also wanted to delete

rohinb2: nice

runhouse/servers/obj_store.py
```diff
@@ -572,11 +580,40 @@ def delete_for_env_servlet_name(env_servlet_name: str, key: Any):
...
+
+        # delete the env servlet actor
+        if env_name in env_servlets:
+            actor = env_servlets[env_name]
+            ray.kill(actor)
+
+            del env_servlets[env_name]
+
+        # delete the local key
+        if self.contains_local(env_name):
```
rohinb2, line 598: I'm confused how this is deleting the env name (it's passing your tests so it seems to be, just confused).

Let's say you're in the server, which has no local storage, which calls `delete_env("env2")`. Then I see that we clear all the keys for `env2`, but wouldn't it not find `env2` in local storage, and therefore `env2` would still  be a key in `base` pointing to `Env(...)`?

carolineechen: good point, looks like this branch is basically a no-op that I didn't realize, and the env name is being deleted either way. I'm a little bit confused where exactly that is happening though when `cluster.delete(env_name)` is called -- do you have an idea? bc the env key shouldn't be part of the `keys_for_env_servlet_name` that's cleared.

PR #403, opened 2024-01-27
deduplicates test code by assigning folder types to test level types

RohanSreerama5:
[WIP]

- deduplicates test code by assigning folder types to test level types
    - generalizes testing to be more cloud agnostic
- adds a `local_folder_docker` fixture
- expands folder to folder testing. Tests across 25 different folder transfer combinations for 
 `local_folder`, `local_folder_docker`, `cluster_folder`, `s3_folder`, and `gcs_folder`
- updates folder fixtures to use `init_args` 
- updates setting of `self._path` in `Folder` class, so that path is set properly even if the `system` passed in is of a `Cluster` type but was passed in the form of a `str`. 
- updates `sshfs` dependency to `sshfs >= 2023.1.0, <= 2023.4.1`

### DIFF
--- a/tests/fixtures/folder_fixtures.py
+++ b/tests/fixtures/folder_fixtures.py
@@ -8,6 +8,12 @@ from .utils import create_gcs_bucket, create_s3_bucket
 
 
 @pytest.fixture
+def dest(request):
+    """Parametrize over multiple folders - useful for running the same test on multiple storage types."""
+    return request.getfixturevalue(request.param)
+
+
+@pytest.fixture
 def folder(request):
     """Parametrize over multiple folders - useful for running the same test on multiple storage types."""
     return request.getfixturevalue(request.param)
@@ -23,13 +29,47 @@ def local_folder(tmp_path):
 
 
 @pytest.fixture
-def cluster_folder(ondemand_cpu_cluster, local_folder):
-    return local_folder.to(system=ondemand_cpu_cluster)
+def local_folder_docker(docker_cluster_pk_ssh_no_auth):
+    args = {
+        "name": "test_docker_folder",
+        "system": docker_cluster_pk_ssh_no_auth,
+        "path": "rh-folder",
+    }
+
+    local_folder_docker = rh.folder(**args)
+    init_args[id(local_folder_docker)] = args
+    local_folder_docker.put(
+        {f"sample_file_{i}.txt": f"file{i}".encode() for i in range(3)}
+    )
+    return local_folder_docker
 
 
 @pytest.fixture
-def s3_folder(local_folder):
-    s3_folder = local_folder.to(system="s3")
+def cluster_folder(ondemand_cpu_cluster):
+    args = {
+        "name": "test_cluster_folder",
+        "system": ondemand_cpu_cluster,
+        "path": "rh-folder",
+    }
+
+    cluster_folder = rh.folder(**args)
+    init_args[id(cluster_folder)] = args
+    cluster_folder.put({f"sample_file_{i}.txt": f"file{i}".encode() for i in range(3)})
+    return cluster_folder
+
+
+@pytest.fixture
+def s3_folder():
+    args = {
+        "name": "test_s3_folder",
+        "system": "s3",
+        "path": "rh-folder",
+    }
+
+    s3_folder = rh.folder(**args)
+    init_args[id(s3_folder)] = args
+    s3_folder.put({f"sample_file_{i}.txt": f"file{i}".encode() for i in range(3)})
+
     yield s3_folder
 
     # Delete files from S3
@@ -37,8 +77,17 @@ def s3_folder(local_folder):
 
 
 @pytest.fixture
-def gcs_folder(local_folder):
-    gcs_folder = local_folder.to(system="gs")
+def gcs_folder():
+    args = {
+        "name": "test_gcs_folder",
+        "system": "gs",
+        "path": "rh-folder",
+    }
+
+    gcs_folder = rh.folder(**args)
+    init_args[id(gcs_folder)] = args
+    gcs_folder.put({f"sample_file_{i}.txt": f"file{i}".encode() for i in range(3)})
+
     yield gcs_folder
 
     # Delete files from GCS
--- a/tests/test_resources/test_modules/test_folder.py
+++ b/tests/test_resources/test_modules/test_folder.py
@@ -17,6 +17,89 @@ def fs_str_rh_fn(folder):
 
 
 class TestFolder(tests.test_resources.test_resource.TestResource):
+
+    MAP_FIXTURES = {"resource": "folder"}
+
+    UNIT = {
+        "folder": ["local_folder"],
+        "dest": ["local_folder"],
+    }
+    LOCAL = {
+        "folder": ["local_folder", "local_folder_docker"],
+        "dest": ["local_folder", "local_folder_docker"],
+    }
+    MINIMAL = {
+        "folder": [
+            "local_folder",
+            "local_folder_docker",
+            "cluster_folder",
+            "s3_folder",
+            "gcs_folder",
+        ],
+        "dest": [
+            "local_folder",
+            "local_folder_docker",
+            "cluster_folder",
+            "s3_folder",
+            "gcs_folder",
+        ],
+    }
+    THOROUGH = {
+        "folder": [
+            "local_folder",
+            "local_folder_docker",
+            "cluster_folder",
+            "s3_folder",
+            "gcs_folder",
+        ],
+        "dest": [
+            "local_folder",
+            "local_folder_docker",
+            "cluster_folder",
+            "s3_folder",
+            "gcs_folder",
+        ],
+    }
+    MAXIMAL = {
+        "folder": [
+            "local_folder",
+            "local_folder_docker",
+            "cluster_folder",
+            "s3_folder",
+            "gcs_folder",
+        ],
+        "dest": [
+            "local_folder",
+            "local_folder_docker",
+            "cluster_folder",
+            "s3_folder",
+            "gcs_folder",
+        ],
+    }
+
+    @pytest.mark.level("unit")
+    def test_send_folder_to_dest(self, folder, dest):
+
+        if (
+            isinstance(folder.system, rh.Cluster)
+            and isinstance(dest.system, rh.Cluster)
+            and folder.system.name == "docker_cluster_pk_ssh"
+            and dest.system.name == "docker_cluster_pk_ssh"
+        ):
+            pytest.skip(
+                "Test skipped when both folder and dest are set to local_folder_docker"
+            )
+
+        system = "here" if dest.system == "file" else dest.system
+        expected_fs_str = "ssh" if isinstance(dest.system, rh.Cluster) else dest.system
+
+        new_folder = folder.to(system=system)
+
+        assert new_folder._fs_str == expected_fs_str
+        assert "sample_file_0.txt" in new_folder.ls(full_paths=False)
+
+        new_folder.rm()
+
     @pytest.mark.level("local")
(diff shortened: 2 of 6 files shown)


tests/test_resources/test_modules/test_folder.py
```diff
@@ -17,6 +17,19 @@ def fs_str_rh_fn(folder):
 
 
 class TestFolder(tests.test_resources.test_resource.TestResource):
+
+    MAP_FIXTURES = {"resource": "folder"}
+
+    UNIT = {"folder": ["s3_folder", "gcs_folder"]}
+    LOCAL = {"folder": ["s3_folder", "gcs_folder"]}
```
carolineechen, line 24: local should include testing folder interaction with a system of our local test docker container. either creating a fixture for some `local_cluster_folder`, or further parameterizing this with a cluster argument as well. And local is run on CI so until we get support for s3 and gcs there, those should remain in minimal up

tests/test_resources/test_modules/test_folder.py
```diff
@@ -17,6 +17,19 @@ def fs_str_rh_fn(folder):
...
+    MINIMAL = {"folder": ["cluster_folder", "s3_folder", "gcs_folder"]}
+    THOROUGH = {
```
carolineechen, line 26: thorough and maximal should include cases too; otherwise passing in an empty list will cause it to not parameterize the folder argument at all in this level

tests/test_resources/test_modules/test_folder.py
```diff
@@ -182,49 +163,23 @@ def test_cluster_and_gcs(self, cluster, cluster_folder):
                 f"on the cluster {cluster.name}. For now please manually enable them directly on the cluster. "
                 f"See https://cloud.google.com/sdk/gcloud/reference/auth/login"
             )
+        
+    @pytest.mark.level("local") # TODO: Needs S3 and GCP credentials
+    def test_folder_and_folder(self, folder):
+        # from one cloud folder to another cloud folder
```
carolineechen, line 169: need to call `.to()` in order for this test to do anything. e.g. `new_folder = folder.to(system=folder.system)`

tests/test_resources/test_modules/test_folder.py
```diff
@@ -182,49 +163,23 @@ def test_cluster_and_gcs(self, cluster, cluster_folder):
...
 
-        s3_folder_to_gcs.rm()
-
-    @pytest.mark.level(
-        "minimal"
-    )  # TODO: fix this test (Needs google-cloud-storage package)
+    @pytest.mark.level("local") # TODO: Fix this test 
     @pytest.mark.skip("Doesn't work properly as only full-bucket copy is supported")
-    def test_gcs_and_s3(self, local_folder, gcs_folder):
+    def test_cloudA_and_cloudB(self, folder):
```
carolineechen, line 176: similar, `.to()` call needs to be specified

tests/test_resources/test_modules/test_folder.py
carolineechen: so if you look at the original `test_s3_and_s3` test (pasted below), the point is to send it to a new folder but on the same system, and test that the contents remain the same. this is different than just testing that the contents exist in the original fixture (which exist on s3)

that being said, I think we can just get rid of this test. `test_cloudA_and_cloudB` should be parameterized to handle the product of `s3_folder, gcs_folder, azure...` to `s3, gcs, etc` and therefore include both `s3_and_s3` and `gcs_and_gcs` options

```
def test_s3_and_s3(local_folder, s3_folder):
    # from one s3 folder to another s3 folder
    new_s3_folder = s3_folder.to(system="s3")
    assert "sample_file_0.txt" in new_s3_folder.ls(full_paths=False)

    new_s3_folder.rm()
```

tests/fixtures/folder_fixtures.py
```diff
@@ -26,6 +26,10 @@ def local_folder(tmp_path):
 def cluster_folder(ondemand_cpu_cluster, local_folder):
     return local_folder.to(system=ondemand_cpu_cluster)
 
+@pytest.fixture
+def local_folder_docker(docker_cluster_pk_ssh, local_folder):
+    return local_folder.to(system=docker_cluster_pk_ssh)
```
carolineechen, line 31: I think this should be `docker_cluster_pk_ssh_no_auth` for the most basic docker fixture. Looking at the fixtures used in the other tests, `docker_cluster_pk_ssh` is just the base docker cluster fixture used to define the other fixtures, but isn't actually explicitly used in testing

tests/test_resources/test_modules/test_folder.py
```diff
@@ -92,52 +105,20 @@ def test_cluster_tos(self, cluster, tmp_path):
...
-        local_from_s3 = s3_folder.to("here", path=tmp_path)
-        assert "sample_file_0.txt" in local_from_s3.ls(full_paths=False)
-        assert local_from_s3._fs_str == "file"
-
-        s3_folder.rm()
-
-    @pytest.mark.level(
-        "local"
-    )  # TODO: fix this test (just needs google-cloud-storage package)
+    @pytest.mark.level("minimal")  # TODO: fix this test
```
carolineechen, line 108: let's keep this as "local"

tests/test_resources/test_modules/test_folder.py
```diff
@@ -92,52 +105,20 @@ def test_cluster_tos(self, cluster, tmp_path):
...
+        # Folder to local
+        local_from_folder = folder.to("here", path=tmp_path)
+        assert "sample_file_0.txt" in local_from_folder.ls(full_paths=False)
+        assert folder._fs_str == "file" 
 
-    @pytest.mark.level("local")  # TODO: fix this test
+    
+    @pytest.mark.level("minimal")  # TODO: fix this test
     @pytest.mark.skip("[WIP] Fix this test")
     def test_cluster_and_s3(self, cluster, cluster_folder):
```
carolineechen, line 123: this can be removed and included as part of `test_cloudA_and_cloudB`

tests/test_resources/test_modules/test_folder.py
```diff
@@ -182,49 +163,35 @@ def test_cluster_and_gcs(self, cluster, cluster_folder):
...
+            assert folder._fs_str == "gs"
+        elif folder._fs_str == "gs": # test gcs and s3
+            folder = folder.to(system="s3")
+            assert folder._fs_str == "s3"
+        elif folder._fs_str == "ssh": # test ssh and s3
+            folder = folder.to(system="s3")
+            assert folder._fs_str == "s3"
+        elif folder._fs_str == "file": # test local and s3
+            folder = folder.to(system="s3")
+            assert folder._fs_str == "s3"
```
carolineechen, line 190: for each input folder, we should be testing sending it to all possible system options. e.g s3 -> s3, gs, azure, cluster, local to make sure that we cover all cases.

think about how you would go about parameterizing the dest system in a way that when this test is run on "local", it only runs the cluster to cluster + here case; and on "minimal" level, it runs all combinations.

tests/test_resources/test_modules/test_folder.py
```diff
@@ -183,49 +147,25 @@ def test_cluster_and_gcs(self, cluster, cluster_folder):
...
-
-    @pytest.mark.level("minimal")  # TODO: fix this test
-    @pytest.mark.skip("Doesn't work properly as only full-bucket copy is supported")
-    def test_s3_and_gcs(self, s3_folder):
-        # *** NOTE: transfers between providers are only supported at the bucket level at the moment (not directory) ***
-
-        s3_folder_to_gcs = s3_folder.to(system="gs")
-        assert "sample_file_0.txt" in s3_folder_to_gcs.ls(full_paths=False)
-        assert s3_folder_to_gcs._fs_str == "gs"
+        new_folder = folder.to(system=folder_mappings[dest]["system"])
```
carolineechen, line 161: I think this could just be

```
if instance(dest.system, cluster):
    expected_fs_str = "file" if dest.system == self.system else "ssh"
else:
    expected_fs_str = dest.system

new_folder = folder.to(system=dest.system)
```

tests/test_resources/test_modules/test_folder.py
```diff
@@ -183,49 +147,25 @@ def test_cluster_and_gcs(self, cluster, cluster_folder):
                 f"See https://cloud.google.com/sdk/gcloud/reference/auth/login"
             )
 
-    @pytest.mark.level("minimal")  # Needs S3 credentials
-    def test_s3_and_s3(self, local_folder, s3_folder):
-        # from one s3 folder to another s3 folder
-        new_s3_folder = s3_folder.to(system="s3")
-        assert "sample_file_0.txt" in new_s3_folder.ls(full_paths=False)
+    @pytest.mark.level("unit")
+    def test_send_folder_to_dest(self, folder, dest, cluster, tmp_path):
```
carolineechen, line 151: can remove cluster and tmp_path

tests/test_resources/test_modules/test_folder.py
carolineechen, line 161: this can probably be simplified to something like:

```
if instance(dest.system, cluster):
    expected_fs_str = "file" if dest.system == self.system else "ssh"
else:
    expected_fs_str = dest.system

new_folder = folder.to(system=dest.system)
assert new_folder._fs_str == expected_fs_str
assert ...
```

tests/test_resources/test_modules/test_folder.py
```diff
@@ -17,6 +17,30 @@ def fs_str_rh_fn(folder):
...
 
 class TestFolder(tests.test_resources.test_resource.TestResource):
+
+    MAP_FIXTURES = {"resource": "folder"}
+
+    UNIT = {
+        "folder": ["local_folder", "local_folder_docker"],
+        "dest": ["local_folder", "local_folder_docker"],
+        }
+    LOCAL = {
```
carolineechen, line 27: cluster_folder (running on a non-docker container) shouldnt be in local

tests/test_resources/test_resource.py
```diff
@@ -138,7 +138,8 @@ def test_sharing(
         # it compatible with ondemand_cluster requires changes
         # that break CI.
         # TODO: Remove this by doing some CI-specific logic.
-        if saved_resource.__class__.__name__ == "OnDemandCluster":
+        config = saved_resource.config_for_rns
+        if saved_resource.__class__.__name__ == "OnDemandCluster" or config["name"] == "docker_cluster_pk_ssh":
```
carolineechen, line 142: why is `config["name"] == "docker_cluster_pk_ssh"` being added here?

tests/fixtures/folder_fixtures.py
```diff
@@ -22,6 +28,22 @@ def local_folder(tmp_path):
...
+@pytest.fixture
+def local_folder_docker(docker_cluster_pk_ssh_no_auth):
+    args = {
+        "name": "test_folder",
+        "system": docker_cluster_pk_ssh_no_auth,
+        "path": "rh-folder",
+    }
+
+    local_folder_docker = rh.folder(**args)
+    init_args[id(local_folder_docker)] = args
```
carolineechen, line 40: can you do the same for all the remaining folder fixtures? you should be able to run minimal and have them all test the base resource tests locally.

tests/fixtures/folder_fixtures.py
```diff
@@ -23,22 +29,65 @@ def local_folder(tmp_path):
...
+    local_folder_docker.put(
+        {f"sample_file_{i}.txt": f"file{i}".encode() for i in range(3)}
+    )
+    return local_folder_docker
+
+
+@pytest.fixture
+def cluster_folder(ondemand_cpu_cluster):
+    args = {
+        "name": "test_folder",
```
carolineechen, line 50: can you name the folders differently to avoid conflicts?

PR #410, opened 2024-01-29
Fix `put_resource` to not store resource in itself if it is an Env.

rohinb2:
If we put an env, we ended up storing the Env resource within the Env servlet
for that env, which doesn't really make sense. Changed put_resource to be
similar to put, where you can choose where the resource is stored, and if it
happens to be an env then we create that servlet, however don't necessarily
store it.

You can still simulate that behavior by adding the logic back on the client
side (in cluster.put_resource) that sets env_name to the env if it is one.

### DIFF
--- a/runhouse/servers/obj_store.py
+++ b/runhouse/servers/obj_store.py
@@ -688,30 +688,32 @@ class ObjStore:
             )
             return self.put_resource_local(resource_config, state, dryrun)
 
-        servlet = ObjStore.get_env_servlet(env_name)
-
         # Normally, serialization and deserialization happens within the servlet
         # However, if we're putting an env, we need to deserialize it here and
         # actually create the corresponding env servlet.
-        if not servlet:
-            servlet = ObjStore.get_env_servlet(env_name, create=True)
-            resource_config, _, _ = deserialize_data(serialized_data, serialization)
-            if resource_config["resource_type"] == "env":
-                runtime_env = (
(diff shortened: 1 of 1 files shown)


rohinb2:
* `main`
  * **PR #410** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/410?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/410?utm_source=stack-comment).

carolineechen (commented):
overall changes make sense but there's some a test failure caused by the change. specifically, for setting env vars in an env we use `system.call(env_name, "_set_env_vars", env_vars)` which calls `call_module_method` and sets the env to be `env = message.env or obj_store.get_env_servlet_name_for_key(module)`. with this PR, this now returns `base` rather than `env_name` and sets the env vars in the base servlet rather than the actual env it was meant for, so env var syncing isn't done properly anymore.

will look into a solution for this.

dongreenberg:
I think this is fair, it just introduces more control, but I'd add back the client side logic to put the env into its own servlet when we call env.to. That should address the issue Caroline raised too.

rohinb2:
kk, fine with that. @carolineechen in the deletion diff, can we still switch it to explicit deletion and test both?

PR #416, opened 2024-01-31
runhouse status

BelSasha:
initial runhouse status

Merge branch 'main' into sb/runhouse-status

### DIFF
--- a/runhouse/main.py
+++ b/runhouse/main.py
@@ -6,19 +6,25 @@ import webbrowser
 from pathlib import Path
 from typing import List, Optional
 
+import requests
+
 import typer
 from rich.console import Console
+
+import runhouse as rh
 
 import runhouse.rns.login
 
 from runhouse import __version__, cluster, configs
 from runhouse.constants import (
+    RAY_KILL_CMD,
     SERVER_LOGFILE,
     SERVER_START_CMD,
     SERVER_STOP_CMD,
     START_NOHUP_CMD,
     START_SCREEN_CMD,
 )
+
 
 # create an explicit Typer application
 app = typer.Typer(add_completion=False)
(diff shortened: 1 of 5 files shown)


BelSasha:
* `main`
  * **PR #416** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/416?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/416?utm_source=stack-comment).

runhouse/main.py
```diff
@@ -98,6 +101,92 @@ def ssh(cluster_name: str, up: bool = typer.Option(False, help="Start the cluste
     c.ssh()
 
 
+@app.command()
+def status(cluster_name: str = typer.Argument(None, help="Name of cluster to check")):
+    """Get the config info about your runhouse cluster(s), as well as the environments setup of the cluster(s)"""
+
+    cluster_or_local = rh.here
+
+    if cluster_or_local == "file":
```
jlewitt1, line 110: I think we can simplify this block - wdyt of something like this: 

```
        if cluster_name is None:
            raise ValueError("Missing argument cluster_name")

        try:
            current_cluster = rh.cluster(name=cluster_name)
            if not current_cluster.is_up():
                console.print(f"Cluster {current_cluster} is not up.")
                return

            current_cluster.connect_server_client()
            config = current_cluster.client.request(
                endpoint="/status", req_type="get"
            )
        except ValueError as e:
            console.print(e)
            return
```

runhouse/main.py
```diff
@@ -98,6 +101,92 @@ def ssh(cluster_name: str, up: bool = typer.Option(False, help="Start the cluste
...
+                        "up."
+                    )
+                return
+    else:
+        from runhouse.globals import obj_store
+
+        config = obj_store.get_status()
+
+    envs = config["envs"]
+    config.pop("envs")
```
jlewitt1, line 137:
```suggestion
    env = config.pop("envs", [])
```

tests/test_resources/test_clusters/test_cluster.py
```diff
@@ -184,3 +186,15 @@ def test_rh_here_objects(self, cluster):
         save_test_table_remote()
         assert "test_table" in cluster.keys()
         assert isinstance(cluster.get("test_table"), rh.Table)
+
+    @pytest.mark.level("local")
+    def test_rh_status(self, cluster):
```
jlewitt1, line 191: This tests the local use case, but prob also worth adding a test where we run the command on the cluster: 
For example: 
`cluster.run(["runhouse status"])`

runhouse/main.py
```diff
@@ -98,6 +101,92 @@ def ssh(cluster_name: str, up: bool = typer.Option(False, help="Start the cluste
     c.ssh()
 
 
+@app.command()
+def status(cluster_name: str = typer.Argument(None, help="Name of cluster to check")):
```
jlewitt1, line 105: can we also add this command to the docs? in `cli.rst`

tests/test_resources/test_clusters/test_on_demand_cluster.py
```diff
@@ -20,3 +24,18 @@ class TestOnDemandCluster(tests.test_resources.test_clusters.test_cluster.TestCl
             "multinode_cpu_cluster",
         ]
     }
+
+    @pytest.mark.level("minimal")
+    def test_rh_status_errors(self, cluster):
```
jlewitt1, line 29: do we need to use a cluster here? instead of tearing down and upping again? If we are just checking the status for a cluster that doesn't exist

runhouse/main.py
```diff
@@ -98,6 +101,101 @@ def ssh(cluster_name: str, up: bool = typer.Option(False, help="Start the cluste
     c.ssh()
 
 
+def print_status(config):
```
jlewitt1, line 104: make private method?

runhouse/servers/http/http_server.py
```diff
@@ -707,6 +707,13 @@ async def call(
 
         return JSONResponse(content=resp)
 
+    @staticmethod
+    @app.get("/status")
+    @validate_cluster_access
+    def get_status(request: Request):
```
jlewitt1, line 713: super clean!

tests/test_resources/test_clusters/test_on_demand_cluster.py
jlewitt1, line 29: This does doesn't need to be limited to an ondemand cluster, thoughts on making it a local test and moving into "test_cluster"? is there any reason why using a local docker cluster wouldn't be enough here?

runhouse/main.py
```diff
@@ -98,6 +102,116 @@ def ssh(cluster_name: str, up: bool = typer.Option(False, help="Start the cluste
...
+    """
+    envs = config["envs"]
+    config.pop("envs", [])
+
+    # print headlines
+    daemon_headline_txt = (
+        "\N{smiling face with horns} Runhouse Daemon is running \N{Runner}"
+    )
+    console.print(daemon_headline_txt, style="bold royal_blue1")
+    console.print(f"{config['name']}", style="bold underline")
```
jlewitt1, line 119: not sure if this is possible, but is there a way to hyperlink the cluster rns address that we display? 

for example, if the cluster name is: /jlewitt1/rh-cpu-https its corresponding Den URL would be:
https://www.run.house/resources/jlewitt1:rh-cpu-https

runhouse/main.py
```diff
@@ -98,6 +102,116 @@ def ssh(cluster_name: str, up: bool = typer.Option(False, help="Start the cluste
...
+            try:
+                current_cluster = rh.cluster(name=cluster_name)
+                if not current_cluster.is_up():
+                    console.print(f"A cluster called {cluster_name} is not up")
+                    return
+                current_cluster.connect_server_client()
+                config = current_cluster.client.request(
+                    endpoint="/status", req_type="get"
+                )
+            except Exception:
```
jlewitt1, line 200: I think we should actually break these out into two separate exceptions to be more descriptive: 
`ValueError`: Cluster not found
`ConnectionError`: Cluster exists but daemon not running properly

runhouse/main.py
```diff
@@ -98,6 +102,116 @@ def ssh(cluster_name: str, up: bool = typer.Option(False, help="Start the cluste
...
+                console.print(
+                    "\N{smiling face with horns} Runhouse Daemon is not running... \N{No Entry}\N{Runner}"
```
jlewitt1, line 202:
```suggestion
                    "\N{smiling face with horns} Runhouse Daemon is not running... \N{No Entry}\ N{Runner}"
```

runhouse/main.py
```diff
@@ -98,6 +102,116 @@ def ssh(cluster_name: str, up: bool = typer.Option(False, help="Start the cluste
...
+            for resource in resources_in_env:
+                resource_name = resource["name"]
+                resource_type = resource["resource_type"]
+                console.print(f"\u2022{resource_name} ({resource_type})")
+
+    return config
+
+
+@app.command()
+def status(cluster_name: str = typer.Argument(None, help="Name of cluster to check")):
```
jlewitt1, line 181:
```suggestion
def status(cluster_name: str = typer.Argument(None, help="Name of cluster to check. If not specified will check the local cluster.")):
```

tests/test_resources/test_clusters/test_cluster.py
```diff
@@ -184,3 +185,48 @@ def test_rh_here_objects(self, cluster):
...
+        assert f"resource_type: {cluster.RESOURCE_TYPE.lower()}" in res
+        assert f"ips: {str(cluster.ips)}" in res
+        assert "Serving 🍦 :" in res
+
+        cluster_keys = cluster.keys()
+        for key in cluster_keys:
+            assert key in res
+
+    @pytest.mark.level("local")
+    def test_rh_status_errors(self, docker_cluster_pk_ssh_no_auth):
```
jlewitt1, line 222: any reason to use a specific fixture here and not just "cluster" which will be parametrized across multiple types?

runhouse/main.py
```diff
@@ -98,6 +104,130 @@ def ssh(cluster_name: str, up: bool = typer.Option(False, help="Start the cluste
...
+                current_cluster = rh.cluster(name=cluster_name)
+                if not current_cluster.is_up():
+                    console.print(f"A cluster called {cluster_name} is not up")
+                    return
+                current_cluster.connect_server_client()
+                config = current_cluster.client.request(
+                    endpoint="/status", req_type="get"
+                )
+            except ValueError as val_e:
+                if str(val_e) == f"Resource {cluster_name} not found.":
```
jlewitt1, line 208: Is there more than one type of ValueError that could be raised here?

PR #421, opened 2024-02-02
Update span dimensions to align with Den 

DenisYay:


DenisYay:
Screenshot 2024-02-02 at 1 53 42 PM

DenisYay:
Confirmed that we see service.namespace, service.version, service_name and deployment_environment

PR #424, opened 2024-02-05
tests requiremrnts.txt

BelSasha:
Installing all requirements for tests (Local and CI) using requirements.txt


tests/test_requirements/aws_test_requirements.txt
```diff
@@ -0,0 +1,8 @@
+runhouse[aws]
+pytest
+tqdm
+datasets
+dask
```
carolineechen, line 5: since there is lot of duplicate requirements between base testing reqs and all these specific testing ones, think I would be in favor of having this directory something like `/additional_requirements/`, and removing the overlaps with `tests/requirements.txt`.

this way, it's easier to read/keep track of, there's no need to propagate updates in base reqs.txt into each and every sub type, and adding new reqs files for future features is shorter too.

tests/requirements.txt
```diff
@@ -1,3 +1,5 @@
 pytest
 httpx
 pytest_asyncio
+datasets
+dask
```
carolineechen, line 5: think I would include `tqdm` here too

.github/workflows/aws_tests.yaml
```diff
@@ -22,7 +22,8 @@ jobs:
 
       - name: Install python packages & dependencies
         run: |
-          pip install runhouse[aws] pytest tqdm datasets dask
+          pip install tests/test_requirements/additional_requirements.txt
```
carolineechen, line 25: I don't think we should be installing additional_requirements.txt for these workflows bc several of them aren't being used and are quite large installations (e.g. torch), let's just try including the minimal necessary for the ones here, I think requirements.txt+aws_test_requirements should be sufficient?

tests/test_requirements/additional_requirements.txt
```diff
@@ -0,0 +1,5 @@
+torch
```
carolineechen, line 1: thoughts on renaming this to tutorials_requirements.txt (or just removing it for now since we don't use tutorials tests right now), and taking out runhouse[sagemaker] + runhouse[aws]? for those we can just directly specify in the command rather than including runhouse libraries in reqs.txt

tests/test_requirements/aws_test_requirements.txt
```diff
@@ -0,0 +1,4 @@
+runhouse[aws]
```
carolineechen, line 1: similarly, can take out runhouse[aws] and explicitly call pip install runhouse[aws] in the workflow.

.github/workflows/aws_tests.yaml
```diff
@@ -22,7 +22,9 @@ jobs:
 
       - name: Install python packages & dependencies
         run: |
-          pip install runhouse[aws] pytest tqdm datasets dask
+          pip install runhouse[aws]
+          pip install -r tests/requirements.txt
+          pip install -r tests/test_requirements/aws_test_requirements.txt
```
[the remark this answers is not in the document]
BelSasha: The aws_test_reqs.txt installs additional packages that are required when executing tests related to AWS. Other tests (such as test_cluster.yaml), require only runhouse[aws], and do not require the packages listed in the aws_test_reqs.txt. To avoid unnecessary installations, I separated the runhouse[aws] install from the aws reqs file.

.github/workflows/cluster_tests.yaml
```diff
@@ -29,7 +29,8 @@ jobs:
 
       - name: Install python packages & dependencies
         run: |
-          pip install runhouse[aws] pytest tqdm datasets
+          runhouse[aws]
```
carolineechen, line 32: pip install runhouse[aws]?

tests/test_requirements/google_tests_requirements.txt
```diff
@@ -0,0 +1,4 @@
+google-api-python-client
+google-cloud-storage
+gcsfs
+fsspec<=2023.5.0
```
carolineechen, line 4: fsspec is already in base reqs.txt, and I think we can remove this altogether since it's a requirement of runhouse already

PR #426, opened 2024-02-05
Introduce cluster.status()

dongreenberg:
Rather than call `runhouse status <cluster_name>` in tests and load from RNS

### DIFF
--- a/runhouse/servers/obj_store.py
+++ b/runhouse/servers/obj_store.py
@@ -882,13 +882,20 @@ class ObjStore:
             # The objects in env can be of any type, and not only runhouse resources,
             # therefore we need to distinguish them when creating the list of the resources in each env.
             for r in resources_in_env:
+                cls = type(r)
+                py_module = cls.__module__
+                cls_name = (
+                    cls.__qualname__
+                    if py_module == "builtins"
+                    else (py_module + "." + cls.__qualname__)
+                )
                 if isinstance(r, runhouse.Resource):
                     resources_in_env_modified.append(
-                        {"name": r.name, "resource_type": r.RESOURCE_TYPE.capitalize()}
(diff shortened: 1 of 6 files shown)


dongreenberg:
* `main`
  * **PR #426** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/426?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/426?utm_source=stack-comment).

runhouse/servers/obj_store.py
```diff
@@ -882,13 +882,20 @@ def get_status(self):
...
+                )
                 if isinstance(r, runhouse.Resource):
                     resources_in_env_modified.append(
-                        {"name": r.name, "resource_type": r.RESOURCE_TYPE.capitalize()}
+                        {"name": r.name, "resource_type": cls_name}
                     )
                 else:
                     resources_in_env_modified.append(
-                        {"name": r, "resource_type": type(r).__name__.capitalize()}
+                        {"name": r, "resource_type": cls_name}
```
BelSasha, line 898: since the cls_name definition is unified for both Runhouse resources and other Python classes,  I think that the if-else block is no longer relevant. 

should be just `{"name": r, "resource_type": cls_name}` for each r.

dongreenberg: Ya agreed, I'm just changing the code minimally to unblock CI. There is actually a bigger issue here that we will need to fix soon having to do with the fact that we can't actually load the objects from all the different envs here to be able to call these methods on them, because this process may not have the dependencies to deserialize them. We'll fix that in a separate PR and clean this up more there.

PR #434, opened 2024-02-07
Only set rns_address upon save

dongreenberg:
Until now we've been setting the _rns_folder of a resource implicitly to
rns_client.current_folder, whether the resource has been saved or not,
so the address is really meaningless. This is a relic from a time when
we tried to save resources constantly as they were created and changed.
Now, it would be much easier to have a reasonable assumption of whether
a resource has been saved and therefore whether its rns_address is
meaningful if passed to another environment. For example, when saving
a resource config as a sub-config of another resource, we need to know
whether we can just save a string because the resource has been saved
or if we need to return the full sub-resource config. This bug in particular
was breaking some things.

This change changes the behavior of `rns_address` to return None if there
is no _rns_folder or no _name, and saves the _rns_folder to the current_folder
only upon save. If a user gives a resource a full rns_address that's their own
business, we will not check if the address is actually pointing to something
when we return the property.

### DIFF
--- a/runhouse/resources/module.py
+++ b/runhouse/resources/module.py
@@ -919,7 +919,7 @@ class Module(Resource):
 
     def rename(self, name: str):
         """Rename the module."""
-        if self.name == name:
+        if self.name == name or self.rns_address == name:
             return
         old_name = self.name
         self.name = name  # Goes through Resource setter to parse name properly (e.g. if rns path)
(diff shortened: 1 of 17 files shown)


dongreenberg:
* `main`
  * **PR #434** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/434?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/434?utm_source=stack-comment).

runhouse/resources/module.py
```diff
@@ -928,9 +928,13 @@ def rename(self, name: str):
             and isinstance(self.system, Cluster)
             and self.system.on_this_cluster()
         ):
-            obj_store.rename(old_key=old_name, new_key=self.rns_address)
+            # We rename the object with the full rns_address here because if it's a resource, we want the
+            # object stored in the obj store to have the updated rns_address, not just the updated key.
+            obj_store.rename(old_key=old_name, new_key=self.rns_address or self.name)
```
rohinb2, line 933: do we need this anymore after my fix?

PR #445, opened 2024-02-08
verify self signed certs in http client

jlewitt1:
Allow for self-signed cert verification for requests made to a cluster started with https

For example: 
```curl --cacert '/Users/josh.l/.rh/certs/rh_server.crt' https://<cluster-ip>/check```

### CODE REVIEW

**Comment by jlewitt1** [comment]
* `main`
  * **PR #445** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/445?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/445?utm_source=stack-comment).

File: runhouse/resources/hardware/cluster.py [hunk 1]
```diff
@@ -642,18 +642,13 @@ def restart_server(
...
-                # If no cert and keyfile already exist, generate them client side and copy them onto the cluster
-                if (
-                    not Path(self.cert_config.cert_path).exists()
-                    or not Path(self.cert_config.key_path).exists()
-                ):
-                    self.cert_config.generate_certs(address=self.address)
-
+                # Need to include the IP address in the cert's SANs, so rebuild on restart to ensure the
+                # correct IP is included. Cert subject name needs to match the target host name (in this case IP)
+                self.cert_config.generate_certs(address=self.address)
```
dongreenberg, line 647: Not sure I follow, why do we need this?

jlewitt1: the subject name on the cert needs to match the target, otherwise verifcation will fail. So if we have a domain we can re-use the same cert across multiple servers, but if we don't the IP address will need to be updated. Do you see an alternative?

dongreenberg: Not sure I follow - with the flow that's in there now I was able to verify a cluster with an existing cert in my filesystem, no customization for IP. What was the error you saw?

PR #470, opened 2024-02-13
Fix `name` not being updated in cluster servlet and breaking `on_this_cluster`.

rohinb2:
In [this PR](https://app.graphite.dev/github/pr/run-house/runhouse/434/Only-set-rns_address-upon-save), @dongreenberg sets it up so RNS address is only set on save.

However, we set the cluster name within the cluster servlet remotely once, and then if the RNS address was updated, on_this_cluster would fail since it checked _current_cluster("name") == self.rns_address. So, when RNS address is updated, I now also send a request to the daemon and set the cluster name in the config

### DIFF
--- a/runhouse/resources/hardware/cluster.py
+++ b/runhouse/resources/hardware/cluster.py
@@ -127,6 +127,30 @@ class Cluster(Resource):
             node=node or self.address,
         )
 
+    def save(
+        self,
+        name: str = None,
+        overwrite: bool = True,
+    ):
+        # return super().save(name=name, overwrite=overwrite)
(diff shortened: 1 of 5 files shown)


rohinb2:
* `main`
  * **PR #470** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/470?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/470?utm_source=stack-comment).

PR #471, opened 2024-02-13
Update cluster.save behavior

carolineechen:
`on_this_cluster` was still being computed incorrectly sometimes, when `self.client` was None and therefore `cluster_name` was not actually being updated during `cluster.save()`. in this change, additionally when assigning the client, check if the cluster was saved (has an `rns_address`), and if so, have a call to the update there.


carolineechen:
* `main`
  * **PR #471** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/471?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/471?utm_source=stack-comment).

PR #479, opened 2024-02-14
remove live state from cluster

BelSasha:


tests/test_resources/test_modules/test_module.py
```diff
@@ -313,7 +313,8 @@ def test_module_from_subclass(self, cluster, env):
         resolved_obj = helper(remote_instance.resolve())
         assert resolved_obj.__class__.__name__ == "SlowPandas"
         assert resolved_obj.size == 20  # resolved_obj.remote.size causing an error
-        assert resolved_obj.config_for_rns == remote_instance.config_for_rns
+        assert resolved_obj.system == remote_instance.system
+        # assert resolved_obj.config_for_rns == remote_instance.config_for_rns
```
jlewitt1, line 317: how come we're removing this? any reason why the configs would be different?

runhouse/resources/resource.py
```diff
@@ -253,6 +253,64 @@ def str_dict_or_resource_to_str(val):
                 return None
         return config
 
+    @classmethod
+    def _update_system_in_config(cls, config):
```
dongreenberg, line 257: What you're doing here and in `update_creds_in_config` is already supported via Resource._check_for_child_configs. Take a look at how Module ensures that the system and env are properly reloaded from a config (or string) inside Module._check_for_child_configs. We shouldn't be doing this within the resource class because it is not presumed that a resource has a system or creds as a sub-resource. Every subclass can overload to specify what sub-resources they need to reconstruct from the config.

runhouse/resources/hardware/cluster_factory.py
```diff
@@ -103,10 +102,14 @@ def cluster(
         except ValueError:
             pass
 
+    from runhouse.resources.secrets.provider_secrets.ssh_secret import SSHSecret
+
+    ssh_creds_secret = SSHSecret.setup_ssh_creds(ssh_creds, name)
```
carolineechen, line 107: note that `ssh_creds` is optional and might be `None`

runhouse/resources/secrets/provider_secrets/ssh_secret.py
```diff
@@ -134,3 +169,86 @@ def _file_to(
             system.call(key, "_write_to_file", path=path, values=values)
             remote_priv_file = file(path=path, system=system)
         return remote_priv_file
+
+    @classmethod
+    def setup_ssh_creds(cls, ssh_creds: Union[dict, str], resource_name: str):
+        """
+        this method creates an SSHSecret instance based o n the passed values. If the passed values are paths to private
+         and/or public keys, this method extracts the content of the files saved in those files, in order for them to
+         be saved in den. (Currently if we just pass a path/to/ssh/key to SSHSecret constructor, the content of the file
```
carolineechen, line 178: I believe that in main currently, the contents are already saved to vault (but not to rns which is not secure).

e.g. the values for the following secret show up in vault
```
secret = rh.provider_secret(provider="ssh", name="sky", path="~/.ssh/sky-key")
secret.save()
````

Comment (carolineechen) [reply runhouse/resources/secrets/provider_secrets/ssh_secret.py:178]: also currently `SSHSecret` is meant to represent specifically an ssh key-pair with the values dict keys corresponding to that, rather than a generic ssh cluster creds secret. adding/changing the values keys will cause the built-in  functions like `.write()` to file to break.

We should eventually build out a `SSH/ClusterCredsSecret` for this case. Another option in the meantime is to convert this to return a generic secret instead (move this function to a general utils file, and can utilize SSHSecret functions to help with loading values etc).

Comment (BelSasha) [reply runhouse/resources/secrets/provider_secrets/ssh_secret.py:178]: Agreed. Made the ssh_cluter creds to be generic secret, and moved this function to secrets/utils.py for now.

File: runhouse/resources/hardware/cluster_factory.py [hunk 5]
```diff
@@ -38,8 +38,8 @@ def cluster(
         name (str): Name for the cluster, to re-use later on.
         host (str or List[str], optional): Hostname (e.g. domain or name in .ssh/config), IP address, or list of IP
             addresses for the cluster (the first of which is the head node).
-        ssh_creds (dict, optional): Dictionary mapping SSH credentials.
-            Example: ``ssh_creds={'ssh_user': '...', 'ssh_private_key':'<path_to_key>'}``
+        ssh_creds (dict, optional): SSH credentials, passed as dictionary. Example:
```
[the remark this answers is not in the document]
BelSasha: The original thought behind allowing ssh_creds to receive `Dict` only was to left the API as is. The users are used to pass ssh_creds as dictionaries. Moreover, if the cluster is being created for the first time (and not from_name), we are creating the secret in the cluster factory. 

I think that maybe we should not pass Secrets to the factory method just yet, since we are planning to have CredsSecret class, so probably it will be better to discuss this when we'll develop the new secret. 

@carolineechen what do you think?

runhouse/main.py
```diff
@@ -96,14 +96,24 @@ def notebook(
...
+
+    try:
+        c = cluster(name=cluster_name)
+    except ValueError:
+        logger.error(
+            f"Could not load cluster called {cluster_name} from Den. Please save it to Den, and rerun."
         )
         raise typer.Exit(1)
+
+    is_shared = f"{c._creds.name}/" in c.ssh_creds.get("ssh_private_key", "")
```
jlewitt1, line 108: not sure we should be upping the cluster as part of this command - if it's down then maybe we just let the user know?

runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -630,14 +519,15 @@ def ssh(self, node: str = None):
...
             subprocess.run(cmd, shell=True, check=True)
 
-        elif node is None:
-            # SSH onto head node - can provide the name as specified in the local ~/.ssh/config
-            subprocess.run(["ssh", f"{self.name}"])
         else:
             # If SSHing onto a specific node, which requires the default sky public key for verification
-            sky_key = Path(self.DEFAULT_KEYFILE).expanduser()
+            ssh_user = self.ssh_creds.get("ssh_user")
+            node = self.ips[0] or self.address
```
jlewitt1, line 525:
```suggestion
            node = node or self.address
```
address is extracted from IPs - if not specified use the head node IP

runhouse/resources/secrets/provider_secrets/ssh_secret.py
```diff
@@ -88,11 +90,14 @@ def _write_to_file(
...
-        new_secret._add_to_rh_config(path)
+        try:
+            new_secret._add_to_rh_config(val=path)
+        except TypeError:
+            pass
 
         return new_secret
 
-    def _from_path(self, path: Union[str, File]):
+    def _from_path(self, path: Union[str, File], public_path=None):
```
jlewitt1, line 100: not sure i follow, why do we need this new argument?

runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -79,7 +78,8 @@ def __init__(
             dryrun=dryrun,
             **kwargs,
         )
-
+        self._creds = kwargs.get("creds", None)
+        self.address = kwargs.get("head_ip", None)
```
jlewitt1, line 82: address is a property that gets set in the parent Cluster class, how come we're overriding that here?

jlewitt1:
* **#544** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/544?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#479** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/479?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

runhouse/resources/hardware/cluster.py
```diff
@@ -185,7 +183,17 @@ def config_for_rns(self):
             ],
         )
         if self.is_up():
-            config["ssh_creds"] = self.ssh_creds
+            creds = (
```
rohinb2, line 186: just for clarity, can we call this `creds_string` or something? was confused when reading

runhouse/resources/hardware/cluster.py
```diff
@@ -54,7 +55,7 @@ def __init__(
         # Name will almost always be provided unless a "local" cluster is created
         name: Optional[str] = None,
         ips: List[str] = None,
-        ssh_creds: Dict = None,
+        creds: "Secret" = None,
```
rohinb2, line 58: seems like this could also be a `str`? I see that in `def ssh_creds` you check if it could be a string and load secret if so

rohinb2: or is that only if it is constructed via `from_config`? fine if so

runhouse/resources/hardware/cluster.py
```diff
@@ -1385,3 +1401,55 @@ def set_connection_defaults(self, **kwargs):
...
+        notify_users: bool = True,
+        headers: Optional[Dict] = None,
+        # Deprecated
+        access_type: Union[ResourceAccess, str] = None,
+    ) -> Tuple[Dict[str, ResourceAccess], Dict[str, ResourceAccess]]:
+
+        # save cluster and creds if not saved
+        self.save()
+
+        # share creds
```
rohinb2, line 1419: can we add a log here to the user saying that the underlying creds are also being shared?

runhouse/resources/secrets/utils.py
```diff
@@ -109,3 +109,91 @@ def _check_file_for_mismatches(path, existing_vals, new_vals, overwrite):
         )
         return True
     return False
+
+
+def setup_cluster_creds(ssh_creds: Union[Dict, str], resource_name):
```
rohinb2, line 114: this looks great, well done

rohinb2 (approved):
awaiting CI passing

dongreenberg:
Confirmed that sharing across users works. The following was run in a fresh environment with a different account:
image

PR #516, opened 2024-02-26
add on-demand shared cluster fixture

jlewitt1:


dongreenberg:
It's pretty expensive to create a whole new cluster just to share it, is there a reason we can't just use the existing ondemand cpu cluster fixture?

PR #519, opened 2024-02-26
Change some logging to go to `sys.stdout`.

rohinb2:
When calling a method, we use the `StreamTee` in `provenance.py` to log to a file _and_ stdout at the same time, which is how we stream logs from `call`.

However, when we call `env.install`, we call things via subprocess, and these logs aren't propagated. Unfortunately there's no neat way to pipe this (I scoured the entirety of StackOverflow) so we just need to set up a pipe, stream the logs, and print things, and then they'll go to `sys.stdout` and be picked up by the remote log generator. 

While I was here, cleaned up a bit of the conda installation logic.

### DIFF
--- a/runhouse/resources/envs/conda_env.py
+++ b/runhouse/resources/envs/conda_env.py
@@ -8,9 +8,9 @@ from typing import Dict, List, Optional, Union
 from runhouse.globals import obj_store
 
 from runhouse.resources.packages import Package
+from runhouse.utils import install_conda
 
 from .env import Env
-from .utils import _install_conda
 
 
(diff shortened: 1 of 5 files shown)


rohinb2:
* **#519** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/519?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/519?utm_source=stack-comment).

PR #595, opened 2024-03-13
Catching module exception in load_config

carolineechen:
ModuleNotFound in module's load_config was previously ignored, for cases when user is loading a (shared) module that doesn't live on their local env. This however also ignores ModuleNotFound errors due to other factors such as missing dependencies, which should be caught.


carolineechen:
* **#595** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/595?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #606, opened 2024-03-15
Selective sky refresh when checking is_up() in internal calls

carolineechen:
ondemand cluster is_up() always refreshes, which is time consuming and can happen multiple times within the same calls (as part of check_server). In our workflow refresh is only useful if the cluster was autodowned or manually taken down, or if address is not set and we need to repopulate cluster stats. Add `refresh=None` or False option for `is_up()`. If refresh is None and called within autostop_mins, don't refresh when checking.

also, don't automatically up() a cluster if it's not up in check_server, leave the responsibility to the user to make sure the cluster is up when they run commands.

### DIFF
--- a/runhouse/resources/hardware/cluster.py
+++ b/runhouse/resources/hardware/cluster.py
@@ -212,7 +212,7 @@ class Cluster(Resource):
         the internal url (including the local connected port rather than the sever port). If cluster is not up,
         returns None.
         """
-        if not self.is_up():
+        if not self.is_up(refresh=None):
             return None
 
         if self.server_connection_type in [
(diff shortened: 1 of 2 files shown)


carolineechen:
* **#606** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/606?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

carolineechen:
#614 helps cut down on calls to sky refresh. will revisit if still seeing excessive refreshing (replacing w/ ping or the like), but the time-based approach in this PR is not as correct/reliable, especially with shared clusters

PR #721, opened 2024-04-11
Don't serialize tracebacks in OutputType.EXCEPTION Responses

dongreenberg:
Previously, we serialized both the exception and the traceback together, so if the deserialization failed (e.g. if there was a library incompatibility between the client and server which blocked unpickling) we couldn't see the traceback either, even though it's a string. It's important that we try to desrialize the exception so users have a chance to catch it properly locally, but if it's impossible to deserialize, we may as well just raise a runtime error and still print the traceback. If the response from the cluster is still in the old format, we'll still try to unpickle it for compatibility, but we can eventually deprecate that.

### DIFF
--- a/runhouse/servers/http/http_server.py
+++ b/runhouse/servers/http/http_server.py
@@ -271,12 +271,12 @@ class HTTPServer:
         except Exception as e:
             logger.exception(e)
             exception_data = {
-                "error": e,
+                "error": serialize_data(e, serialization),
                 "traceback": traceback.format_exc(),
             }
             return Response(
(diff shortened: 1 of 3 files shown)


dongreenberg:
* **#721** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/721?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @dongreenberg and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

PR #751, opened 2024-04-24
Don't replace `base_env` with `system.default_env`.

rohinb2:
There are several instances where, if the env happened to be `base_env`, we replace it with the cluster default env. The problem is we don't always replace this client side and it creates some inconsistencies. Wanted to discuss this logic a bit more

I've marked these with `TODO: (default env)`

cc: @carolineechen when you're back

### DIFF
--- a/runhouse/resources/envs/env.py
+++ b/runhouse/resources/envs/env.py
@@ -206,7 +206,7 @@ class Env(Resource):
         new_env.secrets = self._secrets_to(system)
 
         if isinstance(system, Cluster):
-            # TODO: Think about this. When sending an env that was unnamed to a cluster, should the remote env
+            # TODO: (default env). When sending an env that was unnamed to a cluster, should the remote env
             # be named the same as the local env? Or should it be named the same as the default env?
             # if new_env.name == Env.DEFAULT_NAME:
             #     new_env.name = system.default_env.name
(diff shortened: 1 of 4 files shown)


rohinb2:
* **#751** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/751?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

carolineechen (commented):
I think these todos generally cover the edge case of handling unnamed envs (base_env) being sent to the cluster, in general and also as part of resources like modules/functions/secrets. My general thoughts:

I think it makes more sense to have unnamed envs just default to the cluster's default_env, and have default_env be the catch-all env rather than a separate base_env which can be confusing. if an env is unnamed and sent to the cluster, we can assume the user probably don't need to access it later, and it's rather confusing if they have to use "base_env" to refer to it later when they didn't name it in the first place. 

With the cases of a user passing in an unnamed env to a module constructor or sending a secret, I can see why it could make sense to have it live in it's own separate base_env in some situations, but there's also an issue with having multiple unnamed base_envs for different modules and those getting jumbled up (I think put_resource handles installs fine but overwrites envs on the cluster). I think defaulting to default_env is fine here, but encourage users to use named envs if they want it to be in a more isolated, recoverable environment. 

Another thing, for put_resource for an env that exists on the cluster, we handle the new installs such that all the old/new reqs are present and code run fine, but the resource config is overwriten on the cluster. this might make sense for non base/default envs, but if sending an unnamed env w/ new reqs to the cluster, the existing env shouldn't be replaced. perhaps append the new reqs, or handle the installs but don't overwrite the existing env?

PR #784, opened 2024-05-07
Update den auth default to None in cluster factory

carolineechen:
default value `False` for den auth in cluster factory caused alt_options to be set, so calling `cluster("name")` would sometimes result in a config mismatch and fail to reconstruct the resource correctly. set it to none in factory and correspondingly update to False in cluster init instead.


carolineechen:
* **#784** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/784?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #806, opened 2024-05-20
polling den with cluster status data

BelSasha:


BelSasha:
* **#806** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/806?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

runhouse/servers/cluster_servlet.py
```diff
@@ -217,3 +234,82 @@ async def aclear_all_references_to_env_servlet_name(self, env_servlet_name: str)
...
+            logger.error(
+                "The relevant obj-store is not associated with the cluster servlet, therefore can't get cluster status."
+                " Setting the obj_store can be done using cluster_servlet.set_obj_store(associated_obj_store)."
+            )
+            raise ClusterServletError(
+                "Can't find the obj_store associated with current ClusterServlet."
+            )
+        return await self.obj_store.astatus()
+
+    def is_config_yaml_exists(self):
```
jlewitt1, line 314: do we need a helper here (makes more sense a general util i think but also is this re-used in more than one place?)

runhouse/constants.py
```diff
@@ -62,3 +63,7 @@
 # cluster status constants
 DOUBLE_SPACE_UNICODE = "\u00A0\u00A0"
 BULLET_UNICODE = "\u2022"
+DEFAULT_POST_STATUS_TO_DEN_INTERVAL_SIZE = 120
+# post status to den thread will start running 2 min after cluster servlet setup.
+# This delay enables the cluster setup to finish before running the post status job,
+DELAY_STATUS_POST_THREAD = 120
```
jlewitt1, line 69: name was confusing to me at first - maybe something more general like `STATUS_CHECK_DELAY`

runhouse/resources/hardware/cluster.py
```diff
@@ -263,6 +264,12 @@ def config(self, condensed=True):
             config["ssl_certfile"] = self.cert_config.cert_path
             config["ssl_keyfile"] = self.cert_config.key_path
 
+        try:
+            den_status_ping_interval = self.den_status_ping_interval
+        except AttributeError:
```
jlewitt1, line 269: why would this not exist?

runhouse/servers/cluster_servlet.py
```diff
@@ -217,3 +234,82 @@ async def aclear_all_references_to_env_servlet_name(self, env_servlet_name: str)
...
+            finally:
+
+                await asyncio.sleep(interval_size)
+
+    def send_status_info_to_den_sync(self):
+        asyncio.run(self.asend_status_info_to_den())
+
+    def schedule_post_status(self):
+        # adding post status to den thread
+        logger.info("adding send_status_info_to_den_sync to thread pool")
```
jlewitt1, line 291: maybe we should make these a `logger.debug`

runhouse/servers/cluster_servlet.py
```diff
@@ -41,8 +56,10 @@ async def __init__(
                     )
             self._last_activity = time.time()
             self._last_register = None
-            thread = threading.Thread(target=self.update_autostop, daemon=True)
-            thread.start()
+            autostop_thread = threading.Thread(target=self.update_autostop, daemon=True)
+            autostop_thread.start()
+
+        self.obj_store = None
```
rohinb2, line 56: I don't think we need to store a `self.obj_store` here? We should use the global `obj_store` associated with the ClusterServlet that's imported at the top of the file.

runhouse/servers/cluster_servlet.py
rohinb2: let's just throw this into the top level `utils.py`

rohinb2 (changes requested):

runhouse/servers/cluster_servlet.py
```diff
@@ -217,3 +228,82 @@ async def aclear_all_references_to_env_servlet_name(self, env_servlet_name: str)
...
+                logger.info(
+                    "Temporarily increasing the interval between two status pings to den. Next ping will be in an hour."
+                )
+                await asyncio.sleep(3600)
+            finally:
+
+                await asyncio.sleep(interval_size)
+
+    def send_status_info_to_den_sync(self):
+        asyncio.run(self.asend_status_info_to_den())
```
rohinb2, line 281: don't need to specify `sync` here, it's implied

runhouse/servers/cluster_servlet.py
```diff
@@ -217,3 +228,82 @@ async def aclear_all_references_to_env_servlet_name(self, env_servlet_name: str)
...
+
+    def schedule_post_status(self):
```
rohinb2, line 283: I think we can probably use an asyncio periodic here, instead of a thread. I'd start with that and switch to a thread if needed. The cluster servlet isn't doing expensive work so I'm not too worried about it.

More on this: https://stackoverflow.com/questions/37512182/how-can-i-periodically-execute-a-function-with-asyncio

Also an old PR of mine that does this, you can follow the pattern: https://github.com/run-house/runhouse/pull/701

runhouse/servers/cluster_servlet.py
```diff
@@ -217,3 +228,82 @@ async def aclear_all_references_to_env_servlet_name(self, env_servlet_name: str)
...
+                        f"Successfully updated cluster status in Den. Next ping to den in {round(DEFAULT_STATUS_CHECK_INTERVAL / 60, 2)} minutes."
+                    )
+            except Exception as e:
+                logger.error(
+                    f"Failed to post cluster status to den: {e}. Please check cluster logs for more info."
+                )
+                logger.info(
+                    "Temporarily increasing the interval between two status pings to den. Next ping will be in an hour."
+                )
+                await asyncio.sleep(3600)
```
rohinb2, line 275: what is this constant?

runhouse/main.py
```diff
@@ -685,6 +701,11 @@ def start(
     conda_env: str = typer.Option(
         None, help="Name of conda env corresponding to default env if it is a CondaEnv."
     ),
+    den_status_ping_interval: int = typer.Option(
+        DEFAULT_STATUS_CHECK_INTERVAL,
+        help="The time interval in seconds, that will pass between two cluster-status posts to Den. Relevant only to "
```
jlewitt1, line 706: help text here should be updated too

runhouse/servers/http/http_server.py
```diff
@@ -937,6 +938,13 @@ async def main():
         help="URL of Runhouse Den",
     )
 
+    parser.add_argument(
+        "--den-status-ping-interval",
+        type=int,
+        default=DEFAULT_STATUS_CHECK_INTERVAL,
+        help="Interval between two posts of the cluster status to Den",
```
jlewitt1, line 945: lets keep the text between all these consistent?

runhouse/servers/obj_store.py
```diff
@@ -198,6 +201,16 @@ async def ainitialize(
                 "Warning, cluster servlet is not initialized. Object Store operations will not work."
             )
 
+        owner_has_den_token = Path(CONFIG_YAML_PATH).expanduser().exists()
```
jlewitt1, line 204: name here is a little confusing, really this is just checking for the config file, not the token itself that should be in it

runhouse/servers/cluster_servlet.py
```diff
@@ -217,3 +228,82 @@ async def aclear_all_references_to_env_servlet_name(self, env_servlet_name: str)
...
+            logger.error(
+                "The relevant obj-store is not associated with the cluster servlet, therefore can't get cluster status."
+                " Setting the obj_store can be done using cluster_servlet.set_obj_store(associated_obj_store)."
+            )
+            raise ClusterServletError(
+                "Can't find the obj_store associated with current ClusterServlet."
+            )
+        return await self.obj_store.astatus()
+
+    # def is_config_yaml_exists(self):
```
jlewitt1, line 308: not used anymore?

runhouse/servers/cluster_servlet.py
```diff
@@ -217,3 +234,130 @@ async def aclear_all_references_to_env_servlet_name(self, env_servlet_name: str)
...
+
+        config_cluster = copy.deepcopy(self.cluster_config)
+
+        # poping out creds because we don't want to show them in the status
+        config_cluster.pop("creds", None)
+
+        # getting cluster servlets (envs) and their related objects
+        cluster_servlets = {}
+        cluster_envs_env_utilization_data = {}
+        for env in self._initialized_env_servlet_names:
```
rohinb2, line 328: nit, call this `env_servlet_name`

runhouse/servers/cluster_servlet.py
```diff
@@ -217,3 +234,130 @@ async def aclear_all_references_to_env_servlet_name(self, env_servlet_name: str)
...
+            try:
+                (
+                    objects_in_env_modified,
+                    env_utilization_data,
+                ) = await obj_store.acall_actor_method(
```
rohinb2, line 333: here, what you should do is instead collate via `asyncio.gather`. Instead of awaiting, you add everything.

In terms of handling the exceptions, you can follow this pattern: https://stackoverflow.com/questions/65579733/catch-exception-in-asyncio-gather. When you call `asyncio.gather`, have it return exceptions, and check the data there

runhouse/servers/cluster_servlet.py
```diff
@@ -217,3 +234,130 @@ async def aclear_all_references_to_env_servlet_name(self, env_servlet_name: str)
...
+                cluster_envs_env_utilization_data[env] = {}
+        config_cluster["envs"] = cluster_servlets
+        config_cluster["server_pid"] = get_pid()
+        config_cluster["runhouse_version"] = runhouse.__version__
+
+        # TODO: decide if we need this info at all: cpu_usage, memory_usage, disk_usage
+        cpu_usage = psutil.cpu_percent(interval=1)
+
+        # Fields: `available`, `percent`, `used`, `free`, `active`, `inactive`, `buffers`, `cached`, `shared`, `slab`
+        memory_usage = psutil.virtual_memory()._asdict()
```
rohinb2, line 349: maybe we can even enforce that these fields are there? by putting them into a model or something?

runhouse/servers/env_servlet.py
```diff
@@ -177,5 +179,137 @@ async def adelete_local(self, key: Any):
...
+                "memory_percent_from_cluster": (memory_size_bytes / total_memory) * 100,
+                "total_cluster_memory": total_memory,
+                "env_memory_info": psutil.virtual_memory(),
+            }
+        except psutil.NoSuchProcess:
+            env_memory_usage = {}
+
+        return (env_memory_usage, node_name, total_memory, env_servlet_pid, node_ip)
+
+    def _get_env_gpu_usage(self, env_servlet_pid):
```
rohinb2, line 230: mypy types

rohinb2 (changes requested):

runhouse/servers/cluster_servlet.py
```diff
@@ -217,3 +244,139 @@ async def aclear_all_references_to_env_servlet_name(self, env_servlet_name: str)
...
         return deleted_keys
+
+    ##############################################
+    # Cluster status functions
+    ##############################################
+
+    async def asend_status_info_to_den(self):
+        while True:
+            logger.info("Sending cluster status to Den")
+            interval_size = (await self.aget_cluster_config()).get(
```
jlewitt1, line 255: if `interval_size` is -1, shouldn't we just return here?

runhouse/main.py
```diff
@@ -161,16 +162,20 @@ def _resource_name_to_rns(name: str):
...
+    cluster_config: Dict, cluster_rh_version: str, server_pid: int
+):
     """
     Helping function to the `_print_status` which prints the relevant info from the cluster config.
     """
+    # TODO [SB]: need to modify printing format (colour palette etc).
     if "name" in cluster_config.keys():
         console.print(cluster_config.get("name"))
 
+    print(f"Runhouse v{cluster_rh_version}")
```
rohinb2, line 175: shouldn't we just move this out of the helper into `_print_status`?

runhouse/main.py
```diff
@@ -193,41 +198,46 @@ def _print_cluster_config(cluster_config: Dict):
...
 
     for key in top_level_config:
-        console.print(f"{BULLET_UNICODE} {key}: {cluster_config[key]}")
+        console.print(
+            f"{BULLET_UNICODE} {key.replace('_', ' ')}: {cluster_config[key]}"
+        )
 
-    console.print("\u2022 backend config:")
+    console.print(f"{BULLET_UNICODE} backend config:")
+    console.print(f"{DOUBLE_SPACE_UNICODE}{BULLET_UNICODE} server pid: {server_pid}")
```
rohinb2, line 206: and we can move this out as well

runhouse/servers/cluster_servlet.py
```diff
@@ -217,3 +244,139 @@ async def aclear_all_references_to_env_servlet_name(self, env_servlet_name: str)
...
+                status: ResourceStatusData = await self.astatus()
+                status_data = {
+                    "status": "running",
+                    "resource_type": status.cluster_config.get("resource_type"),
+                    "data": dict(status),
+                }
+                cluster_uri = rns_client.format_rns_address(
+                    status.cluster_config.get("name")
+                )
+                api_server_url = status.cluster_config.get(
```
rohinb2, line 268: it is a bit confusing to use `status.cluster_config` here, even though we already have. i'd use `(await self.aget_cluster_config())` just for clarity

runhouse/servers/cluster_servlet.py
```diff
@@ -217,3 +244,139 @@ async def aclear_all_references_to_env_servlet_name(self, env_servlet_name: str)
...
+                    data=json.dumps(status_data),
+                    headers=rns_client.request_headers(),
+                )
+                if post_status_data_resp.status_code != 200:
+                    logger.error(
+                        f"({post_status_data_resp.status_code}) Failed to send cluster status check to Den: {post_status_data_resp.text}"
+                    )
+                else:
+                    logger.info(
+                        f"Successfully updated cluster status in Den. Next status check will be {round(interval_size / 60, 2)} minutes."
```
rohinb2, line 282: nit, "Nex status check will be _in_"

runhouse/constants.py
```diff
@@ -62,3 +63,11 @@
 # cluster status constants
 DOUBLE_SPACE_UNICODE = "\u00A0\u00A0"
 BULLET_UNICODE = "\u2022"
+STATUS_CHECK_OFF = -1
```
jlewitt1, line 66: can we just reuse `DEFAULT_STATUS_CHECK_INTERVAL` instead of maintaining a separate variable?

runhouse/servers/cluster_servlet.py
```diff
@@ -217,3 +248,141 @@ async def aclear_all_references_to_env_servlet_name(self, env_servlet_name: str)
...
+    # Cluster status functions
+    ##############################################
+
+    async def asend_status_info_to_den(self):
+        while True:
+            logger.info("Sending cluster status to Den")
+            interval_size = (await self.aget_cluster_config()).get(
+                "den_status_ping_interval"
+            )
+            if interval_size == STATUS_CHECK_OFF:
```
jlewitt1, line 262: i think we need this to sit inside the `try` block, so that we still execute the `sleep` inside the `finally`. also `break` instead of `return` might be more appropriate here

runhouse/constants.py
BelSasha: thought about it, but `DEFAULT_STATUS_CHECK_INTERVAL` can be also a positive integer (if in the future we'll decide to turn on pinging to den by default) and in that case we would not like to return.  @jlewitt1

runhouse/servers/cluster_servlet.py
```diff
@@ -217,3 +244,141 @@ async def aclear_all_references_to_env_servlet_name(self, env_servlet_name: str)
...
+                else:
+                    logger.info(
+                        f"Successfully updated cluster status in Den. Next status check will be in {round(interval_size / 60, 2)} minutes."
+                    )
+            except Exception as e:
+                logger.error(
+                    f"Cluster status check has failed: {e}. Please check cluster logs for more info."
+                )
+                logger.warning(
+                    f"Temporarily increasing the interval between two consecutive status checks. Next status check will be in {round(INCREASED_STATUS_CHECK_INTERVAL / 60, 2)} minutes. For changing the interval size, please restart the server."
```
jlewitt1, line 291: wouldn't changing the interval size also require updating the `den_status_ping_interval` on the client side as well? (in addition to restarting the server)

BelSasha: When we restart the server, the` den_status_ping_interval` is updated in the cluster_config.json which is saved on the cluster. Later, the `send_status_info_to_den` job extracts the value from the cluster_config. I could add to the log something like "When restarting, please provide a new interval size. If not provided, interval size will be set to {DEFAULT_INTERVAL_SIZE}". @jlewitt1

jlewitt1: cool yea i meant just a way to indicate more specifically how to actually provide that new interval, makes sense

jlewitt1 (approved):

PR #836, opened 2024-05-26
Enable the user change the setup of the cluster check process 

BelSasha:


BelSasha:
* **#836** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/836?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

runhouse/resources/hardware/cluster.py
```diff
@@ -1645,3 +1658,44 @@ def _check_for_child_configs(cls, config: dict):
...
+        saved_to_den = rns_client.exists(self.name)
+        if not saved_to_den:
+            logger.error(
+                f"The cluster must be saved to Den to {change_type} periodic status checks."
+            )
+            return False
+
+        return True
+
+    def _disable_periodic_status_check(self):
```
jlewitt1, line 1684: maybe "disable_status_ check"

runhouse/resources/hardware/cluster.py
```diff
@@ -1645,3 +1658,44 @@ def _check_for_child_configs(cls, config: dict):
...
+        """
+        Stopping consecutively sending status to Den.
+        """
+        if not self._is_den_authed_and_saved_to_den(change_type="disable"):
+            return
+        self.save_config_to_cluster(status_check_interval=-1)
+
+    def _change_periodic_status_check_interval(self, new_interval: int):
```
jlewitt1, line 1692: thoughts on "enable_status_check"?

runhouse/resources/hardware/cluster.py
```diff
@@ -1645,3 +1658,44 @@ def _check_for_child_configs(cls, config: dict):
             config["default_env"] = default_env
 
         return config
+
+    def _is_den_authed_and_saved_to_den(self, change_type: str):
```
jlewitt1, line 1662: wondering if this check is necessary, is there any harm in just updating the config with the interval field regardless if status check is explicitly enabled or disabled?

runhouse/servers/cluster_servlet.py
```diff
@@ -266,10 +283,22 @@ async def asend_status_info_to_den(self):
...
+                    await self.aupdate_status_check_interval_in_cluster_config()
+                )
+                interval_size = (await self.aget_cluster_config()).get(
+                    "status_check_interval", DEFAULT_STATUS_CHECK_INTERVAL
+                )
                 if interval_size == -1:
+                    if is_config_updated:
+                        logger.info(
+                            f"Disabled periodic cluster status check. For enabling it, please run "
+                            f"cluster.restart_server(). If you want to set up an interval size that is not the "
```
jlewitt1, line 298: instead of callijng restart server should call the enable status check method

runhouse/servers/cluster_servlet.py
```diff
@@ -266,10 +283,22 @@ async def asend_status_info_to_den(self):
...
             try:
-                interval_size = DEFAULT_STATUS_CHECK_INTERVAL
+                is_config_updated = (
+                    await self.aupdate_status_check_interval_in_cluster_config()
+                )
+                interval_size = (await self.aget_cluster_config()).get(
+                    "status_check_interval", DEFAULT_STATUS_CHECK_INTERVAL
+                )
                 if interval_size == -1:
+                    if is_config_updated:
```
jlewitt1, line 295: maybe we should add a log here anytime the config is updated, not just when its turned off

runhouse/servers/cluster_servlet.py
```diff
@@ -118,6 +119,22 @@ def update_autostop(self):
     async def aget_cluster_config(self) -> Dict[str, Any]:
         return self.cluster_config
 
+    async def aupdate_status_check_interval_in_cluster_config(self):
+        cluster_path = Path("~/.rh/cluster_config.json").expanduser()
```
jlewitt1, line 123: i think we have a constant for this file path somewhere?

runhouse/servers/cluster_servlet.py
```diff
@@ -118,6 +119,22 @@ def update_autostop(self):
...
+        with open(cluster_path) as cluster_local_config:
+            local_config = json.load(cluster_local_config)
+            local_interval = local_config.get("status_check_interval")
+            servlet_interval = self.cluster_config.get("status_check_interval")
+            if local_interval != servlet_interval:
```
jlewitt1, line 128: this relates to my other comment, but this doesn't account for None values which it probably should

runhouse/servers/cluster_servlet.py
BelSasha: @jlewitt1 This log is added as part of `aupdate_status_check_interval_in_cluster_config` method

runhouse/resources/hardware/cluster.py
BelSasha: @jlewitt1 Removed the check if the cluster is saved into den. If it is not, the post request will raise an error, which we catch and send an error log, explaining that the cluster is not saved and therefore we can't post its status.

If left the check of den_auth == True because: 
1. It is not pinging den, therefore it is a "cheap" check
2. I think it is in our interests to clarify to the user that this is a feature requires Den auth / Den user. wdyt?

runhouse/resources/hardware/cluster.py
```diff
@@ -1645,3 +1658,40 @@ def _check_for_child_configs(cls, config: dict):
             config["default_env"] = default_env
 
         return config
+
+    def _is_den_authed(self, change_type: str):
```
jlewitt1, line 1662: for simplicity maybe we remove this separate method and just check `if self.den_auth` in-line? can still add logs if appropriate

runhouse/resources/hardware/cluster.py
```diff
@@ -1645,3 +1658,40 @@ def _check_for_child_configs(cls, config: dict):
...
+        """
+        if not self._is_den_authed(change_type="disable"):
+            return
+        self.save_config_to_cluster(status_check_interval=-1)
+
+    def _enable_or_update_status_check(
+        self, new_interval: int = DEFAULT_STATUS_CHECK_INTERVAL
+    ):
+        """
+        Enables periodic status check or changing the interval between to consecutive cluster status checks.
```
jlewitt1, line 1690:
```suggestion
        Enables a periodic status check or updates the interval between consecutive status checks.
```

runhouse/servers/cluster_servlet.py
```diff
@@ -303,12 +337,11 @@ async def asend_status_info_to_den(self):
                 logger.warning(
                     f"Temporarily increasing the interval between two consecutive status checks. "
                     f"Next status check will be in {round(INCREASED_STATUS_CHECK_INTERVAL / 60, 2)} minutes. "
-                    f"For changing the interval size, please restart the server with a new interval size value. "
+                    f"For changing the interval size, please run cluster._change_periodic_status_check_interval(new_interval). "
```
jlewitt1, line 340: dont think this method exists?

runhouse/resources/hardware/cluster.py
```diff
@@ -1645,3 +1658,30 @@ def _check_for_child_configs(cls, config: dict):
             config["default_env"] = default_env
 
         return config
+
+    def _disable_status_check(self):
+        """
+        Stopping consecutively sending status to Den.
+        """
+        if not self.den_auth:
```
jlewitt1, line 1666: what happens if i have a cluster with den auth enabled, then i decide to disable it. Will the status check still continue in this case?

runhouse/resources/hardware/cluster.py
```diff
@@ -1645,3 +1658,30 @@ def _check_for_child_configs(cls, config: dict):
             config["default_env"] = default_env
 
         return config
+
+    def _disable_status_check(self):
+        """
+        Stopping consecutively sending status to Den.
```
jlewitt1, line 1664: the "consecutively" part is a little confusing to me

PR #860, opened 2024-06-03
fix cluster status scheduler init

BelSasha:


runhouse/servers/cluster_servlet.py
```diff
@@ -79,7 +79,7 @@ async def __init__(
 
         # Only send for clusters that have den_auth enabled and if we are logged in with a user's token
         # to authenticate the request
-        if self.cluster_config.get("den_auth", False) and configs.token:
+        if self.cluster_config.get("den_auth", False):
```
rohinb2, line 82: can you add the check to see if `Path.exists` for the `~/.rh/config.yaml` I think you had this before and I switched it

rohinb2: Also I don't get how `configs.token` wouldn't work as a check here -- wouldn't it pull from the backing file?

BelSasha: @rohinb2: Up until now, when `cluster.up()` was called for the first time, the cluster servlet was initialed *before* the `config.yaml `was saved on the cluster. Therefore, both `Path.exists` for the` ~/.rh/config.yaml` and `configs.token` wouldn't work, because the `config.yaml` did not exist on the cluster at the time of check. 

I changed a bit the logic of the restart_server(), so the config.yaml will be created before spinning up / restarting the cluster servlert. I ensured it would be created only if it doesn't exist on the cluster. (Meaning it will be created only when we run `cluster.up()` on a new / terminated cluster <=> `restart_server()` runs for the first time. This way, if someone with access to my cluster will run `restart_server()` on my running cluster, his or her token will not override the token in the existing `config.yaml`. 

Any thoughts / objections?

runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -446,19 +446,6 @@ def up(self):
 
         self.restart_server()
 
-        # Save a limited version of the local ~/.rh config to the cluster with the user's hashed token
```
rohinb2, line 449: Why are you not sending this anymore?

runhouse/servers/cluster_servlet.py
rohinb2: This sounds like a great plan!! I'll read through this more thoroughly

runhouse/resources/hardware/on_demand_cluster.py
BelSasha: Moved it into the restart_server() logic

runhouse/resources/hardware/cluster.py
```diff
@@ -838,6 +851,33 @@ def restart_server(
         # Update the cluster config on the cluster
         self.save_config_to_cluster()
 
+        # Save a limited version of the local ~/.rh config to the cluster with the user's hashed token,
+        # if such does not exist on the cluster
+
+        check_config_yaml_cmd = ["cat ~/.rh/config.yaml"]
+        check_config_yaml_res = self._run_cli_commands_on_cluster(
```
rohinb2, line 858: here, can you just do `self.run`? why do we need this helper

runhouse/resources/hardware/cluster.py
```diff
@@ -859,17 +899,7 @@ def restart_server(
...
-                commands=[cmd], env=self._default_env, node=self.address
-            )
-        else:
-            status_codes = self._run_commands_with_ssh(
-                commands=[cmd],
-                cmd_prefix=self._default_env._run_cmd if self._default_env else "",
-                env_vars=self._default_env.env_vars if self._default_env else {},
-                node=self.address,
-            )
+        status_codes = self._run_cli_commands_on_cluster(commands=[cmd])
```
rohinb2, line 902: can we just do `self.run` here too

rohinb2: or we can leave the old logic here

PR #892, opened 2024-06-13
improvement: add support for running in a marimo notebook

akshayka:
This makes runhouse work in [marimo notebooks](https://github.com/marimo-team/marimo), in addition to Jupyter notebooks.

### DIFF
--- a/runhouse/resources/module.py
+++ b/runhouse/resources/module.py
@@ -968,6 +968,24 @@ class Module(Resource):
         return module_path
 
     @staticmethod
+    def _is_running_in_notebook(module_path: Union[str, None]) -> bool:
+        """Returns True if running in a notebook, False otherwise"""
+
+        # Check if running in an IPython notebook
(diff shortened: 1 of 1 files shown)


dongreenberg:
Love it! Thanks for the PR. One question - IIRC Marimo notebooks are .py files, so wouldn't they be importable natively on the server (and not require a short-circuit to write out the user's Module into a separate file)?

akshayka:
> Love it! Thanks for the PR. One question - IIRC Marimo notebooks are .py files, so wouldn't they be importable natively on the server (and not require a short-circuit to write out the user's Module into a separate file)?

They are Python files. But the function definitions are in their own lexical scopes (notebook files are lazy), so the server won't be able to find them — unless it knew it was looking at a marimo notebook, and knew how to extract the function. We'll have an API for that in the future but don't have it yet.

dongreenberg:
Ok! That makes sense, and appreciate the contribution to make Runhouse work with Marimo.

PR #893, opened 2024-06-14
Llama 3 vLLM GCP example

mkandler:


mkandler:
* **#893** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/893?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

examples/llama3-vllm-gcp/llama3_vllm_gcp.py
```diff
@@ -0,0 +1,188 @@
...
+# The Python code we'll run is contained in an asynchronous function, `main`. To make this guide more readable, it's
+# contents are rendered as top-level code snippets, but they should be included in the `main` method for running.
+# :::
+async def main():
+    gpu_cluster = rh.cluster(
+        name="rh-l4x",
+        instance_type="L4:1",
+        memory="32+",
+        provider="gcp",
+        # (Optional) Include the following to create exposed TLS endpoints
```
jlewitt1, line 104: not sure we've been doing this for other examples, but thoughts on adding a note on how to incorporate autostop? been getting a lot of questions about that lately

examples/llama3-vllm-gcp/llama3_vllm_gcp.py
```diff
@@ -0,0 +1,188 @@
...
+# $ sky check
+# ```
+# We'll be downloading the Llama 3 model from Hugging Face, so we need to set up our Hugging Face token:
+# ```shell
+# $ export HF_TOKEN=<your huggingface token>
+# ```
+#
+# ## Define a Llama 3 model class
+# We import `runhouse` and `asyncio` because that's all that's needed to run the script locally.
+# The actual vLLM imports can happen within the functions that will be sent to the Runhouse cluster via an env.
```
jlewitt1, line 37: minor point of clarification - i read this as the env is responsible for sending the functions to the cluster, rather than representing the environment on the cluster in which the function itself is served

examples/llama3-vllm-gcp/llama3_vllm_gcp.py
mkandler: Oh yeah... I definitely think we should do that! Maybe including it here in the cluster AND also at the end of the example explain how to use `sky status` `sky down`?

jlewitt1: i think showing how to set the `autostop` field in the factory would probably suffice for now, but if it doesn't feel too cluttered to add the sky commands / link to the docs that describe them that that could work too

examples/llama3-vllm-gcp/llama3_vllm_gcp.py
mkandler: Updated a little - lmk if that's better?

examples/llama3-vllm-gcp/llama3_vllm_gcp.py
mkandler: Made an update with both just because I'm not sure it's super easy to find elsewhere - lmk if that seems good!

examples/llama3-vllm-gcp/llama3_vllm_gcp.py
```diff
@@ -0,0 +1,192 @@
...
+    )
+    for text_output in ans:
+        print(f"... Generated Text:\n{prompt}{text_output}\n")
+
+    # :::note{.info title="Note"}
+    # Your initial run of this script may take a few minutes to deploy an instance on GCP, set up the environment,
+    # and load the Llama 3 model. Subsequent runs will reuse the cluster and generally take seconds.
+    # :::
+
+    # ### Advanced: Sharing and TLS endpoints
```
rohinb2, line 155: do we want to include this? It seems overkill / could include in a different example, but I'm fine with it if you think so

rohinb2 (approved):
looks good, awesome job Matt

examples/llama3-vllm-gcp/llama3_vllm_gcp.py
mkandler: It seemed like something missing from the other examples to me and possibly a nature way to highlight the value of Den to new users - I think it's easy to miss out on that value entirely. We could experiment with it here and remove later if we have more/better ways to introduce sharing/den in the examples?

PR #907, opened 2024-06-18
Setting log level when cluster is started

BelSasha:


BelSasha:
* **#907** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/907?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

runhouse/logger.py
```diff
@@ -90,6 +91,51 @@ def formatTime(self, record, datefmt=None):
             return dt.isoformat(timespec="milliseconds")
 
 
+def get_logger(name: str = __name__, log_level: Union[int, str] = logging.INFO):
+    level_name = (
+        logging.getLevelName(log_level) if isinstance(log_level, int) else log_level
+    )
+    LOGGING_CONFIG = {
```
jlewitt1, line 98: this is a bit redundant, since we are defining the config here in this function and also below as part of the default logger.

instead of having two paths to import the logger:
(1) Default
```
import logging
logger = logging.getLogger(__name__)
```
(2) Customized Log Level
```
from runhouse.logger import get_logger
logger = get_logger(name=__name__, log_level=DEFAULT_LOG_LEVEL)
```

maybe we should have everything flow thru the second path. wdyt?

runhouse/main.py
```diff
@@ -798,9 +806,12 @@ def restart(
         False,
         help="Whether HTTP server started from inside a Python call rather than CLI.",
     ),
+    log_level: str = typer.Option(
+        default=DEFAULT_LOG_LEVEL, help="The lowest log level of the printed logs"
```
jlewitt1, line 810:
```suggestion
        default=DEFAULT_LOG_LEVEL, help="Minimum log level for logs to be printed"
```

runhouse/main.py
```diff
@@ -798,9 +806,12 @@ def restart(
...
+    ),
 ):
     """Restart the HTTP server on the cluster."""
-    if name:
+    if name and log_level == DEFAULT_LOG_LEVEL:
```
jlewitt1, line 814: not sure I follow this logic - why do we treat restart differently if the log level is "INFO"?

runhouse/main.py
```diff
@@ -798,9 +806,12 @@ def restart(
         False,
         help="Whether HTTP server started from inside a Python call rather than CLI.",
     ),
+    log_level: str = typer.Option(
```
jlewitt1, line 809: I think we also need to enforce that the value gets converted to uppercase - should be able to do this: 

```
log_level: str = typer.Option(
    default=DEFAULT_LOG_LEVEL, 
    help="Minimum log level for logs to be printed",
    callback=lambda value: value.upper()
)
```

runhouse/servers/cluster_servlet.py
```diff
@@ -356,3 +356,8 @@ async def astatus(self):
 
     def status(self):
         return sync_function(self.astatus)()
+
+    def set_cluster_logger(self, logs_level: str = DEFAULT_LOG_LEVEL):
+        self.logger = get_logger(name=__name__, log_level=logs_level)
+        self.logger.debug(f"Set logger, with level {logs_level}")
```
jlewitt1, line 362: why do we need both debug and info logs here?

runhouse/servers/env_servlet.py
```diff
@@ -65,6 +62,8 @@ class EnvServlet:
     async def __init__(self, env_name: str, *args, **kwargs):
         self.env_name = env_name
 
+        if "logger" in kwargs.keys() and kwargs.get("logger"):
```
jlewitt1, line 65: instead of passing the actual logger here, wouldn't it be easier to pass the log level, and then adjust accordingly?

runhouse/servers/http/http_server.py
```diff
@@ -1153,13 +1170,13 @@ async def main():
...
 
     await obj_store.aset_cluster_config(cluster_config)
-    logger.info("Updated cluster config with parsed argument values.")
 
     await HTTPServer.ainitialize(
         default_env_name=default_env_name,
         conda_env=conda_name,
         enable_local_span_collection=use_local_telemetry
         or configs.data_collection_enabled(),
+        logs_level=parse_args.log_level,
```
jlewitt1, line 1179:
```suggestion
        logs_level=parse_args.log_level.upper(),
```

runhouse/servers/http/http_server.py
```diff
@@ -1253,6 +1270,10 @@ async def main():
         if env_vars:
             cluster.call(cluster.default_env.name, "_set_env_vars", env_vars)
 
+    log_level = parse_args.log_level or DEFAULT_LOG_LEVEL
+    logger.info(f"inside_http main, log_level {log_level}")
```
jlewitt1, line 1274:
```suggestion
    logger.info(f"Setting log level to: {log_level}")
```

runhouse/servers/http/http_server.py
```diff
@@ -1253,6 +1270,10 @@ async def main():
         if env_vars:
             cluster.call(cluster.default_env.name, "_set_env_vars", env_vars)
 
+    log_level = parse_args.log_level or DEFAULT_LOG_LEVEL
```
jlewitt1, line 1273:
```suggestion
    log_level = parse_args.log_level.upper() or DEFAULT_LOG_LEVEL
```

runhouse/servers/http/http_server.py
jlewitt1, line 1179: i think we need to be case sensitive here, would just confirm that the previous change to the CLI arg enforces this is uppercase

runhouse/servers/obj_store.py
```diff
@@ -1587,3 +1589,9 @@ async def astatus(self):
 
     def status(self):
         return sync_function(self.astatus)()
+
+    async def aset_cluster_logger(self, logs_level: str = DEFAULT_LOG_LEVEL):
+        logger.info(f"obj_store.aset_cluster_logger, logs_level: {logs_level}")
```
jlewitt1, line 1594:
```suggestion
        logger.info(f"Setting cluster log level to: {logs_level}")
```

runhouse/servers/http/http_server.py
```diff
@@ -206,6 +211,7 @@ async def _add_username_to_span(request: Request, call_next):
             env_name=default_env_name,
             create=True,
             runtime_env=runtime_env,
+            logger=logger,
```
jlewitt1, line 214: just pass in the log level here? this would allow us to always import the same logger, instead of initializing different ones (related to other comment), and then the cluster servelt can just set the log level passed in: 
`logger.setLevel(log_level)`

runhouse/servers/cluster_servlet.py
```diff
@@ -54,17 +52,17 @@ async def __init__(
         self._key_to_env_servlet_name: Dict[Any, str] = {}
         self._auth_cache: AuthCache = AuthCache(cluster_config)
         self.autostop_helper = None
-
+        self.logger = get_logger(name=__name__, log_level=DEFAULT_LOG_LEVEL)
```
jlewitt1, line 55: any reason why we cant continue with the global logger import at the top of the file?

runhouse/servers/http/http_server.py
```diff
@@ -119,6 +120,9 @@ async def ainitialize(
         *args,
         **kwargs,
     ):
+        from runhouse.logger import logger
+
+        logger.setLevel(kwargs.get("logs_level", DEFAULT_LOG_LEVEL))
```
jlewitt1, line 125: we re-use this further down, prob worth setting to a variable (`kwargs.get("logs_level", DEFAULT_LOG_LEVEL)`)

runhouse/servers/env_servlet.py
```diff
@@ -64,7 +61,11 @@ async def wrapper(*args, **kwargs):
 class EnvServlet:
     async def __init__(self, env_name: str, *args, **kwargs):
         self.env_name = env_name
+        if "logs_level" in kwargs.keys() and kwargs.get("logs_level"):
```
jlewitt1, line 64: i think we can simplify this block to remove the if statement and just do:
`logger.setLevel(kwargs.get("logs_level", DEFAULT_LOG_LEVEL))`

runhouse/servers/http/http_server.py
```diff
@@ -1253,6 +1269,10 @@ async def main():
         if env_vars:
             cluster.call(cluster.default_env.name, "_set_env_vars", env_vars)
 
+    log_level = parse_args.log_level or DEFAULT_LOG_LEVEL
```
jlewitt1, line 1272: do we need the "or" here? won't it have the default value anyway if not explicitly provided?

runhouse/servers/cluster_servlet.py
```diff
@@ -433,3 +431,7 @@ async def asend_cluster_logs_to_den(self):
 
     def send_cluster_logs_to_den(self):
         asyncio.run(self.asend_cluster_logs_to_den())
+
+    def set_cluster_logger(self, logs_level: str = DEFAULT_LOG_LEVEL):
```
jlewitt1, line 435: a little confused by this (also related to my previous comment about initialize a global logger object), instead of re-initializing the logger why can't we just reset its level?

runhouse/logger.py
```diff
@@ -97,35 +100,82 @@ def formatTime(self, record, datefmt=None):
...
         },
-    },
-}
+    }
+    logging.config.dictConfig(LOGGING_CONFIG)
+    logger = logging.getLogger(name=name)
+    return logger
+
+
+logger = get_logger(name=__name__, log_level=DEFAULT_LOG_LEVEL)
```
jlewitt1, line 148: what's this used for?

jlewitt1: actually this might give us a cleaner way of importing the logger everywhere else - i think we can then do: 

```
from runhouse.logger import logger
logger.info("Using the default logger instance")```

File: runhouse/logger.py [hunk 18]
```diff
@@ -97,35 +100,82 @@ def formatTime(self, record, datefmt=None):
...
+
+# LOGGING_CONFIG = {
```
jlewitt1, line 150: delete this?

runhouse/logger.py
```diff
@@ -1,7 +1,10 @@
 import logging
+import logging.config
```
jlewitt1, line 2: why do we import this here?

runhouse/main.py
```diff
@@ -43,7 +44,7 @@
 
 # For printing with typer
 console = Console()
-logger = logging.getLogger(__name__)
+from runhouse.logger import logger
```
jlewitt1, line 47: i think we can even simplify this further to: `from runhouse import logger`

if we add `from .logger import logger` to the main `__init__`

runhouse/resources/hardware/cluster.py
```diff
@@ -895,6 +895,7 @@ def restart_server(
                 else ""
             )
             + " --from-python"
+            + f" --log-level {logging.getLevelName(logging.root.level)}"
```
jlewitt1, line 898: re: whether to have this API support the same level of control as the CLI, maybe we just add a `log_level` arg here and pass that in instead? pretty harmless to add one extra param

runhouse/logger.py
BelSasha: It is used in the `get_logger()` method @jlewitt1

runhouse/resources/folders/folder.py
```diff
@@ -23,7 +22,8 @@
 # SSHFileSystem is not yet builtin.
 # Line above suggested by fsspec devs: https://github.com/fsspec/filesystem_spec/issues/1071
 
-logger = logging.getLogger(__name__)
+from runhouse.logger import logger
```
jlewitt1, line 25: i think we can now do `from runhouse import logger`?

runhouse/servers/env_servlet.py
```diff
@@ -63,12 +61,19 @@ async def wrapper(*args, **kwargs):
 
 class EnvServlet:
     async def __init__(self, env_name: str, *args, **kwargs):
+        from runhouse.logger import logger
+
         self.env_name = env_name
 
+        logs_level = kwargs.get("logs_level", DEFAULT_LOG_LEVEL)
+        logger.setLevel(logs_level)
+        self.logger = logger
```
jlewitt1, line 70: why do we need to re-initialize a logger attribute here instead of using the global one?

PR #916, opened 2024-06-18
Propagate `self.env` if it was set.

rohinb2:
Breaking example, where we initialize `reqs` in the constructor, which gets set in the env, but is not propagated in `.to` because the env has no name. Pretty sure we can just send `self.env` and later we have a check if it has no name to take a different path.

See test for repro.

### DIFF
--- a/tests/test_resources/test_envs/test_env.py
+++ b/tests/test_resources/test_envs/test_env.py
@@ -30,6 +30,12 @@ def np_summer(a, b):
     return int(np.sum([a, b]))
 
 
+def torch_import():
+    import torch
+
+    return str(torch.__version__)
+
+
 @pytest.mark.envtest
 class TestEnv(tests.test_resources.test_resource.TestResource):
     MAP_FIXTURES = {"resource": "env"}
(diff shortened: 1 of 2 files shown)


rohinb2:
* **#916** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/916?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tests/test_resources/test_envs/test_env.py
```diff
@@ -30,6 +30,12 @@ def np_summer(a, b):
     return int(np.sum([a, b]))
 
 
+def torch_import():
+    import torch
```
carolineechen, line 34: could we use a diff library that's not used in any other test env/function yet to make sure its standalone? aws_cpu fixture now uses a torch docker image

carolineechen (approved):
minor comment on the test but core logic looks good

dongreenberg (approved):

rohinb2:
### Merge activity

* **Jun 20, 2:47 PM EDT**: @rohinb2 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/916).

PR #929, opened 2024-06-24
Run cluster status check and log surfacing in the same thread

BelSasha:


BelSasha:
* **#929** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/929?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

runhouse/rns/rns_client.py
```diff
@@ -632,18 +642,16 @@ def contents(self, name_or_path, full_paths):
             full_paths=full_paths
         )
 
-    async def send_status(self, status: ResourceStatusData, cluster_rns_address: str):
+    async def send_status(
```
jlewitt1, line 645: why do we need to pass around `cluster_uri` and api_server_url` here, don't we have it in the `cluster_config`?

runhouse/servers/cluster_servlet.py
```diff
@@ -255,9 +242,35 @@ async def aperiodic_status_check(self):
...
+                            "api_server_url", rns_client.api_server_url
+                        )
+
+                        sent_status = await rns_client.send_status(
+                            status=status,
+                            cluster_uri=cluster_uri,
+                            api_server_url=api_server_url,
+                        )
+
+                        if not sent_status:
```
jlewitt1, line 258: aren't we already surfacing the success/failure logs already for the status + logs (in `rns_client.py`)?

runhouse/rns/rns_client.py
BelSasha: @jlewitt1 we have the cluster name, and we need to modify it to be in the URI format. Since we need the URI both for cluster status check and the logs surfacing, I thought it made sense to modify it once and send it to both methods as arguments. wdyt?

PR #938, opened 2024-06-26
Refactor cluster status check output structure

BelSasha:


BelSasha:
* **#938** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/938?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

runhouse/resources/hardware/cluster.py
```diff
@@ -275,6 +276,8 @@ def config(self, condensed=True):
             config["ssl_certfile"] = self.cert_config.cert_path
             config["ssl_keyfile"] = self.cert_config.key_path
 
+        cpu_or_cuda = detect_cuda_version_or_cpu(self)
+        config["process_unit"] = cpu_or_cuda.upper() if cpu_or_cuda == "cpu" else "GPU"
```
dongreenberg, line 280: This isn't really what the config is for - I'd rather just cache this propery within the ClusterServelet than add it to the config. "processing_unit" also isn't exactly an accurate descriptor - A GPU box also has CPUs, and there are different kinds of GPUs. I'd just call it "has_cuda" for now.

runhouse/servers/cluster_servlet.py
```diff
@@ -71,6 +72,12 @@ async def aget_cluster_config(self) -> Dict[str, Any]:
         return self.cluster_config
 
     async def aset_cluster_config(self, cluster_config: Dict[str, Any]):
+        if "has_cuda" not in cluster_config.keys():
+            cpu_or_cuda = detect_cuda_version_or_cpu()
+            cluster_config["has_cuda"] = (
```
jlewitt1, line 77: shouldn't this be a boolean?

runhouse/servers/env_servlet.py
```diff
@@ -183,14 +183,12 @@ async def adelete_local(self, key: Any):
     async def aclear_local(self):
         return await obj_store.aclear_local()
 
-    def _get_env_cpu_usage(self):
+    def _get_env_cpu_usage(self, cluster_config: dict = obj_store.cluster_config):
```
jlewitt1, line 186: setting the default to a mutable property like that feels wrong, can we just set to `None` instead and then do:
`cluster_config = cluster_config or obj_store.cluster_config`?

runhouse/servers/cluster_servlet.py
BelSasha: Yeah, makes more sense that it will be a boolean.  
@mkandler In case you need to know on the frontend side whether the cluster has GPU, you can check if `cluster_config["has_cuda"] == True`

runhouse/servers/cluster_servlet.py
```diff
@@ -71,6 +72,9 @@ async def aget_cluster_config(self) -> Dict[str, Any]:
         return self.cluster_config
 
     async def aset_cluster_config(self, cluster_config: Dict[str, Any]):
+        if "has_cuda" not in cluster_config.keys():
```
jlewitt1, line 75: is there any risk that a stale `has_cuda` value would be stored here? my point being any harm in just removing this check and setting it each time?

BelSasha: hmmm, checking each time might increase the runtime a bit, but don't think it will cause massive latency. However, everytime the cluster is up (meaning the clusterServlet actor is up) => the "has cuda" value remains the same, therefore I don't see a real reason to set it each time.  Don't have strong opinion abouth this though. 

@jlewitt1 wdyt?

jlewitt1: gotcha yea so then we can keep as is then

PR #953, opened 2024-07-01
Update ray stop command

carolineechen:
runhouse ray not being fully terminated prior to starting ray again sometimes causing problems`


carolineechen:
* **#953** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/953?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

dongreenberg:
Hm..... not sure this is the right way to shut ray down. I think there's a brief grace period where Ray tries to clean up all its processes (which might be necessary to cleanly free up complex resources like GPU memory), and this could disrupt that. We should check whether killing Ray this way is ok for our use cases. It looks like Ray has a grace period of 10 seconds to clean up when you run `ray stop`, which is very long. I'm adding a linear issue to confirm 1) when we restart_server it doesn't break autostop and 2) all the ray processes are cleaned up (probably need to grab all the PIDs before restarting).

PR #956, opened 2024-07-02
Docs installation page

mkandler:
Meant to be a page we can easily link to in the introduction of any tutorial - so we don't find ourselves repeating all the installation steps over and over.

Are there any "gotchas" I missed here?

- [x] Need to add a link to this in the navigation too (Other Topics to start?)

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -1,0 +1,195 @@
+Installation and Setup
+======================
+
+An overview of common installation and setup steps when working with Runhouse.
+
+- Installing the Runhouse library
+- Cloud providers and credentials
+- Cluster lifecycle management
+- Authenticate with Runhouse Den
+
+Install the Runhouse library
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Optionally, begin by setting up a virtual env with `Conda <https://conda.io/projects/conda/en/latest/user-guide/install/index.html>`
+or your package and environment management tool of choice. This allows you to maintain an environment with a
+specific version of Python and installed packages.
+
+.. code-block:: shell
+
+    $ conda create -n rh_env python=3.9.15
+    $ conda activate rh_env
+
+For local-only development, the Runhouse base package can be installed with:
+
+.. code-block:: shell
+    $ pip install runhouse
+
+**Recommended**: To use Runhouse to launch on-demand clusters, please instead run the following command.
+This additionally installs SkyPilot, which is used for launching fresh VMs through your cloud provider.
+
+.. code-block:: shell
+    $ pip install "runhouse[sky]"
+
+Alternatively, if you plan to use Runhouse with a specific cloud provider, you can install the package with
+that provider's CLI included - or choose multiple providers. SkyPilot will also be included with these.
+
+.. code-block:: shell
+    # Cloud-specific installation
+    $ pip install "runhouse[aws]"
+    # Include multiple providers
+    $ pip install "runhouse[aws,gcp]"
+
+
+Configure Cloud Credentials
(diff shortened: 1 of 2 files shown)


docs/installation.rst
```diff
@@ -0,0 +1,217 @@
...
+
+- Installing the Runhouse library
+- Cloud providers and credentials
+- Cluster lifecycle management
+- Authenticate with Runhouse Den
+
+Install the Runhouse library
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Optionally, to begin, set up a virtual environment with `Conda <https://conda.io/projects/conda/en/latest/user-guide/install/index.html>`.
```
jlewitt1, line 14: Conda is one type of virtual env that a user might use,  can we make it more clear that it isn't the only path?

docs/installation.rst
```diff
@@ -0,0 +1,217 @@
...
+    $ az account set -s <subscription_id>
+    $ sky check
+
+For more info: `SkyPilot | Azure <https://skypilot.readthedocs.io/en/latest/getting-started/installation.html#azure>`
+
+
+Other Providers
+---------------
+
+For a full list of providers and configuration details: `SkyPilot documentation <>`.
```
jlewitt1, line 135: missing link? (https://skypilot.readthedocs.io/en/latest/getting-started/installation.html#cloud-account-setup)

docs/installation.rst
```diff
@@ -0,0 +1,217 @@
...
+- & more
+
+
+Cluster Lifecycle Management
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To ensure that you have full control over the availability of your clusters, and the cost associated
+with maintaining them, you can may find the following SkyPilot commands helpful:
+
+- ``sky status`` - Displays a list of all your clusters by name.
```
jlewitt1, line 149: thoughts on also linking to the sky API docs for each command, or you think thats just information overload?

docs/installation.rst
```diff
@@ -0,0 +1,217 @@
...
+- ``Upload your local .rh config to Runhouse? [y/N]:`` - Use updates to your local config to modify your saved account settings
+
+If you are running Runhouse strictly in code (like in a notebook), you can also login to your account
+with the Python API:
+
+.. code-block:: python
+    import runhouse as rh
+    rh.login()
+
+Misc.
```
jlewitt1, line 197: this section feels a little out of place since it's really just about runhouse secrets management, maybe we put this in the secrets docs and link out to that directly from the examples where relevant?

mkandler: Yeah that's a good point - it was reoccurring in the examples I was working on but definitely only makes sense for very specific use cases... probably best to include there, or even just keep in specific examples? Maybe something a little more generic than this in the secrets docs if not already covered

docs/installation.rst
mkandler: Good point - maybe just a link to https://skypilot.readthedocs.io/en/latest/reference/cli.html#cluster-cli

docs/installation.rst
mkandler: Agreed, I'll update to clarify

mkandler: Does the updated version seem better?

docs/installation.rst
jlewitt1: In the secrets api docs there's a section starting with: "You can also include a list of secrets in a Runhouse env object", I think this captures that, although I didn't see a great way for us to link to that directly

mkandler: Yeah I think maybe a separate project to clarify and break that up a bit more... could live on the envs and packages page too

mkandler:
@jlewitt1 does this seem good to go?

PR #1038, opened 2024-07-18
Change `env` defaulting for `Module` to be more robust.

rohinb2:


rohinb2:
* **#1038** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1038?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

runhouse/resources/module.py
```diff
@@ -49,6 +49,24 @@
 ]
 
 
+def _env_for_module(passed_in_env=None, system=None):
```
dongreenberg, line 52: Can this be named more clearly?

runhouse/resources/module.py
```diff
@@ -49,6 +49,24 @@
...
+    active_env_servlet_env = obj_store.get_process_env()
+    if active_env_servlet_env:
+        return active_env_servlet_env
+
+    if system and isinstance(system, Cluster):
+        return system.default_env
+
+    potential_default_env = _get_env_from(_default_env_if_on_cluster())
+    if potential_default_env:
+        return potential_default_env
```
dongreenberg, line 65: I don't think this is accurate - if no system is set and the obj_store doesn't have storage, we shouldn't even put it in the default env, just treat it the same as if created outside of the cluster entirely.

runhouse/resources/module.py
```diff
@@ -49,6 +49,24 @@
...
+
+    return Env()
```
dongreenberg, line 67: I'm not sure about this either - if we don't have an env to infer, I think we should just fall back to the cluster constructor's behavior, not set a default or placeholder.

PR #1064, opened 2024-07-25
updated sending logs to den according to the new den endpoint

BelSasha:


BelSasha:
* **#1064** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1064?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

runhouse/rns/rns_client.py
```diff
@@ -664,20 +665,38 @@ def _get_logs(self):
...
 
             post_logs_resp = requests.post(
                 f"{api_server_url}/resource/{cluster_uri}/logs",
                 data=json.dumps(logs_data),
                 headers=self.request_headers(),
             )
-            return post_logs_resp.status_code
+            return post_logs_resp.status_code, prev_end_log_line, new_end_log_line
         except Exception as e:
             logger.error(
```
jlewitt1, line 699: if the request to den fails (not 200), this exception won't actually be thrown right? this is meant to catch the other logic leading up to the request?

runhouse/rns/rns_client.py
```diff
@@ -664,20 +665,38 @@ def _get_logs(self):
         cleaned_log_lines = [ColoredFormatter.format_log(line) for line in log_lines]
         return " ".join(cleaned_log_lines)
 
-    async def send_cluster_logs_to_den(self, cluster_uri: str, api_server_url: str):
+    def _generate_logs_file_name(self):
+        current_timestamp = datetime.datetime.utcnow().strftime("%Y_%m_%d_%H_%M_%S")
+        return f"{current_timestamp}_{S3_LOGS_FILE_NAME}"
```
jlewitt1, line 670: can we actually change to `SERVER_LOGS_FILE_NAME` to be more precise?

runhouse/rns/rns_client.py
```diff
@@ -664,20 +665,38 @@ def _get_logs(self):
         cleaned_log_lines = [ColoredFormatter.format_log(line) for line in log_lines]
         return " ".join(cleaned_log_lines)
 
-    async def send_cluster_logs_to_den(self, cluster_uri: str, api_server_url: str):
+    def _generate_logs_file_name(self):
```
jlewitt1, line 668: im not sure if `rns_client.py` is the best place for all these methods, feels like `http_utils.py` might be more appropriate?

runhouse/rns/rns_client.py
BelSasha: correct :)

runhouse/rns/rns_client.py
BelSasha: originally it was placed in `cluster_servlert.py`, which makes sense cause those functions (send_cluster_logs_to_den, send_status) are executed by a thread in the cluster servelet.

Maybe it is worth moving them back to `cluster_servlert.py`, wdyt? @jlewitt1

jlewitt1: Yea I think that works too

runhouse/rns/rns_client.py
jlewitt1: So small bit but the log message in the exception handler maybe can be more accurate, bc it's not a problem sending the logs to done it's a problem generating the logs before they are sent

PR #1097, opened 2024-08-01
send cluster status to den when running runhouse status cli command

BelSasha:


BelSasha:
* **#1097** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1097?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

dongreenberg:
I'd rather not have this logic duplicated between the ClusterServlet and main.py, and generally we try to have interfaces with ClusterServlet and EnvServlet go through the obj_store, not directly from main.py or HTTP_server. Can we just modify the `ClusterServlet.aperiodic_cluster_checks` method to call a `ClusterServlet.asend_cluster_status(status)` method, and then have `obj_store.astatus` take an optional argument `send_to_den=True`, which calls ClusterServlet.asend_cluster_status(status) before returning it?

runhouse/servers/http/http_client.py
```diff
@@ -267,10 +267,14 @@ def check_server(self):
                 f"but local Runhouse version is ({runhouse.__version__})"
             )
 
-    def status(self, resource_address: str):
+    def status(self, resource_address: str, send_to_den: bool = False):
         """Load the remote cluster's status."""
         # Note: Resource address must be specified in order to construct the cluster subtoken
-        return self.request("status", req_type="get", resource_address=resource_address)
+        return self.request(
+            f"status?send_to_den={send_to_den}",
```
jlewitt1, line 274: why do we need to include this param in the request itself made to den?

runhouse/main.py
```diff
@@ -455,30 +455,44 @@ def status(
...
-        cluster_config = copy.deepcopy(cluster_status.get("cluster_config"))
+        # cluster status type: ResourceStatusData
+        cluster_status, send_status_resp = obj_store.status(send_to_den=True)
+
+        cluster_config = copy.deepcopy(cluster_status.cluster_config)
         current_cluster: Cluster = Cluster.from_config(cluster_config)
-        return _print_status(cluster_status, current_cluster)
 
-    if cluster_name is None:
+    elif cluster_name is None:
```
jlewitt1, line 464: instead of making this an elif, can we make this an if condition first? it feels out of place logically

runhouse/servers/cluster_servlet.py
```diff
@@ -261,38 +261,41 @@ async def aperiodic_cluster_checks(self):
                     break
 
                 logger.debug("Performing cluster checks")
-                status: ResourceStatusData = await self.astatus()
+                if should_send_status_and_logs_to_den:
+                    # status type: ResourceStatusData
+                    status, send_status_to_den_resp = await self.astatus(
```
jlewitt1, line 266: isn't is possible for `send_status_to_den_resp` to be None here if we're not sending to Den? is that being accounted for below?

runhouse/servers/http/http_client.py
BelSasha: It is the status http request that is sent to the cluster, not to den :)

runhouse/servers/cluster_servlet.py
BelSasha: Yes. If `should_send_status_and_logs_to_den` is false, we don't send the status info to den, and the `send_status_to_den_resp` is set to `None`. Later on, in line 277, we check if `should_send_status_and_logs_to_den` is false, and if this is the case, we break ( = exiting the status check loop). Therefore it is ok if the `send_status_to_den_resp` is None in case `should_send_status_and_logs_to_den` is false.

runhouse/main.py
```diff
@@ -452,33 +453,48 @@ def status(
...
+
     # case we are inside the cluster
-    if cluster_or_local != "file":
+    elif cluster_or_local != "file":
+
         # If we are on the cluster load status directly from the object store
-        cluster_status: dict = dict(obj_store.status())
-        cluster_config = copy.deepcopy(cluster_status.get("cluster_config"))
+        # cluster status type: ResourceStatusData
+        cluster_status, send_status_resp = obj_store.status(send_to_den=True)
```
jlewitt1, line 466: i think we can consolidate this logic, and have everything flow thru: 
```
current_cluster.status(
                resource_address=current_cluster.rns_address, send_to_den=True
            )
```

since this will check if we are on the cluster, and call the object store if so anyway

runhouse/servers/cluster_servlet.py
```diff
@@ -441,10 +444,24 @@ async def astatus(self):
...
+        if send_to_den:
+            logger.debug("Sending cluster status to Den")
+            cluster_rns_address = config_cluster.get("name")
+            cluster_uri = rns_client.format_rns_address(cluster_rns_address)
+            api_server_url = config_cluster.get(
+                "api_server_url", rns_client.api_server_url
+            )
+            # returning both status data and the post response because they are both used in the periodic cluster check
+            # loop in the cluster servlet.
+            return status_data, await ClusterServlet.save_status_metrics_to_den(
```
jlewitt1, line 457: its unclear what `await ClusterServlet.save_status_metrics_to_den` returns,  can we first set this to a variable? 
also i think we can move all the logic from inside this if statement into that function

jlewitt1: actually maybe we can decouple these two actions? meaning status only handles returning the status data, and the send to den is separate

runhouse/servers/cluster_servlet.py
```diff
@@ -261,38 +261,41 @@ async def aperiodic_cluster_checks(self):
...
                     )
                 elif status_code != 200:
                     logger.error(
-                        f"{status_code}: Failed to send cluster status to Den: {resp.json()}"
+                        f"{status_code}: Failed to send cluster status to Den: {send_status_to_den_resp.json()}"
                     )
                 else:
                     logger.debug("Successfully sent cluster status to Den.")
+
+                    cluster_rns_address = status.cluster_config.get("name")
```
jlewitt1, line 293: feel like we're using thses lines in a few places,  maybe we use a helper

tests/test_resources/test_clusters/test_on_demand_cluster.py
```diff
@@ -201,6 +201,41 @@ def test_fn_to_docker_container(self, ondemand_aws_cluster):
...
     ####################################################################################################
+
+    @pytest.mark.level("minimal")
+    @pytest.mark.clustertest
+    def test_status_scheduler_basic_flow(self, cluster):
+
+        cluster.save()
+        # the scheduler start running in a delay of 1 min, so the cluster startup will finish properly.
+        # Therefore, the test needs to sleep for a while.
+        time.sleep(60)
```
jlewitt1, line 212: is there an easy way to adjust the time of the scheduler to much shorter, rather than having to have this test wait over a minute to run?

BelSasha: yeah, I'll remove the sleeping time, cause the scheduler does not start running in delay any more. ([1131](https://github.com/run-house/runhouse/pull/1131/files)

runhouse/main.py
```diff
@@ -452,23 +454,23 @@ def status(
...
-        return _print_status(cluster_status, current_cluster)
-
-    if cluster_name is None:
+    if cluster_or_local == "file" and not cluster_name:
         # If running outside the cluster must specify a cluster name
         console.print("Missing argument `cluster_name`.")
         return
 
+    # case we are inside the cluster
+    elif cluster_or_local != "file":
```
jlewitt1, line 463: i think we can delete this entire elif statement? bc the call inside the try / except handles the case where we are inside the cluster

BelSasha: yeah, i have it fixed, will push in a few

BelSasha: yeah but we need to create a cluster instance to call `current_cluster.status()`, which handles the case where we are inside the cluster. In order to create the cluster instance i need its name, and when we are running `runhouse status` inside the cluster we allow the user not to pass the cluster name. 

So this elif clause is still relevant unfortunately.

jlewitt1: can we avoid parsing the file manually tho? if we're on the cluster `rh.here.name` should suffice

runhouse/main.py
```diff
@@ -420,7 +423,11 @@ def status(
     cluster_name: str = typer.Argument(
         None,
         help="Name of cluster to check. If not specified will check the local cluster.",
-    )
+    ),
+    send_to_den: bool = typer.Option(
+        default=False,
+        help="Update den with the loaded status or not",
```
jlewitt1, line 429:
```suggestion
        help="Whether to update Den with the status",
```

runhouse/resources/hardware/cluster.py
```diff
@@ -739,18 +739,45 @@ def connect_server_client(self, force_reconnect=False):
...
         else:
             status = self.call_client_method(
                 "status",
                 restart_server=False,
                 resource_address=resource_address or self.rns_address,
+                send_to_den=send_to_den,
             )
+
+        # don't want to return it as part of the status, it is for internal usage only
+        status = dict(status) if not isinstance(status, dict) else status
```
jlewitt1, line 763: confused about this - we're not returning a normal dict?

runhouse/resources/hardware/cluster.py
```diff
@@ -739,18 +739,45 @@ def connect_server_client(self, force_reconnect=False):
...
+        send_to_den_resp = status.pop("send_to_den_resp")
```
jlewitt1, line 764: why do we need to pop this / include it in the reponse data to begin with? don't we already know its value since it's passed into this function

runhouse/servers/cluster_servlet.py
```diff
@@ -435,10 +415,32 @@ async def astatus(self):
...
+                status=status_data
+            )
+
+            save_in_den_resp = rns_client.save_status_metrics_to_den(
+                status=status_data,
+                cluster_uri=cluster_uri,
+                api_server_url=api_server_url,
+            )
+
+            response_info = {
```
jlewitt1, line 433: do we need to create new objects and store this? what about logging a warning if it fails, otherwise do nothing?

tests/test_resources/test_clusters/test_cluster.py
```diff
@@ -754,48 +764,15 @@ def test_send_status_to_db(self, cluster):
...
-        )
-
     ####################################################################################################
     # Default env tests
     ####################################################################################################
 
     @pytest.mark.level("local")
     @pytest.mark.clustertest
     def test_default_env_in_status(self, cluster):
+        # Status method return two values - the status data, and the response from den, if the status is sent to den.
```
jlewitt1, line 774: are we still returning 2 values? maybe im missing that?

runhouse/resources/hardware/cluster.py
```diff
@@ -739,18 +739,45 @@ def connect_server_client(self, force_reconnect=False):
                 system=self,
             )
 
-    def status(self, resource_address: str = None):
-        """Loads the status of the Runhouse daemon running on the cluster."""
+    def status(self, resource_address: str = None, send_to_den: bool = False):
+        """
+        Loads the status of the Runhouse daemon running on the cluster.
+        :param resource_address: resource's rns address (/<owner_username>/<resource_name>)
```
jlewitt1, line 745: can we use google doc strings in line with the rest of the repo? (same with some of the other methods in `main.py` actually)

runhouse/resources/hardware/cluster.py
BelSasha: `send_to_den_resp` contains info that we want to log, in case the send to den was not successful. See more about that in my comment below. :)

runhouse/servers/cluster_servlet.py
BelSasha: yeah, but since we are sending the status to den from the cluster servlet, the warning log will be logged to the server.log on the cluster. But we want the log to propagate to the user local terminal / notebook / IDE (wherever the user calls `cluster.status()` / `runhouse status`)

runhouse/rns/rns_client.py
```diff
@@ -33,6 +39,7 @@ class ResourceStatusData(BaseModel):
     env_servlet_processes: Dict[str, Dict[str, Any]]
     server_pid: int
     runhouse_version: str
+    send_to_den_resp: Optional[UpdateDenResponse] = None
```
jlewitt1, line 42: this feels a bit out of place to include for the resource status data

BelSasha: yeah, basically I agree. I do wonder if there is a better place to include the response, so it will be propagated back to `cluster.status()` without making `ClusterServlet.status()` to return two values (status itself + send to den response info)

The `send_to_den_resp` is for internal use only, and won't be displayed to users as part of status (only case it will "displayed" is by being logged if necessary)

jlewitt1: oh i see, this seems to be a question of how necessary it is to have that propagate all the way back, I think it really only matters when `send_to_den=True`. 

But that still doesn't solve how to actually return the data from the `atatus call`. maybe we make it a dict instead? meaning `astatus` returns  a dict with the status data + send to den resp, and then within cluster `status` we can then surface the send to den call result (where relevant), but still only return the status data like we do now?

runhouse/rns/rns_client.py
```diff
@@ -21,7 +22,12 @@
     ResourceAccess,
 )
 
-from runhouse.utils import locate_working_dir
+from runhouse.utils import locate_working_dir, sync_function
+
+
+class UpdateDenResponse(BaseModel):
```
jlewitt1, line 28: maybe we just reuse the existing `Response` we already have?

runhouse/rns/rns_client.py
```diff
@@ -630,3 +637,31 @@ def contents(self, name_or_path, full_paths):
         return folder(name=name_or_path, path=folder_url).resources(
             full_paths=full_paths
         )
+
+    async def asave_status_metrics_to_den(
+        self, status: ResourceStatusData, cluster_uri: str, api_server_url: str
+    ):
+        from runhouse.resources.hardware.utils import ResourceServerStatus
+
+        resource_info = dict(status)
```
jlewitt1, line 646: small nit - i think we should use .dict() for pydantic models (https://docs.pydantic.dev/1.10/usage/exporting_models/)

runhouse/rns/rns_client.py
BelSasha: I'm thinking to remove those `UpdateDenResponse` and `ResourceStatusData` at all, since we have validation of the status schema on the den side. wdyt?

jlewitt1: i think `ResourceStatusData` is still helpful on the OSS side to ensure we are constructing it properly

BelSasha: btw - tried to reuse `Response`, but pydentic throws an error that  Response can't be used as a value type in a  model (makes sense).

runhouse/main.py
```diff
@@ -443,32 +451,30 @@ def status(
...
+        )
+        raise typer.Exit(1)
 
-    if cluster_name is None:
-        # If running outside the cluster must specify a cluster name
-        console.print("Missing argument `cluster_name`.")
-        return
+    else:
+        # getting the name from config so we'll get the full rns address.
+        current_cluster: Cluster = Cluster.from_name(name=rh.here.config().get("name"))
```
jlewitt1, line 465: can we do `rh.here.rns_address`?

runhouse/main.py
```diff
@@ -443,32 +451,30 @@ def status(
...
-        current_cluster: Cluster = Cluster.from_name(name=cluster_name)
-        cluster_status: dict = current_cluster.status(
-            resource_address=current_cluster.rns_address
+        cluster_status = current_cluster.status(
+            resource_address=current_cluster.rns_address, send_to_den=send_to_den
+        )
+
+        # when we run runhouse status in the cluster, cluster_status is of type ResourceStatusData.
+        # when we run runhouse status outside the cluster, cluster_status is of type dict.
+        cluster_status = (
```
jlewitt1, line 474: would it be simpler for consistency to always return a dict?

runhouse/servers/cluster_servlet.py
```diff
@@ -434,11 +412,33 @@ async def astatus(self):
...
+                status=status_data,
+                cluster_uri=rns_client.format_rns_address(
+                    self.cluster_config.get("name")
+                ),
+                api_server_url=self.cluster_config.get(
+                    "api_server_url", rns_client.api_server_url
+                ),
+            )
+
+            response_info = {
```
jlewitt1, line 432: what if we just returned the response object from the request? that way we could later reference its status code or message pretty conveniently

runhouse/resources/hardware/cluster.py
```diff
@@ -739,18 +739,37 @@ def connect_server_client(self, force_reconnect=False):
...
-            status = self.call_client_method(
+            status, send_to_den_resp = self.call_client_method(
                 "status",
                 restart_server=False,
                 resource_address=resource_address or self.rns_address,
+                send_to_den=send_to_den,
             )
+
+        if send_to_den:
+            send_to_den_status_code = send_to_den_resp.get("status_code")
```
jlewitt1, line 759: if we return  the resp object here, we can call `resp.status_code`, and then in the warning message also just do `resp.json()`

runhouse/resources/hardware/cluster.py
```diff
@@ -739,18 +739,37 @@ def connect_server_client(self, force_reconnect=False):
...
-        """Loads the status of the Runhouse daemon running on the cluster."""
+    def status(self, resource_address: str = None, send_to_den: bool = False):
+        """Load the status of the Runhouse daemon running on a cluster."""
+
+        # TODO [sb]: add to docs the response schema of the status output, so users will know what to expect.
         # Note: If running outside a local cluster need to include a resource address to construct the cluster subtoken
         # Allow for specifying a resource address explicitly in case the resource has no rns address yet
         if self.on_this_cluster():
-            status = obj_store.status()
+            status, send_to_den_resp = obj_store.status(send_to_den=send_to_den)
```
jlewitt1, line 749:
```suggestion
            status, den_resp = obj_store.status(send_to_den=send_to_den)
```

runhouse/rns/rns_client.py
```diff
@@ -630,3 +631,31 @@ def contents(self, name_or_path, full_paths):
...
+        self, status: ResourceStatusData, cluster_uri: str, api_server_url: str
+    ):
+        from runhouse.resources.hardware.utils import ResourceServerStatus
+
+        resource_info = status.dict()
+        env_servlet_processes = resource_info.pop("env_servlet_processes")
+
+        status_data = {
+            "status": ResourceServerStatus.running,
+            "resource_type": resource_info.get("cluster_config").get("resource_type"),
```
jlewitt1, line 645: pop it from the dict since we're also sending `resource_info` as its own field?

runhouse/servers/http/http_server.py
```diff
@@ -784,9 +784,9 @@ async def get_keys(request: Request, env_name: Optional[str] = None):
     @staticmethod
     @app.get("/status")
```
jlewitt1, line 785:
```suggestion
    @app.get("/status/{send_to_den}")
```

tests/test_resources/test_clusters/test_on_demand_cluster.py
```diff
@@ -201,6 +201,39 @@ def test_fn_to_docker_container(self, ondemand_aws_cluster):
...
     # Status tests
     ####################################################################################################
+
+    @pytest.mark.level("minimal")
+    @pytest.mark.clustertest
+    def test_status_scheduler_basic_flow(self, cluster):
+
+        cluster.save()
+
+        cluster_logs = cluster.run([f"cat {SERVER_LOGFILE_PATH}"], stream_logs=False)[
```
jlewitt1, line 211: just noting we might be able to replace this with a call to the new /logs endpoint, which might be ready soonish

tests/test_resources/test_clusters/test_on_demand_cluster.py
```diff
@@ -201,6 +201,39 @@ def test_fn_to_docker_container(self, ondemand_aws_cluster):
     ####################################################################################################
     # Status tests
     ####################################################################################################
+
+    @pytest.mark.level("minimal")
+    @pytest.mark.clustertest
+    def test_status_scheduler_basic_flow(self, cluster):
+
+        cluster.save()
```
jlewitt1, line 209: don't we save the fixture by default anyways?

tests/test_resources/test_clusters/test_on_demand_cluster.py
```diff
@@ -201,6 +201,39 @@ def test_fn_to_docker_container(self, ondemand_aws_cluster):
...
+        ][1]
+
+        cluster_uri = rh.globals.rns_client.format_rns_address(cluster.rns_address)
+        api_server_url = rh.globals.rns_client.api_server_url
+        assert (
+            f'HTTP Request: POST {api_server_url}/resource/{cluster_uri}/cluster/status "HTTP/1.1 200 OK"'
+            in cluster_logs
+        )
+
+        cluster_uri = rh.globals.rns_client.format_rns_address(cluster.rns_address)
```
jlewitt1, line 222: some of these are duplicated above

tests/test_resources/test_clusters/test_on_demand_cluster.py
BelSasha: added a TODO comment as a reminder for that.

runhouse/main.py
```diff
@@ -390,6 +389,9 @@ def _print_status(status_data: dict, current_cluster: Cluster):
     :param config: cluster's  config
     :return: cluster's  config
     """
+    print("status data is:r")
```
jlewitt1, line 392: why are we printing all this?

runhouse/main.py
```diff
@@ -443,32 +451,27 @@ def status(
...
+        raise typer.Exit(1)
 
-    if cluster_name is None:
-        # If running outside the cluster must specify a cluster name
-        console.print("Missing argument `cluster_name`.")
-        return
+    else:
+        # case we are running inside the cluster. If saved in den (== has rns_address), load cluster from name.
+        # Else, load from config.
+        cluster_name = rh.here.rns_address
```
jlewitt1, line 466: if we're on the cluster why can't we just do: `current_cluster = rh.here`?

runhouse/rns/rns_client.py
```diff
@@ -30,7 +32,7 @@ class ResourceStatusData(BaseModel):
     system_cpu_usage: float
     system_memory_usage: Dict[str, Any]
     system_disk_usage: Dict[str, Any]
-    env_servlet_processes: Dict[str, Dict[str, Any]]
+    env_servlet_processes: Optional[Dict[str, Dict[str, Any]]]
```
jlewitt1, line 35: why is this optional now?

runhouse/servers/cluster_servlet.py
```diff
@@ -258,46 +238,43 @@ async def aperiodic_cluster_checks(self):
...
                 if status_code == 404:
                     logger.info(
                         "Cluster has not yet been saved to Den, cannot update status or logs."
                     )
                 elif status_code != 200:
                     logger.error(
-                        f"{status_code}: Failed to send cluster status to Den: {resp.json()}"
+                        f"Failed to send cluster status to Den: {den_resp.json()}"
                     )
                 else:
```
jlewitt1, line 262: is it not possible for this to be some other error code? what if theres a 500? we would still be treating it as a successful request here

runhouse/servers/cluster_servlet.py
```diff
@@ -258,46 +238,43 @@ async def aperiodic_cluster_checks(self):
...
                         logs_resp_status_code,
                         new_start_log_line,
                         new_end_log_line,
                     ) = await self.send_cluster_logs_to_den(
-                        cluster_uri=cluster_uri,
-                        api_server_url=api_server_url,
+                        cluster_uri=rns_client.format_rns_address(
+                            self.cluster_config.get("name")
+                        ),
+                        api_server_url=self.cluster_config.get(
```
jlewitt1, line 275: why do we need to pass in `cluster_uri` and `api_server_url` here? doesnt the cluster servlet already have these values?

runhouse/resources/hardware/cluster.py
```diff
@@ -727,17 +727,36 @@ def connect_server_client(self, force_reconnect=False):
...
+                logger.info(
+                    "Cluster has not yet been saved to Den, cannot update status or logs."
+                )
+
+            elif send_to_den_status_code != 200:
+                logger.warning(
+                    f"Failed to send cluster status to den: {den_resp.json()}"
+                )
+
+            # TODO [sb]: the above code (error handling) repeat itself. Make it a utils method.
```
jlewitt1, line 758: can remove this note?

runhouse/rns/rns_client.py
```diff
@@ -630,3 +632,33 @@ def contents(self, name_or_path, full_paths):
         return folder(name=name_or_path, path=folder_url).resources(
             full_paths=full_paths
         )
+
+    async def asave_status_metrics_to_den(
```
jlewitt1, line 636: wdyt about putting this in the cluster servelet instead?

runhouse/servers/cluster_servlet.py
BelSasha: yeah, agreed we could use the the cluster config we've previously loaded. 

I pass the `api_server_url` because we'll load it from the cluster config anyway, so we either load if from the config before passing it to `send_cluster_logs_to_den`, or we pass the config we loaded to `send_cluster_logs_to_den` and load the api_server_url there. Since we already loaded the cluster config in `aperiodic_cluster_checks` (which calls to `send_cluster_logs_to_den` I don't see a reason to load the cluster config again in `send_cluster_logs_to_den`). wdyt?

runhouse/rns/rns_client.py
BelSasha: hmmm, yeah at first it made sense to me to put this in the rns_client, but on second thought it is indeed more appropriate the have it living inside the cluster servlet.

runhouse/resources/hardware/cluster.py
```diff
@@ -727,17 +727,34 @@ def connect_server_client(self, force_reconnect=False):
                 system=self,
             )
 
-    def status(self, resource_address: str = None):
-        """Loads the status of the Runhouse daemon running on the cluster."""
+    def status(self, resource_address: str = None, send_to_den: bool = False):
+        """Load the status of the Runhouse daemon running on a cluster."""
+
+        # TODO [sb]: add to docs the response schema of the status output, so users will know what to expect.
```
jlewitt1, line 733: can remove this todo

runhouse/servers/cluster_servlet.py
```diff
@@ -348,7 +352,7 @@ async def _update_autostop(self, status: ResourceStatusData):
                 > 0
                 for resource_name in resource["env_resource_mapping"].keys()
             )
-            for resource in status.env_servlet_processes.values()
+            for resource in status.get("env_servlet_processes").values()
```
jlewitt1, line 355:
```suggestion
            for resource in status.get("env_servlet_processes", {}).values()
```

runhouse/servers/cluster_servlet.py
```diff
@@ -214,27 +217,37 @@ async def aclear_all_references_to_env_servlet_name(self, env_servlet_name: str)
     ##############################################
     # Periodic Cluster Checks APIs
     ##############################################
-    @staticmethod
-    async def save_status_metrics_to_den(
-        status: ResourceStatusData, cluster_uri: str, api_server_url: str
-    ):
+
+    async def asave_status_metrics_to_den(self, status: dict, cluster_uri: str):
```
jlewitt1, line 221: can't we use `self._cluster_uri` instead of passing it in as a param ?

tests/test_resources/test_clusters/test_on_demand_cluster.py
```diff
@@ -289,6 +318,7 @@ def test_logs_surfacing_scheduler_basic_flow(self, cluster):
             headers=headers,
         )
 
+        # TODO [sb]: call cluster's /logs endpoint once introduced.
```
jlewitt1, line 321: actualy not sure we even need this part where we read the cluster logs on the server, shouldnt it be enough that things are flowing prperly if we get a positive response back from Den? (and same for the abovre in `test_status_scheduler_basic_flow`

BelSasha: I think we actually need to get the logs from server.log, to verify that the last server logs were indeed sent to den (we check the content of the get logs request, not just that the request to den succeeded).

tests/test_resources/test_clusters/test_on_demand_cluster.py
```diff
@@ -300,7 +330,7 @@ def test_logs_surfacing_scheduler_basic_flow(self, cluster):
         cluster_logs = "\n".join(cluster_logs)  # make logs list into one string
 
         assert (
-            "Performing cluster checks: potentially sending to Den, surfacing logs to Den or updating autostop."
+            f'HTTP Request: POST {api_server_url}/resource/{cluster_uri}/cluster/status "HTTP/1.1 200 OK"'
```
jlewitt1, line 333: how reliable is this response (ex: if the logger format changes, can we rely on this?) maybe we can confirm this request works more generically

tests/test_resources/test_clusters/test_on_demand_cluster.py
```diff
@@ -201,6 +201,36 @@ def test_fn_to_docker_container(self, ondemand_aws_cluster):
...
+        api_server_url = rh.globals.rns_client.api_server_url
+        headers = rh.globals.rns_client.request_headers()
+
+        # TODO [sb]: call cluster's /logs endpoint once introduced.
+        cluster_logs = cluster.run([f"cat {SERVER_LOGFILE_PATH}"], stream_logs=False)[
+            0
+        ][1]
+
+        assert (
+            f'HTTP Request: POST {api_server_url}/resource/{cluster_uri}/cluster/status "HTTP/1.1 200 OK"'
```
jlewitt1, line 219: similar note as below

runhouse/servers/cluster_servlet.py
```diff
@@ -258,46 +271,38 @@ async def aperiodic_cluster_checks(self):
...
+
                     prev_end_log_line = cluster_config.get("end_log_line", 0)
                     (
                         logs_resp_status_code,
                         new_start_log_line,
                         new_end_log_line,
                     ) = await self.send_cluster_logs_to_den(
-                        cluster_uri=cluster_uri,
-                        api_server_url=api_server_url,
+                        cluster_uri=self._cluster_uri,
```
jlewitt1, line 305: we can reference this within the function instead of sending as a param?

tests/test_resources/test_clusters/test_cluster.py
```diff
@@ -204,7 +203,7 @@ def test_cluster_endpoint(self, cluster):
             headers=rh.globals.rns_client.request_headers(),
         )
         assert r.status_code == 200
-        status_data = r.json()
+        status_data = r.json()[0]
```
jlewitt1, line 206: what changeed here about the data we get in the response?

tests/test_resources/test_clusters/test_on_demand_cluster.py
BelSasha: @jlewitt1 yeah, good point. I updated the test, to make sure that the status was updated in the last minute (DEFAULT_STATUS_CHECK_INTERVAL). If the cluster periodic loop running as it should, the status should be updated every DEFAULT_STATUS_CHECK_INTERVAL minute(s).

tests/test_resources/test_clusters/test_on_demand_cluster.py
BelSasha: updated this test accordingly as well. (see comment below)

runhouse/main.py
```diff
@@ -443,32 +451,22 @@ def status(
...
+        )
+        raise typer.Exit(1)
 
-    if cluster_name is None:
-        # If running outside the cluster must specify a cluster name
-        console.print("Missing argument `cluster_name`.")
-        return
+    else:
+        # we are inside the cluster
+        current_cluster = rh.here
```
jlewitt1, line 465: we already check `rh.here` at the top, can re-use

runhouse/main.py
```diff
@@ -479,6 +477,7 @@ def status(
             "\N{smiling face with horns} Runhouse Daemon is not running... \N{No Entry} \N{Runner}"
         )
         return
+
     return _print_status(cluster_status, current_cluster)
```
jlewitt1, line 481: i know not part of this PR, but can we update the `print_status` docstrings to use google format similar to the rest of the docs? also does it need to return anything? not sure we're doing anything with `status_data`

runhouse/resources/hardware/cluster.py
```diff
@@ -741,17 +741,37 @@ def connect_server_client(self, force_reconnect=False):
...
-            status = self.call_client_method(
+            status_res = self.call_client_method(
                 "status",
                 resource_address=resource_address or self.rns_address,
+                send_to_den=send_to_den,
             )
+
+        # parsing the status response, depending on the send_to_den flag. If send_to_den, both the status and the
+        # response to den will be retrieved.
+        status, den_resp = status_res if len(status_res) == 2 else (status_res, None)
```
jlewitt1, line 760: would prefer if the status calls return types stayed consistent, whether its thru the object store or HTTP call. we can just have den_resp be None in that case - meaning something like: 
`status_res, _ = obj_store.status(send_to_den=send_to_den)`

runhouse/servers/cluster_servlet.py
```diff
@@ -58,6 +58,14 @@ async def __init__(
         if cluster_config.get("resource_subtype", None) == "OnDemandCluster":
             self.autostop_helper = AutostopHelper()
 
+        self._cluster_uri = rns_client.format_rns_address(
```
jlewitt1, line 61: is it possible for the cluster config not to have a name saved here?

runhouse/servers/cluster_servlet.py
```diff
@@ -493,11 +499,21 @@ async def astatus(self):
...
+        status_data = ResourceStatusData(**status_data).dict()
+
+        if send_to_den:
+
+            logger.debug("Sending cluster status to Den")
+            den_resp = self.save_status_metrics_to_den(status=status_data)
+
+            return status_data, den_resp
+
         return status_data
```
jlewitt1, line 513:
```suggestion
        return status_data, None
```

runhouse/servers/cluster_servlet.py
```diff
@@ -258,46 +271,37 @@ async def aperiodic_cluster_checks(self):
...
-                status: ResourceStatusData = await self.astatus()
+                status, den_resp = await self.astatus(
+                    send_to_den=should_send_status_and_logs_to_den
+                )
+
                 if should_update_autostop:
                     logger.debug("Updating autostop")
                     await self._update_autostop(status)
 
                 if not should_send_status_and_logs_to_den:
```
jlewitt1, line 282: if this is `False` why do we even call `astatus` above?

jlewitt1: or is it bc we need some of that status data to update the autostop?

runhouse/servers/cluster_servlet.py
```diff
@@ -258,46 +271,37 @@ async def aperiodic_cluster_checks(self):
...
                     logger.info(
                         "Cluster has not yet been saved to Den, cannot update status or logs."
                     )
                 elif status_code != 200:
                     logger.error(
-                        f"{status_code}: Failed to send cluster status to Den: {resp.json()}"
+                        f"Failed to send cluster status to Den: {den_resp.json()}"
                     )
                 else:
+
```
jlewitt1, line 296:
```suggestion

```

runhouse/servers/cluster_servlet.py
BelSasha: Because `self._update_autostop` uses the status info. The `_update_autostop` function checks if there are running functions at the moment, and updates the autostop accordingly.

tests/test_resources/test_clusters/test_on_demand_cluster.py
```diff
@@ -201,6 +202,37 @@ def test_fn_to_docker_container(self, ondemand_aws_cluster):
     ####################################################################################################
     # Status tests
     ####################################################################################################
+
+    @pytest.mark.level("minimal")
```
jlewitt1, line 206: these tests should prob be in `test_cluster` no? doesn't really need to be limited to ondemand

tests/test_resources/test_clusters/test_on_demand_cluster.py
```diff
@@ -201,6 +202,37 @@ def test_fn_to_docker_container(self, ondemand_aws_cluster):
...
+            headers=headers,
+        )
+
+        assert get_status_data_resp.status_code == 200
+        resp_data = get_status_data_resp.json().get("data")[0]
+        assert resp_data.get("status") == ResourceServerStatus.running
+
+        # check that the status was updated in the DB in the DEFAULT_STATUS_CHECK_INTERVAL minute(s),
+        # cause the status periodic check runs every DEFAULT_STATUS_CHECK_INTERVAL minute(s).
+        status_updated_at = datetime.datetime.fromisoformat(
```
jlewitt1, line 225: this feels more like a den test (that the DB was actually updated), what's important in OSS is that the status was collected as part of the periodic cluster check and actually sent off to den (with a 200).. maybe just confirming in the cluster logs tghat the ping succeeded would work. wdyt?

tests/test_resources/test_clusters/test_on_demand_cluster.py
```diff
@@ -299,13 +331,20 @@ def test_logs_surfacing_scheduler_basic_flow(self, cluster):
         ]  # clean log formatting
         cluster_logs = "\n".join(cluster_logs)  # make logs list into one string
 
+        resp_data = get_logs_data_resp.json().get("data")
+        # check that the status was updated in the DB in the DEFAULT_STATUS_CHECK_INTERVAL minute(s),
+        # cause the status periodic check runs every DEFAULT_STATUS_CHECK_INTERVAL minute(s).
+        status_updated_at = datetime.datetime.fromisoformat(
```
jlewitt1, line 337: simimlar note as above, this shouldn't really have to depend on den to test that logs collection is working.. maybe we just check the config for line numbers to ocnfirm that they've been updated?

tests/test_resources/test_clusters/test_on_demand_cluster.py
BelSasha: yeah, but we are using `cluster.teardown() ` which is an on-demand cluster method. Runhouse Cluster does not have such method. Moreover, seems like `runhouse stop` is not a part of runhouse cli API (not in the docs - is it on purpose?), so not sure if we should use it in the test to this use-case.

jlewitt1: why do we need to do teardown? i see those tests are being skipped anyways maybe we can remove them

tests/test_resources/test_clusters/test_on_demand_cluster.py
BelSasha: Not sure that checking if the line numbers has been updated is the best solution here:

1. There is an edge case where the lines are not updated, because no new logs were generated (if the cluster is just up, and not functions / modules are running)
2. We need to wait at least 60 seconds to test this, since the cluster checks loop is updating the line numbers, and we need to wait that it will be executed at least once during the test... This will add sufficient latency to the test suite.

tests/test_resources/test_clusters/test_on_demand_cluster.py
BelSasha: to test that once we call cluster.teardown() the updated status is sent to den. We could delete it for now since they are skipped, but I think we should add a separate test suite that runs tests related to cluster termination (I think we've previously talked about it). wdyt?

runhouse/main.py
```diff
@@ -240,9 +239,15 @@ def _print_envs_info(
     env_servlet_processes: Dict[str, Dict[str, Any]], current_cluster: Cluster
 ):
     """
-    Prints info about the envs in the current_cluster.
-    Prints the resources in each env, and the CPU and GPU usage of the env (if exists).
+    <<<<<<< HEAD
```
jlewitt1, line 242: rebasing issues?

tests/test_resources/test_clusters/test_on_demand_cluster.py
jlewitt1: you mean confirming that when you call teardown the status gets updated in Den? maybe i'm missing something but i don't see that in this test?

tests/test_resources/test_clusters/test_on_demand_cluster.py
```diff
@@ -299,13 +331,20 @@ def test_logs_surfacing_scheduler_basic_flow(self, cluster):
...
-            "Performing cluster checks: potentially sending to Den, surfacing logs to Den or updating autostop."
-            in cluster_logs
+        assert "Cluster checks have failed" not in cluster_logs
+        assert "Failed to send cluster logs to Den: " not in cluster_logs
+
+        cluster_status = cluster.status()
+        logs_start_line = cluster_status.get("cluster_config").get(
+            "start_log_line", None
         )
+        logs_end_line = cluster_status.get("cluster_config").get("end_log_line", None)
```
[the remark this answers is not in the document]
BelSasha: we could do that, but it this case we'll have to wait 1 min to ensure that the cluster check loop were executed during the test. This will cause quite latency in executing the tests....

In this case i think that calling den is less "harmful". wdyt? @jlewitt1

runhouse/servers/cluster_servlet.py
```diff
@@ -492,11 +505,21 @@ async def astatus(self):
             "server_memory_usage": memory_usage,
             "server_gpu_usage": server_gpu_usage,
         }
-        status_data = ResourceStatusData(**status_data)
-        return status_data
 
-    def status(self):
-        return sync_function(self.astatus)()
+        # converting status_data to ResourceStatusData instance to verify we constructed the status data correctly
+        status_data = ResourceStatusData(**status_data).dict()
```
jlewitt1, line 510: should prob update with the new pydantic syntax

jlewitt1 (approved):

PR #1101, opened 2024-08-02
remove callable param from folder method

jlewitt1:


jlewitt1:
* **#1101** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1101?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

dongreenberg:
I think we should revert this, I really don't want to have Callable params to built-in modules if we can avoid it. It's basically building an assumption of pickling into the http call to the module which we don't want to do.

PR #1104, opened 2024-08-04
start cluster checks when cluster servelet starts

BelSasha:

PR #1103, opened 2024-08-04
save clusters creds only when sharing the cluster

BelSasha:

PR #1105, opened 2024-08-04
save by default clusters with enabled_telemetry_collection

BelSasha:


BelSasha:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1105?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1105** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1105?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1104** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1104?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1103** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1103?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

runhouse/resources/hardware/cluster.py
```diff
@@ -948,6 +953,9 @@ def restart_server(
             logs_level = DEFAULT_LOG_LEVEL
         logger.info(f"Setting cluster log level to: {logs_level}")
 
+        if not enable_telemetry_collection:
+            enable_telemetry_collection = rh.configs.telemetry_collection_enabled()
```
jlewitt1, line 957: confused about this - shouldn't it be disabled here?

jlewitt1: or maybe im confused by the variable names - we're checking here if telemetry is enabled if no argument is explicitly provided? if it's not explicitly provided why not just set the default to True?

runhouse/servers/http/http_server.py
```diff
@@ -124,6 +126,16 @@ async def ainitialize(
             logger.info(f"setting logs level to {log_level}")
         runtime_env = {"conda": conda_env} if conda_env else None
 
+        enable_telemetry_collection = kwargs.get("enable_telemetry_collection", None)
```
jlewitt1, line 129: do we need all this logic if we just set the default to True?

runhouse/servers/http/http_server.py
```diff
@@ -668,7 +680,7 @@ def get_status(request: Request):
     @staticmethod
     def _collect_cluster_stats():
         """Collect cluster metadata and send to Grafana Loki"""
-        if not configs.data_collection_enabled():
+        if not configs.telemetry_collection_enabled():
```
jlewitt1, line 683: shouldn't we be checking this value when doing the periodic cluster checks?

.github/workflows/setup_rh_config/action.yaml
```diff
@@ -23,7 +23,7 @@ runs:
       run: |
         mkdir ~/.rh && touch ~/.rh/config.yaml
         echo "default_folder: /${{ inputs.username }}" > ~/.rh/config.yaml
-        echo "disable_data_collection: true" >> ~/.rh/config.yaml
+        echo "enable_telemetry_collection: True" >> ~/.rh/config.yaml
```
jlewitt1, line 26: shouldn't `true` remain lowercase when running the echo command?

docs/security-and-authentication.rst
```diff
@@ -24,13 +24,13 @@ Removing Collected Data
 If you would like us to remove your collected data, please contact
 the `Runhouse team <https://www.run.house/about>`_ (first name at run.house)
 
-Disabling Data Collection
+Disabling Telemetry Collection
 -----------------------------------
-To disable data collection and error tracking collection, set the environment variable :code:`DISABLE_DATA_COLLECTION`
-to :code:`True`. Alternatively, set :code:`disable_data_collection` to :code:`true` in your
+To disable telemetry collection and error tracking collection, set the environment variable :code:`ENABLE_TELEMETRY_COLLECTION`
+to :code:`False`. Alternatively, set :code:`enable_telemetry_collection` to :code:`False` in your
```
jlewitt1, line 30: should be `false` in a yaml file i believe

runhouse/servers/http/http_server.py
BelSasha: yeah, I even think that we should not create the periodic cluster checks thread if `telemetry_collection_enabled()` is `false`. (if the user want to change `telemetry_collection_enabled` to true, he or she will be required to restart the server -> a "new" cluster servlet instance will be created -> periodic cluster checks thread will be created because ` telemetry_collection_enabled = true` )

wdyt?

runhouse/servers/http/http_server.py
BelSasha: I think we do: it covers the edge case when in the user's runhouse config `enable_telemetry_collection` is set to `true`, but he or she set restarted a specific cluster with `enable_telemetry_collection = false` (or the other way around)

runhouse/resources/hardware/cluster.py
```diff
@@ -967,6 +970,11 @@ def restart_server(
             )
             + " --from-python"
             + f" --log-level {logs_level}"
+            + (
+                " --enable-telemetry-collection"
+                if enable_telemetry_collection
+                else " --no-enable-telemetry-collection"
```
jlewitt1, line 976: we shouldn't have to explicitly set it to not being enabled, if the `--enable-telemetry-collection` flag is provided that should be enough

runhouse/resources/hardware/cluster_factory.py
```diff
@@ -192,7 +194,7 @@ def cluster(
     )
     c.set_connection_defaults(**kwargs)
 
-    if den_auth:
+    if den_auth or rh.configs.telemetry_collection_enabled():
```
jlewitt1, line 197: do we need this here? if den auth is not enabled then we shouldn't be saving regardless

runhouse/main.py
```diff
@@ -836,12 +840,19 @@ def restart(
         help="Minimum log level for logs to be printed",
         callback=lambda value: value.upper(),
     ),
+    enable_telemetry_collection: bool = typer.Option(
```
jlewitt1, line 843: actually it's slightly more of an inconvenience to the user to do it this way, bc this is saying you need to provide the "--enable-telemetry-collection" flag explicitly in order for it to be enabled, but really it should be the opposite (enabled by default unless otherwise explicitly specified) 

so maybe we do: 
```
    parser.add_argument(
        "--disable-telemetry",
        action="store_true",  # if providing --disable-telemetry will turn off all telemetry / data collection
        default=argparse.SUPPRESS,  # If user didn't specify, attribute will not be present (not False)
        help="Whether to collect telemetry data",
    )
```

runhouse/resources/hardware/cluster.py
```diff
@@ -832,6 +833,8 @@ def restart_server(
             restart_ray (bool): Whether to restart Ray. (Default: ``True``)
             env (str or Env, optional): Specified environment to restart the server on. (Default: ``None``)
             restart_proxy (bool): Whether to restart Caddy on the cluster, if configured. (Default: ``False``)
+            logs_level (str): The level of logs that are printed to the server.log file (Default: ``None``)
+            enable_telemetry_collection (bool): whether to collect cluster telemetry or not (Default: ``None``)
```
jlewitt1, line 837:
```suggestion
            enable_telemetry_collection (bool): Whether to collect telemetry data on the cluster (Default: ``True``)
```

runhouse/resources/hardware/cluster.py
BelSasha: yeah but if the user does not provide `enable_telemetry_collection`, we set it by default to `true`. Therefore, we need to check if the user provided `enable_telemetry_collection = false `when calling cluster.restart_server(), and than restart it with no telemetry collection. Or do I miss something here?

runhouse/servers/obj_store.py
```diff
@@ -56,6 +57,7 @@ def get_cluster_servlet(
     create_if_not_exists: bool = False,
     runtime_env: Optional[Dict] = None,
     logs_level: str = DEFAULT_LOG_LEVEL,
+    enable_telemetry_collection: bool = None,
```
jlewitt1, line 60: set to True instead of None?

runhouse/servers/obj_store.py
```diff
@@ -157,6 +162,7 @@ async def ainitialize(
         setup_cluster_servlet: ClusterServletSetupOption = ClusterServletSetupOption.GET_OR_CREATE,
         runtime_env: Optional[Dict] = None,
         logs_level: str = DEFAULT_LOG_LEVEL,
+        enable_telemetry_collection: bool = None,
```
jlewitt1, line 165: also set to true here?

jlewitt1: i feel like im missing part the logic flow - to avoid confusion in the http server we should be setting its final value, and everywhere else it should be passed in as a boolean

runhouse/servers/http/http_server.py
BelSasha: yeah, we are checking it.

btw - it seems that this method is obsolete, looks like we don't use it because it causing latency.

runhouse/servers/http/http_server.py
```diff
@@ -124,6 +126,10 @@ async def ainitialize(
             logger.info(f"setting logs level to {log_level}")
         runtime_env = {"conda": conda_env} if conda_env else None
 
+        disable_telemetry = kwargs.get("disable_telemetry")
+        if disable_telemetry:
+            rh.configs.disable_telemetry_collection()
```
jlewitt1, line 131: can we add a log here like we do if we are updating the log level?

runhouse/rns/defaults.py
```diff
@@ -31,7 +31,7 @@ class Defaults:
         "default_autostop": -1,
         "use_spot": False,
         "use_local_configs": True,
-        "disable_data_collection": False,
+        "enable_telemetry_collection": True,
```
jlewitt1, line 34: should we keep this naming consistent with the server's logic (`disable_telemetry` or something like that)?

jlewitt1 (approved):

PR #1106, opened 2024-08-05
periodic cluster checks loop sends info logs in debug mode

BelSasha:


BelSasha:
* **#1106** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1106?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

dongreenberg:
I wouldn't switch back to requests, we really don't need two http dependencies. Also I think it's a much simpler assumption that we do these async, especially if we reuse the code elsewhere in the cluster servlet. Can we just set the log level for httpx or silence this properly?

PR #1115, opened 2024-08-06
Add general gpu info to cluster status

BelSasha:


runhouse/servers/cluster_servlet.py
```diff
@@ -419,22 +450,34 @@ async def astatus(self):
...
         # Fields: `total`, `used`, `free`, `percent`
         disk_usage = psutil.disk_usage("/")._asdict()
 
+        server_pid: int = get_pid()
+
+        # get general gpu usage
+        server_gpu_usage = (
+            self._get_node_gpu_usage(server_pid)
+            if self.cluster_config.get("has_cuda", False)
+            else {}
```
mkandler, line 467: Small nit, would it be better to return `None` here? Makes the logic a bit easier on the front-end (can check if missing instead of if missing OR missing values)

mkandler (approved):
LGTM - once again, really like the clarification with the new variables!

PR #1128, opened 2024-08-08
create an enable_telemetry flag for clusters

BelSasha:


BelSasha:
* **#1128** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1128?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

docs/security-and-authentication.rst
```diff
@@ -34,3 +34,15 @@ local Runhouse config (:code:`~/.rh/config.yaml`), or in Python:
 
     import runhouse as rh
     rh.configs.disable_data_collection()
+
+
+Disabling Cluster Telemetry Recording
```
BelSasha, line 39: Decided to call it Cluster Telemetry Recording after taking a look at some Ray [documentation](https://docs.ray.io/en/latest/_modules/ray/serve/handle.html#). Don't have strong opinion about that though.

docs/security-and-authentication.rst
```diff
@@ -34,3 +34,15 @@ local Runhouse config (:code:`~/.rh/config.yaml`), or in Python:
 
     import runhouse as rh
     rh.configs.disable_data_collection()
+
+
+Disabling Cluster Observability
+---------------------------------------
+To disable cluster observability, set the environment variable :code:`enable_observability`
+to :code:`False`. Alternatively, set :code:`enable_observability` to :code:`false` in your
+local Runhouse config (:code:`~/.rh/config.yaml`), or in Python:
```
jlewitt1, line 43: maybe we make this clear that doing so will disable telemetry across all subsequent clusters? since this is more of a "global" setting

runhouse/resources/hardware/cluster.py
```diff
@@ -918,6 +924,17 @@ def restart_server(
                 )
                 cluster_cert_path = f"{base_caddy_dir}/{self.cert_config.CERT_NAME}"
 
+        if enable_observability is None:
+            enable_observability = rh.configs.observability_enabled
+
+            # edge case: user disabled cluster observability and then wants to enable it again.
+            if not enable_observability:
+                rh.configs.enable_observability()
```
jlewitt1, line 932: do we want to be enabling or disabling at the config level? we should prob allow the user to control this at the cluster level, rather than globally

runhouse/servers/cluster_servlet.py
```diff
@@ -46,6 +47,10 @@ async def __init__(
         if not cluster_config:
             cluster_config = load_cluster_config_from_file()
 
+        if cluster_config:
```
jlewitt1, line 50: maybe move this a couple lines lower until after we've already set `self.cluster_config`? that way we could also remove this check

runhouse/servers/cluster_servlet.py
```diff
@@ -114,6 +119,13 @@ async def aset_cluster_config_value(self, key: str, value: Any):
 
         return self.cluster_config
 
+    async def asave_cluster_config_locally(self, config: Dict[str, Any]):
+        import os
+
+        with open(os.path.expanduser(CLUSTER_CONFIG_PATH), "w+") as config_file:
+            json.dump(config, config_file)
+            config_file.write("\n")
```
jlewitt1, line 127: small nit but why do we need to add this newline here?

runhouse/servers/http/http_server.py
```diff
@@ -115,15 +117,20 @@ async def ainitialize(
         default_env_name=None,
         conda_env=None,
         from_test: bool = False,
+        disable_observability: bool = False,
+        logs_level: str = DEFAULT_LOG_LEVEL,
```
jlewitt1, line 121: minor point but seems like some of these log changes are kinda separate from the observability changes.. is there another reason why they are included in this PR?

runhouse/servers/cluster_servlet.py
```diff
@@ -114,6 +119,13 @@ async def aset_cluster_config_value(self, key: str, value: Any):
 
         return self.cluster_config
 
+    async def asave_cluster_config_locally(self, config: Dict[str, Any]):
```
jlewitt1, line 122: why do we need this instead of just updating the cluster config stored in the `ClusterServlet`

tests/test_resources/test_clusters/test_cluster.py
```diff
@@ -896,3 +897,42 @@ def test_switch_default_env(self, cluster):
...
+    @pytest.mark.level("local")
+    @pytest.mark.clustertest
+    @pytest.mark.skip(
+        "Stopping and restarting the server mid-test causes some errors, need to fix"
+    )
+    def test_enable_and_disable_observability(self, cluster):
+        import json
+
+        # check that by default cluster telemetry is enabled
+        cluster_config_on_cluster = cluster.run(
```
jlewitt1, line 910: why not use `cluster.config_for_rns`?

tests/test_resources/test_clusters/test_cluster.py
```diff
@@ -896,3 +897,42 @@ def test_switch_default_env(self, cluster):
         # set it back
         cluster.default_env = test_env
         cluster.delete(new_env.name)
+
+    @pytest.mark.level("local")
+    @pytest.mark.clustertest
+    @pytest.mark.skip(
+        "Stopping and restarting the server mid-test causes some errors, need to fix"
```
jlewitt1, line 904: what's the issue with restarting the server?

runhouse/servers/cluster_servlet.py
BelSasha: Yeah, since it is only for internal usage, there is not really a necessity to update the cluster config locally.

Having said that, I do wonder - don't we want the user to be aware that the observability is disabled or enabled on a specific cluster? If such field will be saved on the cluster_config, it could be displayed on cluster metadata in den. 

Another way to display it will be by calling cluster.status(). The cluster_config retrieved by the status method is the cluster_config saved in the cluster_servlet, so we'll be able to know if the cluster telemetry is on of not. We could add Matt to the discussion on how to display it in the UI.

wdyt?

tests/test_resources/test_clusters/test_cluster.py
BelSasha: If we won't save the `enable_observability`on the local cluster_config on the cluster, this test will be modified. (Probably will use cluster.status() to get the value of `enable_observability`)

docs/security-and-authentication.rst
BelSasha: Add more information about what is a cluster observability, before explaining how to disable it.

docs/security-and-authentication.rst
```diff
@@ -34,3 +34,15 @@ local Runhouse config (:code:`~/.rh/config.yaml`), or in Python:
 
     import runhouse as rh
     rh.configs.disable_data_collection()
+
+
+Disabling Clusters Observability
```
BelSasha, line 39: Disabling Clusters Observability -> Cluster Observability

.github/workflows/setup_rh_config/action.yaml
```diff
@@ -27,3 +27,4 @@ runs:
         echo "token: ${{ inputs.token }}" >> ~/.rh/config.yaml
         echo "username: ${{ inputs.username }}" >> ~/.rh/config.yaml
         echo "api_server_url: ${{ inputs.api_server_url }}" >> ~/.rh/config.yaml
+        echo "enable_observability: true" >> ~/.rh/config.yaml
```
BelSasha, line 30: Make enable_observability -> disable_observability everywhere in the code, for constancy.

runhouse/resources/hardware/cluster.py
```diff
@@ -837,10 +840,16 @@ def restart_server(
             restart_ray (bool): Whether to restart Ray. (Default: ``True``)
             env (str or Env, optional): Specified environment to restart the server on. (Default: ``None``)
             restart_proxy (bool): Whether to restart Caddy on the cluster, if configured. (Default: ``False``)
+            logs_level (str): cluster's log level: INFO, DEBUG, WARNING, ERROR, EXCEPTION
+            disable_observability (bool): enable or disable cluster observability. If true, the cluster will NOT be observed by runhouse, and cluster telemetry info won't be collected. If false, the cluster will be observed by runhouse, and cluster telemetry info will be collected. If None, the cluster observability will be set to the default cluster_observability, set in the local runhosue config.yaml (Default: ``None``). More info about cluster observability can be found in our `Security and Authentication <https://www.run.house/docs/security-and-authentication#cluster-observability>`__ documentation.
```
jlewitt1, line 844:
```suggestion
            disable_observability (bool): Whether to enable cluster observability. If ``True``, no telemetry data will be collected on the cluster. If not specified, will first look for the `disable_observability` value set in the local Runhouse config (`~/.rh/config.yaml`).  If not specified, observability on the cluster will be enabled by default. More info about cluster observability can be found in our `Security and Authentication <https://www.run.house/docs/security-and-authentication#cluster-observability>`__ documentation.
```

runhouse/resources/hardware/cluster.py
```diff
@@ -837,10 +840,16 @@ def restart_server(
...
 
-        Example:
+        Examples:
```
jlewitt1, line 846: i don't think we need this section, seems self explanatory

runhouse/resources/hardware/cluster.py
```diff
@@ -837,10 +840,16 @@ def restart_server(
             restart_ray (bool): Whether to restart Ray. (Default: ``True``)
             env (str or Env, optional): Specified environment to restart the server on. (Default: ``None``)
             restart_proxy (bool): Whether to restart Caddy on the cluster, if configured. (Default: ``False``)
+            logs_level (str): cluster's log level: INFO, DEBUG, WARNING, ERROR, EXCEPTION
```
jlewitt1, line 843:
```suggestion
            logs_level (str): Log level to use for the cluster logger. Options include: ``INFO``, ``DEBUG``, ``WARNING``, ``ERROR``, ``EXCEPTION``
```

docs/security-and-authentication.rst
```diff
@@ -34,3 +34,39 @@ local Runhouse config (:code:`~/.rh/config.yaml`), or in Python:
...
     import runhouse as rh
     rh.configs.disable_data_collection()
+
+
+Cluster Observability
+---------------------------------------
+Runhouse collects various telemetry data by default on clusters. This data will be used to provide better observability
+into logs, traces, and metrics associated with clusters. We will not sell data or buy any observability data collected.
+
+To disable observability for subsequent clusters, set the environment variable :code:`disable_observability`
```
jlewitt1, line 44:
```suggestion
To disable observability globally for all clusters, set the environment variable :code:`disable_observability`
```

docs/security-and-authentication.rst
```diff
@@ -34,3 +34,39 @@ local Runhouse config (:code:`~/.rh/config.yaml`), or in Python:
...
+
+.. code-block:: python
+
+    import runhouse as rh
+    my_cluster = rh.cluster(name="my-gpu-cluster", instance="CPU:2+").up_if_not()
+    my_cluster.restart_server(disable_observability=True)  # for disabling observability on the specific cluster
+
+To enable cluster observability, simply run :code:`my_cluster.restart_server(disable_observability=True)`.
+
+Alternatively, you may ssh to the cluster, and restart the server using `Runhouse CLI command <https://www.run.house/docs/api/cli#runhouse.main.restart>`__ :
```
jlewitt1, line 65:
```suggestion
Alternatively, you may SSH onto the cluster, and restart the Runhouse Server Daemon using the `restart command <https://www.run.house/docs/api/cli#runhouse.main.restart>`__ :
```

runhouse/resources/hardware/cluster.py
```diff
@@ -922,6 +931,13 @@ def restart_server(
         # Update the cluster config on the cluster
         self.save_config_to_cluster()
 
+        if disable_observability is None:
+            disable_observability = not rh.configs.observability_enabled
+
+            # # edge case: user disabled cluster observability and then wants to enable it again.
```
jlewitt1, line 937: can we remove this?

runhouse/rns/defaults.py
```diff
@@ -288,3 +288,15 @@ def data_collection_enabled(self) -> bool:
             return False
 
         return True
+
+    @property
+    def observability_enabled(self):
```
jlewitt1, line 293: following up on note above, i think we want something like this at the cluster level too, since this represents a global configuration

runhouse/servers/http/http_server.py
```diff
@@ -1148,12 +1167,16 @@ async def main():
         # a local `runhouse start`
         cluster_config["server_connection_type"] = "tls" if use_https else "none"
 
+    disable_observability = getattr(parse_args, "disable_observability", False)
```
jlewitt1, line 1170: looks like we're already setting the value above?

tests/test_resources/test_clusters/test_cluster.py
```diff
@@ -1009,3 +1009,26 @@ def test_switch_default_env(self, cluster):
         # set it back
         cluster.default_env = test_env
         cluster.delete(new_env.name)
+
+    @pytest.mark.level("local")
+    @pytest.mark.clustertest
+    # @pytest.mark.skip(
+    #     "Stopping and restarting the server mid-test causes some errors, need to fix"
+    # )
+    def test_enable_and_disable_observability(self, cluster):
```
jlewitt1, line 1018: i feel like the more thorough test would be going thru the different ways observability can be set (ex: in the config.yaml, in the restart command, or initially in the constructor), and making sure we extract the right value. 

if the value is set in the cluster servlet properly not sure we need to be stopping and restarting the server anyways

docs/security-and-authentication.rst
```diff
@@ -34,3 +34,50 @@ local Runhouse config (:code:`~/.rh/config.yaml`), or in Python:
...
+local Runhouse config (:code:`~/.rh/config.yaml`), or in Python:
+
+.. code-block:: python
+
+    import runhouse as rh
+    rh.configs.disable_observability()
+
+Please note that the config.yaml and the disable_observability environment variable are **global settings**, which apply to every new cluster you will create using Runhouse.
+
+To disable observability for a specific cluster, you may restart the cluster Runhouse server, using `Runhouse Cluster API <https://www.run.house/docs/api/python/cluster#runhouse.Cluster.restart_server>`__.
```
jlewitt1, line 55: a little confusing - could say "restart the Runhouse API server on the cluster"?

docs/security-and-authentication.rst
```diff
@@ -34,3 +34,50 @@ local Runhouse config (:code:`~/.rh/config.yaml`), or in Python:
...
+
+.. code-block:: python
+
+    import runhouse as rh
+    my_cluster = rh.cluster(name="my-gpu-cluster", instance="CPU:2+").up_if_not()
+    my_cluster.restart_server(disable_observability=True)  # for disabling observability on the specific cluster
+
+To enable cluster observability, simply run :code:`my_cluster.restart_server(disable_observability=True)`.
+
+Alternatively, you may ssh onto the cluster, and restart Runhouse Server Daemon using the `restart command <https://www.run.house/docs/api/cli#runhouse.main.restart>`__:
```
jlewitt1, line 65:
```suggestion
Alternatively, you may ssh onto the cluster, and restart the Runhouse server daemon using the `restart command <https://www.run.house/docs/api/cli#runhouse.main.restart>`__:
```

docs/security-and-authentication.rst
```diff
@@ -34,3 +34,50 @@ local Runhouse config (:code:`~/.rh/config.yaml`), or in Python:
...
+
+Please note that the config.yaml and the disable_observability environment variable are **global settings**, which apply to every new cluster you will create using Runhouse.
+
+To disable observability for a specific cluster, you may restart the cluster Runhouse server, using `Runhouse Cluster API <https://www.run.house/docs/api/python/cluster#runhouse.Cluster.restart_server>`__.
+
+.. code-block:: python
+
+    import runhouse as rh
+    my_cluster = rh.cluster(name="my-gpu-cluster", instance="CPU:2+").up_if_not()
+    my_cluster.restart_server(disable_observability=True)  # for disabling observability on the specific cluster
```
jlewitt1, line 61: not sure if this is the usage path we want to highlight for disabling observability, bc in this case its entirely possible that the cluster will be started with it enabled

docs/security-and-authentication.rst
```diff
@@ -34,3 +34,50 @@ local Runhouse config (:code:`~/.rh/config.yaml`), or in Python:
...
+Alternatively, you may ssh onto the cluster, and restart Runhouse Server Daemon using the `restart command <https://www.run.house/docs/api/cli#runhouse.main.restart>`__:
+
+.. code-block:: bash
+
+   $ runhouse ssh my-cluster
+   $ runhouse restart --disable-observability
+
+To enable cluster observability, simply run :code:`runhouse restart` or :code:`runhouse restart --no-disable-observability`.
+
+Finally, you may create a brand new Runhouse cluster, and set its :code:`disable_observability` to :code:`True` or :code:`False`, according on your needs.
```
jlewitt1, line 74:
```suggestion
Finally, you may create a brand new Runhouse cluster, and set :code:`disable_observability` to :code:`True` if you would like to launch the cluster without any observability.
```

docs/security-and-authentication.rst
```diff
@@ -34,3 +34,50 @@ local Runhouse config (:code:`~/.rh/config.yaml`), or in Python:
...
+
+.. code-block:: python
+
+    import runhouse as rh
+    my_cluster = rh.cluster(name="my-gpu-cluster", instance="CPU:2+", disable_observability=True).up_if_not()
+
+    # ...
```
jlewitt1, line 81: we can remove these comments

docs/security-and-authentication.rst
BelSasha: how about "restart the Runhouse server on the cluster" or "restart the Runhouse server daemon on the cluster"?

runhouse/servers/cluster_servlet.py
```diff
@@ -214,6 +218,26 @@ async def aclear_all_references_to_env_servlet_name(self, env_servlet_name: str)
     ##############################################
     # Periodic Cluster Checks APIs
     ##############################################
+
+    def _observability_disabled(self) -> bool:
+        """If disable_observability value was changed on the cluster level, it was changed on the local
```
jlewitt1, line 223: is there any reason to even look at the value in the config.yaml on the cluster? since when we restart the runhouse server we'll be saving whatever value was passed into the `restart` command directly into the cluster servlet?

runhouse/resources/hardware/cluster.py
```diff
@@ -115,6 +117,11 @@ def __init__(
         if self._default_env and not self._default_env.name:
             self._default_env.name = _unnamed_default_env_name(self.name)
 
+        disable_observability = kwargs.get(
```
jlewitt1, line 120: does this actually flow thru? i think we may need to add this to the `alt_options`?

runhouse/resources/hardware/cluster.py
```diff
@@ -190,6 +197,21 @@ def default_env(self, env):
                 "Run `cluster.restart_server()` to restart the Runhouse server on the new default env."
             )
 
+    @property
+    def observability_disabled(self) -> bool:
+        return self._disable_observability
+
+    @observability_disabled.setter
```
jlewitt1, line 204: shouldn't these be a property instead of a setter? meaning when we call `self.observability_disabled`, thats when we compare the value passed in (self._observablity_disabled) to the value stored in the config

runhouse/resources/hardware/cluster.py
```diff
@@ -339,6 +361,8 @@ def config(self, condensed=True):
             config["ssl_certfile"] = self.cert_config.cert_path
             config["ssl_keyfile"] = self.cert_config.key_path
 
+        config["disable_observability"] = self._disable_observability
```
jlewitt1, line 364:
```suggestion
        config["disable_observability"] = self.disable_observability
```

runhouse/resources/hardware/cluster.py
```diff
@@ -919,6 +948,11 @@ def restart_server(
                 )
                 cluster_cert_path = f"{base_caddy_dir}/{self.cert_config.CERT_NAME}"
 
+        if disable_observability is None:
```
jlewitt1, line 951: oh i see, so here we would actually need the setter to update the property? so i guess we really should have both

BelSasha: Yeah, but maybe we could just do something like `self.disable_observability = <DEFAULT disable_observability value>` instead of having the setter? Because the setting the disable_observability for true to false complicates our managing of this (prob will need to restart the server..)

jlewitt1:
@BelSasha the next step here is simplifying the logic to only handle this at the global level

runhouse/rns/defaults.py
```diff
@@ -31,7 +31,7 @@ class Defaults:
         "default_autostop": -1,
         "use_spot": False,
         "use_local_configs": True,
-        "disable_data_collection": False,
```
jlewitt1, line 34: we should still keep the data collection param, that's a separate type of data collection (ex: sentry)

runhouse/servers/cluster_servlet.py
```diff
@@ -47,9 +47,12 @@ async def __init__(
         self.cluster_config: Optional[Dict[str, Any]] = (
             cluster_config if cluster_config else {}
         )
+
+        self.cluster_config["disable_observability"] = not configs.observability_enabled
```
jlewitt1, line 51: why do we need to set it here? shouldn't it already be in the config?

runhouse/servers/cluster_servlet.py
```diff
@@ -214,6 +217,14 @@ async def aclear_all_references_to_env_servlet_name(self, env_servlet_name: str)
     ##############################################
     # Periodic Cluster Checks APIs
     ##############################################
+
+    def _observability_disabled(self) -> bool:
```
jlewitt1, line 221: the function name and response seem to be mismatched - if we're checking that observability is disabled shouldn't it just return the value of `disable_observability` in the config?

runhouse/rns/defaults.py
```diff
@@ -288,3 +288,15 @@ def data_collection_enabled(self) -> bool:
             return False
 
         return True
+
+    @property
+    def observability_enabled(self):
+        return not self.get("disable_observability", False)
+
+    def disable_observability(self):
```
jlewitt1, line 296: i think these should be in `rns_client`, like we have for other config values. which can check both env vars andd the config file for values

runhouse/servers/cluster_servlet.py
jlewitt1: also why do we need this function here as opposed to reusing the existing one defined in `rns_client`?

runhouse/rns/defaults.py
BelSasha: but don't we want to to be similar to `Disabling Data Collection`? 
image

if we move those methods to rns_client:
a. the user will not be able to change disable_observability easily via python (he'll need to change the config.yaml or the env var)
b. as mentioned before, it will not match the API of changing the data_collection flag.

runhouse/resources/hardware/cluster.py
```diff
@@ -862,6 +865,13 @@ def status(self, resource_address: str = None, send_to_den: bool = False):
             elif den_resp_status_code != 200:
                 logger.warning("Failed to send cluster status to Den")
 
+        if not configs.observability_enabled and status.get("env_servlet_processes"):
+            logger.warning(
+                "Observability not currently enabled on the cluster. Metrics are stale and will "
```
BelSasha, line 870: @jlewitt1 Maybe "Observability currently not enabled on the cluster" or "Currently, observability is not enabled on the cluster"?

runhouse/resources/hardware/cluster.py
```diff
@@ -1615,12 +1625,7 @@ def run_python(
         """
         cmd_prefix = "python3 -c"
         command_str = "; ".join(commands)
-        command_str_repr = (
```
jlewitt1, line 1618: can revert this too now?

jlewitt1 (approved):

PR #1096, opened 2024-08-01
add option to skip loading resources from den

jlewitt1:

PR #1135, opened 2024-08-12
reloading ondemand cluster from sky db + cluster config

jlewitt1:


jlewitt1:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1135?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1135** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1135?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1096** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1096?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

runhouse/resources/hardware/cluster.py
```diff
@@ -180,19 +181,65 @@ def from_name(
...
+            raise ValueError(f"Cluster {name} not found locally")
+
+        cluster_ip = sky_data["handle"].head_ip
+        if not cluster_ip:
+            raise ValueError(
+                f"Failed to load cluster {name} locally, IP address not found"
+            )
+
+        try:
+            resp = requests.get(
```
carolineechen, line 226: do you know when this request succeeds/fails?

just to clarify expected behavior - wondering if this request fails but a sky cluster is detected whether or not we should go ahead and create it anyways or if it should fail (or something in between, maybe checking that runhouse server is up?). currently in main we have this branch where if from_name fails but the cluster is detected locally it's loosely reconstructed (set properties like provider/instance type may be lost, but I'm working on adding observed properties to the config now).

https://github.com/run-house/runhouse/pull/1093/files#diff-e3919e7abe1f5cd55b08e55638e03b6935258ed6dd0c50afc4f419cb2a59b554R467

jlewitt1: oh nice, actually this logic should prob be called within that block that you recently added for checking the state. 

i set a timeout on the request, but it can fail for a number of reasons (daemon is down, port isn't open, etc.). I feel like us surfacing a more relevant exception is more appropriate than re-upping the cluster entirely in this case

carolineechen: I think I tried upping a local only cluster and the request timed out (but the runhouse server was up) so I wasn't sure why that was happening. so if that's the case then maybe it's too strict of a criteria or we should check another way? but yea agreed we shouldn't create/re-up clusters if theyre down.

PR #1155, opened 2024-08-15
remove file and blob from docs

jlewitt1:


jlewitt1:
* **#1155** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1155?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

docs/api/python.rst
```diff
@@ -51,7 +51,7 @@ and rich debugging and accessibility interfaces built-in.
 
 Data Abstractions
 ------------------------------------
-The Folder, Blob, and File APIs provide a simple interface for storing, recalling, and moving data between
+The Folder APIs provide a simple interface for storing, recalling, and moving data between
```
carolineechen, line 54: maybe we could update this blurb as well? reads a little generic since we were trying to address all the different data abstractions initially but now we only need to describe folder. and we should remove the last sentence here too

docs/api/python.rst
```diff
@@ -51,26 +51,31 @@ and rich debugging and accessibility interfaces built-in.
...
+your laptop.
 
-.. toctree::
-   :maxdepth: 1
+.. code-block:: python
+
+    import runhouse as rh
+
+    # Load a folder which points to a particular directory on S3
+    s3_folder = rh.folder(path"/main-bucket/my-folder", system="s3")
```
jlewitt1, line 73: @carolineechen what do you think of this example? we already have tutorials for functions clusters and modules, figured this would be the most useful way of using a folder for now (though we may end up totally deprecating it)

carolineechen: I think the example's fine, but we haven't included examples here / have something similar in the api tutorials so I think we can leave it out.

carolineechen (approved):
let's remove the folder example from python.rst, otherwise looks good!

PR #1158, opened 2024-08-15
send status to den logging msgs cleanup

BelSasha:


runhouse/servers/telemetry/utils.py
```diff
@@ -0,0 +1,17 @@
+from runhouse.logger import logger
+
+
+def send_status_to_den_succeeded(send_to_den_resp: dict):
```
jlewitt1, line 4: im not sure we want a helper for this, the messages should probably be slightly different anyways bc they are being used in totally different contexts

rohinb2: Agree w/ @jlewitt1 , this is something we can just do eveyr time

rohinb2 (changes requested):

PR #1163, opened 2024-08-16
Add Airflow example

py-rh:
This is technically not runnable Python but I think it makes sense to show callables + airflow code and import the model class and leave that out?


py-rh:
https://www.run.house/examples/airflow-runhouse-model-training-example-pytorch-mnist
For easier viewing

examples/airflow-torch-training/airflow_example_torch_train.py
```diff
@@ -1,17 +1,117 @@
...
+    # cluster.save() ## Use if you have a Runhouse Den account to save and monitor the resource.
+
+
+# We will send the function to download data to the remote cluster and then invoke it to download the data to the remote machine. You can imagine that this is a data access or pre-processing step after which data is prepared.
+def access_data_callable(**kwargs):
+    logger.info("Step 2: Access data")
+    env = rh.env(name="test_env", reqs=["torch", "torchvision"])
+
+    cluster = rh.cluster(name="a10g-cluster").up_if_not()
+    remote_download = rh.function(DownloadData).to(cluster, env=env)
```
dongreenberg, line 73: Python convention is for only classes to be uppercased this way, this should be `download_data` both here and the TorchBasicExample

examples/airflow-torch-training/airflow_example_torch_train.py
```diff
@@ -1,17 +1,117 @@
...
+def train_model_callable(**kwargs):
+    cluster = rh.cluster(name="a10g-cluster").up_if_not()
+
+    env = rh.env(name="test_env", reqs=["torch", "torchvision"])
+
+    remote_torch_example = rh.module(SimpleTrainer).to(
+        cluster, env=env, name="torch-basic-training"
+    )
+
+    model = remote_torch_example()
```
dongreenberg, line 89: nit: Can we call this `remote_trainer` or something like that to be careful to distinguish it from the model proper (which lives inside the trainer)?

dongreenberg (approved):

py-rh:
@carolineechen - @dongreenberg asked me to ask you to give it a quick readthrough before we merged.

PR #1178, opened 2024-08-20
logger + log level updates

jlewitt1:
Set log level on cluster according to local `RH_LOG_LEVEL` env var. If not set use the default


docs/debugging-logging.rst
```diff
@@ -21,12 +21,16 @@ Alternatively, to see logs on your local machine while running a remote function
...
+----------
+You can set the log level to control the verbosity of the Runhouse logs. You can adjust the log level by
+setting the environment variable ``RH_LOG_LEVEL`` to your desired level.
 
 Debugging
 ~~~~~~~~~
 
 For general debugging that doesn't occur within remote function calls, you can add
-``import pdb; pdb.set_trace()`` whereever you want to set your debugging session.
+``import pdb; pdb.set_trace()`` wherever you want to set your debugging session.
```
rohinb2, line 33: should replace this with just `breakpoint()`

runhouse/logger.py
```diff
@@ -5,21 +5,21 @@
 def get_logger(name: str = __name__):
     logger = logging.getLogger(name)
 
+    # Clear any existing handlers - avoid duplicate logs and maintain consistent log setup across modules
+    logger.handlers.clear()
```
rohinb2, line 9: wait won't this clear each time each file calls `get_logger`? so when each file is imported and runs... we'll be running tons of these clear and re-adds

tests/fixtures/docker_cluster_fixtures.py
```diff
@@ -379,6 +379,7 @@ def docker_cluster_pk_ssh(request, test_org_rns_folder):
         ],
         working_dir=None,
         name="default_env",
+        env_vars={"RH_LOG_LEVEL": os.getenv("RH_LOG_LEVEL") or TESTING_LOG_LEVEL},
```
rohinb2, line 382: oh if you set this locally will it propagate server side?

jlewitt1: yea as part of the restart flow / `env.to(cluster)`

runhouse/logger.py
jlewitt1: yea good call we should kill that

rohinb2 (approved):
solid

rohinb2:
### Merge activity

* **Aug 27, 5:50 PM EDT**: @rohinb2 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/1178).

PR #1186, opened 2024-08-22
verify that cluster config is saved properly on the cluster

BelSasha:


BelSasha:
* **#1186** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1186?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

runhouse/resources/hardware/cluster.py
```diff
@@ -216,13 +216,22 @@ def save_config_to_cluster(
...
         )
 
+        for output in save_config_outputs:
+            for res in output:
+                # if run_res[1] or run_res[2] are not empty strings, that means we got an error while saving the
+                # cluster config
+                if res[1] or res[2]:
+                    logger.error("Failed to save cluster config locally")
+                    logger.error(res[1] + res[2])
+                    raise ValueError()
```
BelSasha, line 233: Kinda on the fence here about raising an exception. On the one hand, if some error occurs while saving the config, we want to kill the cluster setup, cause the ClustetServlet setup depends on the local config saved on the cluster. One the other hand, this check is for internal Runhouse usage and debuggability, not sure we need to surface this error to the user by raising an exception.

jlewitt1: i would lean towards the latter and not raising an exception in this case as you noted, and keeping the behavior for now as is

PR #1191, opened 2024-08-23
Add VPC mention

py-rh:
Add a brief description that you can launch within VPCs on the Clusters page.


carolineechen (approved):
just a small format nit

PR #1197, opened 2024-08-25
modify cluster status checks tests

BelSasha:


tests/test_servers/test_servlet.py
```diff
@@ -85,3 +86,111 @@ def test_get_obj_does_not_exist(self, test_servlet):
...
+            "acheck_cluster_status_and_logs",
+            interval_size=60,
+            send_to_den=False,
+        )
+
+        assert logs_resp_status_code is None
+        assert new_start_log_line is None
+        assert new_end_log_line is None
+
+        os.remove(SERVER_LOGFILE)
```
BelSasha, line 196: os.remove -> path.unlink

tests/test_servers/test_servlet.py
```diff
@@ -85,3 +86,111 @@ def test_get_obj_does_not_exist(self, test_servlet):
...
+        from runhouse.rns.rns_client import ResourceStatusData
+
+        expected_keys = list(ResourceStatusData.__fields__.keys())
+        for k in status.keys():
+            assert k in expected_keys
+        assert status_code_cluster_status is None
+
+    @pytest.mark.level("local")
+    def test_cluster_checks_thread_logic_logs(self, cluster):
+        from runhouse.servers.obj_store import ObjStore
```
BelSasha, line 146: put the test in a try-finally block, so the SERVER_LOGFILE will be removed

tests/test_servers/test_servlet.py
```diff
@@ -85,3 +86,111 @@ def test_get_obj_does_not_exist(self, test_servlet):
...
+
+        assert status
+        assert isinstance(status, dict)
+        assert isinstance(logs_resp_status_code, int)
+        assert isinstance(new_start_log_line, int)
+        assert isinstance(new_end_log_line, int)
+
+        (
+            _,
+            logs_resp_status_code,
```
BelSasha, line 181: Just check that logs_resp_status_code is 200

tests/test_resources/test_clusters/test_on_demand_cluster.py
```diff
@@ -201,97 +196,13 @@ def test_fn_to_docker_container(self, ondemand_aws_cluster):
...
-        cluster_logs = "\n".join(cluster_logs)  # make logs list into one string
-
-        assert "Cluster checks have failed" not in cluster_logs
-        assert "Failed to send cluster status to Den: " not in cluster_logs
-
     @pytest.mark.level("minimal")
-    @pytest.mark.skip("Test requires terminating the cluster")
-    def test_set_status_after_teardown(self, cluster):
-
+    def test_set_status_after_teardown(self, cluster, mocker):
```
BelSasha, line 200: add call to den to verify that status is equal to terminated.

runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -560,6 +560,9 @@ def teardown(self):
         Example:
             >>> rh.ondemand_cluster("rh-cpu").teardown()
         """
+
+        status_code_den_request = None
```
BelSasha, line 564: Remove, redundent

runhouse/servers/cluster_servlet.py
```diff
@@ -67,10 +68,10 @@ async def __init__(
         )
 
         logger.info("Creating periodic_cluster_checks thread.")
```
jlewitt1, line 70:
```suggestion
        logger.debug("Creating periodic_cluster_checks thread.")
```

runhouse/servers/cluster_servlet.py
```diff
@@ -2,6 +2,7 @@
 import copy
 import datetime
 import json
+import os.path
```
jlewitt1, line 5: dumb nit - can we use `Path(file_path).exists()` instead - just for consistency?

tests/test_servers/test_servlet.py
```diff
@@ -85,3 +85,61 @@ def test_get_obj_does_not_exist(self, test_servlet):
         assert resp.output_type == "exception"
         error = deserialize_data(resp.data["error"], "pickle")
         assert isinstance(error, KeyError)
+
+    @pytest.mark.level("local")
+    def test_cluster_checks_thread_is_running(self, test_cluster_servlet):
+
+        resp_status_checks_thread = ObjStore.call_actor_method(
```
jlewitt1, line 92: is this obj store method only used for testing purposes? is there a way we can check it's running without it, or alternatively if it's not running would we easily know that in the other tests and having this test is not necessary?

tests/test_servers/test_servlet.py
```diff
@@ -85,3 +85,61 @@ def test_get_obj_does_not_exist(self, test_servlet):
...
+            "_cluster_periodic_thread_alive",
+        )
+
+        assert resp_status_checks_thread
+
+    @pytest.mark.level("local")
+    def test_cluster_checks_thread_logic_status(self, cluster, test_cluster_servlet):
+        from runhouse.servers.obj_store import ObjStore
+
+        cluster.save()
```
jlewitt1, line 103: is the cluster not already being saved?

tests/test_servers/test_servlet.py
```diff
@@ -85,3 +85,61 @@ def test_get_obj_does_not_exist(self, test_servlet):
...
+
+        assert den_status_resp is None
+
+    @pytest.mark.level("local")
+    def test_cluster_checks_thread_logic_logs(self, cluster, test_cluster_servlet):
+        from runhouse.constants import SERVER_LOGFILE
+
+        cluster.save()
+
+        server_logfile_path = Path(SERVER_LOGFILE)
```
jlewitt1, line 128: shouldn't we do this on the cluster instead of locally?

tests/test_servers/test_servlet.py
BelSasha: Not sure there is an other way to test it, that will not require creating an additional method in the ClusterServlet. 
1. We need to call this thread.is_alive() on the ClusterServlet itself, and we'll be able to do it only if the cluster servlet will have the `_cluster_periodic_thread_alive` method. Another way to test it is to get the `cluster_checks_thread` itself, but this will also require creating a get method.
2. The other tests executes and tests the specific logic of the periodic_cluster_checks thread. We executing the helping functions (`acheck_cluster_logs`, `acheck_cluster_status`) locally. So those tests are not really check that the thread itself is running.

runhouse/servers/cluster_servlet.py
```diff
@@ -84,6 +85,16 @@ async def aset_cluster_config(self, cluster_config: Dict[str, Any]):
 
         self.cluster_config = cluster_config
 
+        new_cluster_name = self.cluster_config.get("name", None)
+
+        if self._cluster_name != new_cluster_name:
```
jlewitt1, line 90: confused by what's going on here - why do we need to reset the cluster name if we area already pulling it from the cliuster config above?

runhouse/servers/cluster_servlet.py
```diff
@@ -232,7 +243,9 @@ async def asave_status_metrics_to_den(self, status: dict):
 
         status_data = {
             "status": ResourceServerStatus.running,
-            "resource_type": status_copy.get("cluster_config").pop("resource_type"),
+            "resource_type": status_copy.get("cluster_config").pop(
```
jlewitt1, line 246: why wouldn't resource type be included?

runhouse/servers/cluster_servlet.py
```diff
@@ -549,7 +595,9 @@ async def send_cluster_logs_to_den(
             "start_line": prev_end_log_line,
             "end_line": new_end_log_line,
         }
-
+        logger.info(
```
jlewitt1, line 598:
```suggestion
        logger.debug(
```

runhouse/servers/cluster_servlet.py
BelSasha: because according to the status schema, it is saved outside of the cluster config, so it is redundant to save it twice.

runhouse/servers/cluster_servlet.py
BelSasha: We are setting a new cluster_config, so we want to verify that the cluster name saved as ClusterServlet property is matching with the cluster_name set in the new cluster_config we set.

jlewitt1: should we even be allowing updating the name here? i feel like that can lead to all sorts of issues with mismatches between what's saved on the cluster vs in den

jlewitt1 (approved):

PR #1209, opened 2024-08-28
print cloud info in cluster status

BelSasha:
Output will looks as follows:

**Single-node GPU**
imageAWS g5.xlarge cluster | :earth_africa: us-east-1 | :moneybag: $1.006/hr

**Multi-node GPU**
image2x AWS g5.xlarge cluster | :earth_africa: us-east-1 | :moneybag: $1.006/hr

**Single-node CPU**
image AWS m6i.xlarge cluster | :earth_africa: us-east-1 | ::money_with_wings:: $0.096/hr

**Multi-node CPU**
image 2x AWS m6i.xlarge cluster | :earth_africa: us-east-1 | ::money_with_wings:: $0.096/hr

### DIFF
--- a/tests/test_resources/test_clusters/test_cluster.py
+++ b/tests/test_resources/test_clusters/test_cluster.py
@@ -591,7 +591,12 @@ class TestCluster(tests.test_resources.test_resource.TestResource):
         assert "node: " in status_output_string
         assert status_output_string.count("node: ") >= 1
 
-        # if it is a GPU cluster, check GPU print as well
+        cloud_properties = cluster.config().get("launched_properties", None)
+        if cloud_properties:
+            properties_to_check = ["cloud", "instance_type", "region", "cost_per_hour"]
+            for p in properties_to_check:
+                property_value = cloud_properties.get(p)
+                assert property_value in status_output_string
 
     @pytest.mark.level("local")
     @pytest.mark.clustertest
--- a/runhouse/main.py
+++ b/runhouse/main.py
@@ -39,7 +39,6 @@ from runhouse.resources.hardware.ray_utils import (
     kill_actors,
 )
 
-
 # create an explicit Typer application
 app = typer.Typer(add_completion=False)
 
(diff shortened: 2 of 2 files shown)


BelSasha:
* **#1209** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1209?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tests/test_resources/test_clusters/test_cluster.py
```diff
@@ -590,7 +590,16 @@ def status_cli_test_logic(self, cluster, status_cli_command: str):
         assert "node: " in status_output_string
         assert status_output_string.count("node: ") >= 1
 
-        # if it is a GPU cluster, check GPU print as well
+        cloud_properties = cluster.config().get("launched_properties", None)
+        if cloud_properties:
+            properties_to_check = ["cloud", "instance_type", "region", "cost_per_hour"]
+            for p in properties_to_check:
+                property_value = (
```
jlewitt1, line 597: what about the `cloud` property necessitates this?

BelSasha: We could add a validation for the `launched_properties` in the status output, but i think it is more of a den test.
The point of test is to make sure that the cloud info is printed as expected to `stdout`. 
Or did you mean something else?

jlewitt1: i meant what makes cloud differnt from the other properties, but seems like it's just bc of how Sky formats it?

BelSasha: yes, i think so. We could also double check it with Caroline.

carolineechen: you added this line `cloud = cloud_properties.get("cloud").upper()` in `_print_cloud_properties`. I think this might be in response to me adding the `cloud = str(launched_resource.cloud).lower()` in setting the launched properties cloud attribute?

In the sky status resource, they have the cloud capitalized (e.g. Kubernetes, AWS, GCP, Lambda, etc) but I thought it simpler to just use the lowercase form (since that's what we prefer/use in our examples anyways). I think we can leave it that way for printing out in status too? But if we want the original form then we should update the launched_properties attribute to that too.

carolineechen (approved):

PR #1219, opened 2024-08-30
Add Multicloud Airflow example

py-rh:


py-rh:
I lied, it's just one example not two :/ 

Thought it's better to add this one than make Kubeflow/Argo work all at once

mkandler (approved):
👍

PR #1225, opened 2024-09-02
Introduce cluster list cli cmd

BelSasha:


BelSasha:
* **#1225** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1225?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

runhouse/main.py
```diff
@@ -578,6 +594,93 @@ def status(
     _print_status(cluster_status, current_cluster)
 
 
+async def aget_clusters_from_den():
```
jlewitt1, line 597: why do we have to do this async?

runhouse/main.py
```diff
@@ -578,6 +594,93 @@ def status(
...
+            den_cluster.get("status") if den_cluster.get("status") else "unknown"
+        )
+        cluster_status_colored = get_status_color(cluster_status)
+        table.add_row(cluster_name, cluster_type, cluster_status_colored)
+
+    console.print(table)
+
+    if len(on_demand_clusters_sky) > 0:
+        console.print(
+            f"There are {len(on_demand_clusters_sky)} live clusters that are not saved in Den. To get information about them, please run [bold italic]sky status -r[/bold italic]."
```
jlewitt1, line 676:
```suggestion
            f"There are {len(on_demand_clusters_sky)} live clusters that are not saved in Den. For more information please run [bold italic]sky status -r[/bold italic]."
```

runhouse/utils.py
```diff
@@ -652,3 +652,15 @@ def get_gpu_usage(collected_gpus_info: dict, servlet_type: ServletType):
...
+
+class StatusColors(str, Enum):
+    RUNNING = "[green]Running[/green]"
+    SERVER_DOWN = "[orange1]Runhouse server down[/orange1]"
+    TERMINATED = "[red]Terminated[/red]"
+    UNKNOWN = "Unknown"
+    LOCAL_CLUSTER = "[bright_yellow]Local cluster[/bright_yellow]"
+
+
+def get_status_color(status: str):
```
jlewitt1, line 665: could make it a class method? 

```
    @classmethod
    def get_status_color(cls, status: str):
        return getattr(cls, status.upper()).value
```

runhouse/main.py
```diff
@@ -578,6 +594,93 @@ def status(
...
+            f"This feature is available only for Den users. Please run {sky_cli_command_formatted} to get on-demand cluster(s) information or sign-up Den."
+        )
+        return
+
+    on_demand_clusters_sky = sky.status(refresh=True)
+
+    clusters_in_den_resp = get_clusters_from_den()
+
+    if clusters_in_den_resp.status_code != 200:
+        logger.error(f"Failed to load {rns_client.username}'s clusters from Den")
```
jlewitt1, line 633:
```suggestion
        logger.error(f"Failed to load clusters from Den for username: {rns_client.username}")
```

runhouse/main.py
```diff
@@ -578,6 +594,93 @@ def status(
...
+    # logged out case
+    if not rh.configs.token:
+        # TODO [SB]: adjust msg formatting (coloring etc)
+        sky_cli_command_formatted = f"{italic_bold_ansi}sky status -r{reset_format}"  # will be printed bold and italic
+        console.print(
+            f"This feature is available only for Den users. Please run {sky_cli_command_formatted} to get on-demand cluster(s) information or sign-up Den."
+        )
+        return
+
+    on_demand_clusters_sky = sky.status(refresh=True)
```
jlewitt1, line 628:
```suggestion
    sky_clusters = sky.status(refresh=True)
```

runhouse/main.py
```diff
@@ -578,6 +594,93 @@ def status(
...
+def cluster_list():
+    """Load Runhouse clusters"""
+    import sky
+
+    # logged out case
+    if not rh.configs.token:
+        # TODO [SB]: adjust msg formatting (coloring etc)
+        sky_cli_command_formatted = f"{italic_bold_ansi}sky status -r{reset_format}"  # will be printed bold and italic
+        console.print(
+            f"This feature is available only for Den users. Please run {sky_cli_command_formatted} to get on-demand cluster(s) information or sign-up Den."
```
jlewitt1, line 624:
```suggestion
            f"Runhouse token required to view all clusters. Please run `runhouse login` to load your token. To view on-demand clusters run {sky_cli_command_formatted}"
```

runhouse/main.py
BelSasha: because we use `httpx_client = httpx.AsyncClient()`. We it use to be consistent with other calls we make to den from OSS. 

There is a a sync wrapper for `aget_clusters_from_den`.

runhouse/utils.py
BelSasha: Yeah, will make the change in  cluster status list [PR](https://github.com/run-house/runhouse/pull/1233), because in this PR we move all core logic of clusters list to be static / class methods of `rh.Cluster`.

jlewitt1: oh didnt mean to move it into cluster, jsut making it a class method of `StatusColors`, but yea doesn't really matter

BelSasha: oh I see, fair point, change it

runhouse/main.py
```diff
@@ -576,6 +592,95 @@ def status(
...
+    return clusters_in_den_resp
+
+
+def get_clusters_from_den():
+    return asyncio.run(aget_clusters_from_den())
+
+
+@cluster_app.command("list")
+def cluster_list():
+    """Load Runhouse clusters"""
```
jlewitt1, line 614:
```suggestion
    """List all Runhouse clusters. Includes those saved locally and in Den."""
```

runhouse/main.py
```diff
@@ -576,6 +592,95 @@ def status(
...
+        # getting the on-demand clusters that are not saved in den.
+        sky_clusters = [
+            cluster
+            for cluster in sky_clusters
+            if f'/{rns_client.username}/{cluster.get("name")}'
+            not in clusters_in_den_names
+        ]
+
+    total_clusters = len(clusters_in_den)
+    table_title = f"[bold cyan]{rns_client.username}'s Clusters (Total: {total_clusters})[/bold cyan]"
```
jlewitt1, line 653: thoughts on left-aligning this text instead of it being in the center above the table?

jlewitt1, line 653:
```suggestion
    table_title = f"[bold cyan]{rns_client.username}'s clusters (Total: {total_clusters})[/bold cyan]"
```

jlewitt1, line 653: wondering if instead of just giving a total, we also show some sort of high level breakdown for number of clusters that are running vs not running (in the case that no filters are specified).. wdyt?

runhouse/main.py
```diff
@@ -576,6 +592,95 @@ def status(
...
+    # logged out case
+    if not rh.configs.token:
+        # TODO [SB]: adjust msg formatting (coloring etc)
+        sky_cli_command_formatted = f"{italic_bold_ansi}sky status -r{reset_format}"  # will be printed bold and italic
+        console.print(
+            f"Runhouse token required to view all clusters. Please run `runhouse login` to load your token. To view on-demand clusters run {sky_cli_command_formatted}"
+        )
+        return
+
+    sky_clusters = sky.status(refresh=True)
```
jlewitt1, line 626: any value to making `refresh` configurable here? if its True will query the cloud provider directly each time

runhouse/main.py
jlewitt1: ah nevermind, i see you did that in the following PR :)

jlewitt1: Maybe something like: "Clusters for {username} (Total: {total_clusters})"

runhouse/main.py
jlewitt1: i think would be cleaner to just call `rns_client.session.get`, instead of setting up separate helpers and http client to manage this

runhouse/main.py
BelSasha: I think that if we won't call pass `refresh=True`, we might get results that are cached. That means that some clusters maybe shown as running, even thought they are not. Having said that, if there are running sky clusters which are not saved in den, we are displaying the following message to the users: `There are N live clusters that are not saved in Den. For more information, please run sky status -r`. So maybe it's fine not to pass refresh=True here. wdyt?

runhouse/main.py
```diff
@@ -576,6 +590,94 @@ def status(
...
+    clusters_in_den_resp = rns_client.session.get(
+        f"{rns_client.api_server_url}/resource",
+        params=get_clusters_params,
+        headers=rns_client.request_headers(),
+    )
+
+    return clusters_in_den_resp
+
+
+def get_clusters_from_den():
```
jlewitt1, line 605: if we use the rns sesson client i think we can call it directly without wrapping in asyncio?

runhouse/main.py
jlewitt1: yea i think as long as we're instructing the user to refresh with -r then its fine to not do so here

jlewitt1 (approved):

PR #1228, opened 2024-09-03
load cluster token from Den instead of generating client side

jlewitt1:
- generate cluster tokens in Den instead of client side. tokens are now OIDC compliant, and required for enabling telemetry
- prevent regenerating the cluster tokens unnecessarily on the client side
- ping new den API on clusters in order to validate the token received

### DIFF
--- a/runhouse/rns/rns_client.py
+++ b/runhouse/rns/rns_client.py
@@ -1,4 +1,3 @@
-import hashlib
 import importlib
 import json
 import os
@@ -197,7 +196,6 @@ class RNSClient:
             headers: dict = self._configs.request_headers
 
         if not headers:
-            # TODO: allow this? means we failed to load token from configs
             return None
 
         if "Authorization" not in headers:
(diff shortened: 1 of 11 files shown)


jlewitt1:
* **#1228** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1228?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

runhouse/rns/rns_client.py
```diff
@@ -220,18 +218,33 @@ def request_headers(
...
+            data=json.dumps(token_payload),
+            headers=self._configs.request_headers,
+        )
+        if resp.status_code != 200:
+            raise Exception(
+                f"Received [{resp.status_code}] from Den POST '{uri}': Failed to load cluster token: {load_resp_content(resp)}"
+            )
+
+        token_data = read_resp_data(resp)
+        return token_data.get("token")
```
mkandler, line 247: Should this method store the token in the config if it's not set already? To prevent future requests?

jlewitt1: I added a decorator to store the value of the cluster token for a given resource address an an in-mem cache for the life of the python process, otherwise i think could get messy if we're having to store lots of different tokens depending on the resource address. wdyt?

mkandler: Seems like a good solution to me! Is that cache invalidatible? Might need to do that if a token becomes stale but probably not an immediate problem

mkandler (approved):
👍

PR #1229, opened 2024-09-03
add num ports to try for ssh tunnel as constant

jlewitt1:


jlewitt1:
* **#1229** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1229?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

dongreenberg:
I actually don't want this to be a cluster setting, it's not a property of the cluster itself and we mostly don't need to add another setting to the cluster for it. I think we can leave everything as is but just make NUM_PORTS_TO_TRY a global instead of hardcoded as 10, so if a user needs to change it they can just do `runhouse.constants.NUM_PORTS_TO_TRY = 10` as a stopgap. We can figure out how to make it a proper setting in the future but I don't think it's really a property of the cluster itself.

runhouse/resources/hardware/cluster.py
```diff
@@ -116,6 +116,9 @@ def __init__(
         if self._default_env and not self._default_env.name:
             self._default_env.name = _unnamed_default_env_name(self.name)
 
+        self._tunnel_timeout = kwargs.get("tunnel_timeout")
```
carolineechen, line 119: should this be `self._tunnel_timeout = kwargs.get("tunnel_timeout", TUNNEL_TIMEOUT)` and similarly with num_ports_to_try below?

jlewitt1: deferring to donny's temp approach for now , we should revisit this when we make it a real setting

carolineechen (approved):

PR #1231, opened 2024-09-04
add last_active info to cluster list

BelSasha:

PR #1227, opened 2024-09-03
By default, cluster list prints only running clusters

BelSasha:
Output will be as follows:

image

PR #1233, opened 2024-09-04
cluster list filters

BelSasha:
`runhouse cluster list --a`  output:
image


BelSasha:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1233?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1233** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1233?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1231** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1231?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1227** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1227?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1225** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1225?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

runhouse/main.py
```diff
@@ -595,27 +593,94 @@ def status(
...
+            # Default behaviour: Print Running clusters that were active it the last 24 hours
+            now = datetime.datetime.utcnow().replace(tzinfo=datetime.timezone.utc)
+            if (now - last_active_at).total_seconds() <= LAST_ACTIVE_AT_TIMEFRAME:
+                table = _add_cluster_as_table_row(table, rh_cluster)
+
+
+def _print_msg_live_sky_clusters(live_clusters_not_in_den: int):
+    if live_clusters_not_in_den > 0:
+        live_sky_clusters_str = (
+            f"There are {live_clusters_not_in_den} live clusters that are not saved in Den."
```
jlewitt1, line 646: i think we can just write cluster(s) instead of branching on 1 vs more than 1 cluster

runhouse/main.py
```diff
@@ -595,27 +593,94 @@ def status(
...
+def supported_time_units():
+    return ", ".join(list(TIME_UNITS.keys()))
 
 
 @cluster_app.command("list")
-def cluster_list():
+def cluster_list(
+    get_all_clusters: bool = typer.Option(
+        False,
+        "--a",
```
jlewitt1, line 666:
```suggestion
        "--all",
```
(for the long option)

runhouse/main.py
```diff
@@ -595,27 +593,94 @@ def status(
...
+        "-a",
+        help="Get all den clusters. Up to 50 most recently active clusters will be displayed.",
```
jlewitt1, line 668: doesn't this also include sky clusters?

runhouse/resources/hardware/cluster.py
```diff
@@ -1957,3 +1964,195 @@ def _folder_mv(
...
+            den_clusters=den_clusters, sky_live_clusters=sky_live_clusters
+        )
+
+        (
+            running_clusters,
+            not_running_clusters,
+        ) = Cluster._get_running_and_not_running_clusters(clusters=den_clusters)
+        all_clusters = running_clusters + not_running_clusters
+
+        return all_clusters, running_clusters, sky_live_clusters
```
jlewitt1, line 2158: wondering if it makes more sense to return a single dict instead of seaparate lists, wdyt?

runhouse/resources/hardware/cluster.py
```diff
@@ -1957,3 +1964,195 @@ def _folder_mv(
...
+            cluster_filters=cluster_filters
+        )
+        if den_clusters_resp.status_code != 200:
+            logger.error(f"Failed to load {rns_client.username}'s clusters from Den")
+            den_clusters = []
+        else:
+            den_clusters = den_clusters_resp.json().get("data")
+
+        # get sky live clusters
+        sky_live_clusters = sky.status(refresh=True)
```
jlewitt1, line 2143: just confirming - does this by default only show live ones? (what about ones in "INIT" state for example?)

runhouse/resources/hardware/cluster.py
```diff
@@ -1957,3 +1964,195 @@ def _folder_mv(
...
+        )
+
+        return running_clusters, not_running_clusters
+
+    ###############################
+    # Cluster list
+    ###############################
+    @staticmethod
+    def list(
+        get_all_clusters: Optional[bool] = False,
```
jlewitt1, line 2119: do we need a param for this? or can we just say if the others aren't specifed then we just bring all of them

runhouse/main.py
BelSasha: Nope, if there live sky clusters that are not saved in den, a relevant message will be displayed.

runhouse/resources/hardware/cluster.py
BelSasha: Don't really have a strong opinion about that, we could return a single dict, probably it will be more elegant to do so.

runhouse/resources/hardware/cluster.py
BelSasha: According to skypilot [documentation](https://skypilot.readthedocs.io/en/latest/reference/api.html#sky-status), `sky.status()` will return clusters with the following statuses: `UP`, `INIT` or `STOPPED`. 

If the cluster is saved in den, we won't treat it as sky cluster, even if its status is `INIT` or `STOPPED`. 

It’s more of a product question, but we’ll probably need to decide how to handle `INIT` clusters that are saved to the den. I think it might be worth consulting with Donny about this because I remember him mentioning that we want to minimize the use of sky commands and create a wrapper around them, so users will interact with the Runhouse API instead of the Skypilot API.

runhouse/resources/hardware/cluster.py
BelSasha: hmm, i actually think that maybe it is worth keeping it.

If no filters are specified, the current default behaviour is that we display only running clusters that where active in the last 24 hours. Therefore, we won't need to get all clusters (running, terminated etc), and then filter the den response to match the default behaviour. Instead, we'll just send a query to den to get the running clusters in the last 24 hours.

BelSasha: rename `get_all_clusters` to `show_all`

runhouse/main.py
```diff
@@ -610,138 +610,140 @@ def get_clusters_from_den():
...
-        }
-        running_clusters.append(
-            cluster_info
-        ) if cluster_status == "running" else not_running_clusters.append(cluster_info)
+    all_clusters = clusters.get("all_clusters", None)
+    running_clusters = clusters.get("running_clusters", None)
+    sky_clusters = clusters.get("sky_clusters", None)
+
+    if not all_clusters:
+        console.print("No existing clusters.")
```
jlewitt1, line 711: can we avoid printing an empty table in this scenario? 
and the header that we show (Running: 0/0, Total Displayed), etc. can also be removed in this case

jlewitt1, line 711:
```suggestion
        console.print("No clusters found.")
```

runhouse/main.py
```diff
@@ -610,138 +610,140 @@ def get_clusters_from_den():
...
 
 
-@cluster_app.command("list")
-def cluster_list():
-    """Load Runhouse clusters"""
-    import sky
+def _create_output_table(
+    total_clusters: int, running_clusters: int, displayed_clusters: int
+):
+    displayed_running_clusters = (
```
jlewitt1, line 616: wondering if we should have some sort of pagination feature here, meaning if for example there are 100 clusters and we show 10 when the command is first run, we could make it easy for the user to toggle through the complete list. wdyt? 

cc @mkandler - we're doing something similar for the resources page in Den right now also

runhouse/resources/hardware/cluster.py
```diff
@@ -1959,3 +1967,202 @@ def _folder_mv(
 
     def _folder_exists(self, path: Union[str, Path]):
         return self.client.folder_exists(path=path)
+
+    ###################################
+    # Cluster list helping methods
+    ###################################
+
+    @staticmethod
+    def _parse_time_duration(duration: str):
```
jlewitt1, line 1976: some of these feel more like pure "utils", and less specific to the cluster class specifically. thoughts on moving these into the `utils.py` instead?

runhouse/resources/hardware/cluster.py
```diff
@@ -1959,3 +1967,202 @@ def _folder_mv(
...
+            logger.error(f"Failed to load {rns_client.username}'s clusters from Den")
+            den_clusters = []
+        else:
+            den_clusters = den_clusters_resp.json().get("data")
+
+        # get sky live clusters
+        sky_live_clusters = sky.status(refresh=True)
+
+        if not sky_live_clusters and not den_clusters:
+            return [], [], []
```
jlewitt1, line 2151:
```suggestion
            return {}
```

runhouse/resources/hardware/cluster.py
```diff
@@ -1959,3 +1967,202 @@ def _folder_mv(
...
+
+    ###############################
+    # Cluster list
+    ###############################
+    @staticmethod
+    def list(
+        show_all: Optional[bool] = False,
+        since: Optional[bool] = None,
+        status: Optional[Union[str, ResourceServerStatus]] = None,
+    ):
```
jlewitt1, line 2127: can we make this a classmethod instead and add docstrings?

runhouse/main.py
```diff
@@ -610,138 +610,140 @@ def get_clusters_from_den():
...
+    show_all: bool = typer.Option(
+        False,
+        "--a",
+        "--all",
+        help="Get all den clusters. Up to 50 most recently active clusters will be displayed.",
+    ),
+    since: Optional[str] = typer.Option(
+        None,
+        "--since",
+        help="Time duration to filter on. Minimum time to filter on is 1 minute. If the provided value will be less "
```
jlewitt1, line 683:
```suggestion
        help="Time duration to filter on. Minimum allowable filter is 1 minute. If the provided value will be less "
```

runhouse/main.py
```diff
@@ -610,138 +610,140 @@ def get_clusters_from_den():
...
+        "than that, the filter will be set to 1 minute by default. You may filter by seconds (s), "
+        "minutes (m), hours (h) or days (s). Examples: 30s, 15m, 2h, 3d.",
+    ),
+    cluster_status: Optional[str] = typer.Option(
+        None,
+        "--status",
+        help="Cluster status to filter on. Supported filter values: running, terminated, runhouse_daemon_down.",
```
jlewitt1, line 690: thoughts on changing the input option to `down` instead to make it easier for the CLI command? can still internally use the `runhouse_daemon_down` value

runhouse/main.py
```diff
@@ -610,138 +610,140 @@ def get_clusters_from_den():
...
+    ),
+):
+    """Load Runhouse clusters"""
+
+    # logged out case
+    if not rh.configs.token:
+        # TODO [SB]: adjust msg formatting (coloring etc)
+        sky_cli_command_formatted = f"{italic_bold_ansi}sky status -r{reset_format}"  # will be printed bold and italic
+        console.print(
+            f"This feature is available only for Den users. Please run {sky_cli_command_formatted} to get on-demand cluster(s) information or sign-up Den."
```
jlewitt1, line 700:
```suggestion
            f"Listing clusters requires a Runhouse token. Please run  `runhouse login` to get your token, or run {sky_cli_command_formatted} to list locally stored on-demand clusters."
```

runhouse/resources/hardware/cluster.py
```diff
@@ -1959,3 +1967,202 @@ def _folder_mv(
...
+
+        return running_clusters, not_running_clusters
+
+    ###############################
+    # Cluster list
+    ###############################
+    @staticmethod
+    def list(
+        show_all: Optional[bool] = False,
+        since: Optional[bool] = None,
```
jlewitt1, line 2125: isn't this a string?

runhouse/resources/hardware/cluster.py
```diff
@@ -1959,3 +1967,202 @@ def _folder_mv(
...
+        else:
+            den_clusters = den_clusters_resp.json().get("data")
+
+        # get sky live clusters
+        sky_live_clusters = sky.status(refresh=True)
+
+        if not sky_live_clusters and not den_clusters:
+            return [], [], []
+
+        sky_live_clusters = Cluster._get_unsaved_live_clusters(
```
jlewitt1, line 2153: a little confused by the naming - just to clarify:

`sky_live_clusters`: sky clusters found in the local sky DB but not saved in den? 
`running_clusters`: running clusters which are saved in Den?
`not running clusters`: clusters that are terminated / unknown / down which are also saved in Den?

runhouse/resources/hardware/cluster.py
```diff
@@ -1959,3 +1967,202 @@ def _folder_mv(
...
+            return
+
+        return time_duration
+
+    @staticmethod
+    def _parse_filters(since: str, cluster_status: str):
+        cluster_filters = {}
+
+        if since:
+            last_active_in: int = Cluster._parse_time_duration(
```
jlewitt1, line 2000: can we wrap in a try/except, so in the event an invalid value is passed in we just use the default?

runhouse/resources/hardware/cluster.py
BelSasha: exactly, will add this explanation as a comment

runhouse/main.py
```diff
@@ -590,159 +592,149 @@ def status(
...
-    else:
-        clusters_in_den = clusters_in_den_resp.json().get("data")
+    return table
 
-    clusters_in_den_names = [cluster.get("name") for cluster in clusters_in_den]
 
-    if not sky_clusters and not clusters_in_den:
-        console.print("No existing clusters.")
+def _add_clusters_to_output_table(
+    table: Table, clusters: list[Dict], filters_requested: dict = None
```
jlewitt1, line 637: doesnt look like we're using `filters_requested` param here

runhouse/main.py
```diff
@@ -590,159 +592,149 @@ def status(
...
+        table=table, clusters=clusters_to_print, filters_requested=filters_requested
+    )
 
     console.print(table)
 
-    if len(sky_clusters) > 0:
-        console.print(
-            f"There are {len(sky_clusters)} live cluster(s) that are not saved in Den. For more information, please run [bold italic]sky status -r[/bold italic]."
-        )
+    _print_msg_live_sky_clusters(sky_clusters=len(sky_clusters))
```
jlewitt1, line 737: small nit - if only using in one place no need for a separate helper for this print?

runhouse/resources/hardware/cluster.py
```diff
@@ -1946,3 +1950,162 @@ def _folder_mv(
 
     def _folder_exists(self, path: Union[str, Path]):
         return self.client.folder_exists(path=path)
+
+    @staticmethod
+    def get_clusters_from_den(cluster_filters: dict):
```
jlewitt1, line 1955: this one still feels odd to me being inside the `cluster` class. i would expect methods here to be specifc to a specific cluster object, not a retreival of multiple cluster objects. we could also keep it inthis file, just as a standalone func outside of the cluster class

runhouse/resources/hardware/cluster.py
```diff
@@ -1946,3 +1950,162 @@ def _folder_mv(
...
+        clusters_in_den_resp = rns_client.session.get(
+            f"{rns_client.api_server_url}/resource",
+            params=get_clusters_params,
+            headers=rns_client.request_headers(),
+        )
+
+        return clusters_in_den_resp
+
+    @staticmethod
+    def _get_unsaved_live_clusters(den_clusters: list[Dict], sky_live_clusters: list):
```
jlewitt1, line 1981: similar note as above, these aren't really methods that are speciifc to a specific runhouse cluster object, probably make sense for them to live elsewhere

runhouse/resources/hardware/cluster.py
```diff
@@ -1946,3 +1950,162 @@ def _folder_mv(
...
+        show_all: Optional[bool] = False,
+        since: Optional[str] = None,
+        status: Optional[Union[str, ResourceServerStatus]] = None,
+    ):
+        """
+        Returns user's runhouse clusters saved in Den. If filters are provided, only clusters that are matching the
+        filters are returned. If not filters are provided, clusters that were active in the last 24 hours are returned.
+
+        Args:
+            show_all Optional[bool]: Whether to get ALL user's clusters saved in den.
```
jlewitt1, line 2064: nit - we should be consistent and follow the same google doctrings format: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

so for example: 
show_all (bool, optional)

jlewitt1, line 2064:
```suggestion
            show_all Optional[bool]: Whether to list all clusters saved in Den (max 50).
```

right now we aren't really listing all of them right? only until some max value?

runhouse/resources/hardware/cluster.py
```diff
@@ -1946,3 +1950,162 @@ def _folder_mv(
...
+            since Optional[int]: Clusters that were active in the last N seconds (= the provided value) will be returned.
```
jlewitt1, line 2065:
```suggestion
            since Optional[int]: Clusters that were active in the specified time period will be returned.
```

could also be minuites, days or hours?

runhouse/main.py
```diff
@@ -590,159 +592,149 @@ def status(
...
+        if running_clusters < displayed_clusters
+        else displayed_clusters
     )
+    table_title = f"[bold cyan]Clusters for {rns_client.username} (Running: {displayed_running_clusters}/{running_clusters}, Total Displayed: {displayed_clusters}/{total_clusters})[/bold cyan]"
 
-    return clusters_in_den_resp
+    table = Table(title=table_title)
 
+    if not filters_requested:
+        table.caption = f"[reset]Showing clusters that were active in the last {LAST_ACTIVE_AT_TIMEFRAME / HOUR} hours."
```
jlewitt1, line 611:
```suggestion
        table.caption = f"[reset]Showing clusters that were active in the last{int(LAST_ACTIVE_AT_TIMEFRAME / HOUR)} hours."
```

nit - looks cleaner as an int if we're showing hours

runhouse/resources/hardware/utils.py
```diff
@@ -248,3 +250,58 @@ def up_cluster_helper(cluster, capture_output: Union[bool, str] = True):
...
+
+        if cluster_status.lower() == "down":
+            # added server_down for BC
+            cluster_status = "runhouse_daemon_down"
+
+        elif cluster_status.lower() not in ResourceServerStatus.__members__.keys():
+            logger.warning(
+                f"Invalid cluster status provided ({cluster_status}), status filter set to 'Running'."
+            )
+            cluster_status = ResourceServerStatus.running.value
```
jlewitt1, line 303: should be able to just do `ResourceServerStatus.running` since that class inherits from `str`

runhouse/resources/hardware/cluster.py
```diff
@@ -1946,3 +1950,162 @@ def _folder_mv(
...
+                "Status": cluster_status,
+                "Last Active (UTC)": last_active_at,
+            }
+            running_clusters.append(
+                cluster_info
+            ) if cluster_status == "running" else not_running_clusters.append(
+                cluster_info
+            )
+
+        # TODO: will be used if we'll need to print not-running clusters
```
jlewitt1, line 2035: can remove?

runhouse/main.py
```diff
@@ -590,159 +592,149 @@ def status(
...
     # Sort not-running clusters by the 'Status' column
     # not_running_clusters = sorted(not_running_clusters, key=lambda x: x["Status"])
 
     # Sort clusters by the 'Last Active (UTC)' column
-    not_running_clusters = sorted(
-        not_running_clusters,
-        key=lambda x: (x["Last Active (UTC)"] is None, x["Last Active (UTC)"]),
-        reverse=True,
-    )
+    # not_running_clusters = sorted(
```
jlewitt1, line 708: still needed?

runhouse/main.py
```diff
@@ -590,159 +592,149 @@ def status(
...
+    show_all: bool = typer.Option(
+        False,
+        "--a",
+        "--all",
+        help="Get all den clusters. Up to 50 most recently active clusters will be displayed.",
+    ),
+    since: Optional[str] = typer.Option(
+        None,
+        "--since",
+        help="Time duration to filter on. Minimum allowable filter is 1 minute. If the provided value will be less "
```
jlewitt1, line 672:
```suggestion
        help="Time duration to filter on.  "
```
i think its duplicative since we already say the min allowable is 1 minute

runhouse/resources/hardware/cluster.py
BelSasha: yeah, makes sense. Moved to hardware utils file.

runhouse/resources/hardware/cluster.py
BelSasha: yeah, added usage example to the docstring.

runhouse/main.py
```diff
@@ -590,155 +592,129 @@ def status(
...
 
-    running_clusters = []
-    not_running_clusters = []
+@cluster_app.command("list")
+def cluster_list(
+    show_all: bool = typer.Option(
+        False,
+        "--a",
+        "--all",
+        help="Get all den clusters. Up to 50 most recently active clusters will be displayed.",
```
jlewitt1, line 658:
```suggestion
        help="Get all clusters saved in Den. Up to 50 most recently active clusters will be displayed.",
```

runhouse/main.py
```diff
@@ -590,155 +592,129 @@ def status(
     _print_status(cluster_status, current_cluster)
 
 
-def get_clusters_from_den():
-    get_clusters_params = {"resource_type": "cluster", "folder": rns_client.username}
-    clusters_in_den_resp = rns_client.session.get(
-        f"{rns_client.api_server_url}/resource",
-        params=get_clusters_params,
-        headers=rns_client.request_headers(),
+def _create_output_table(
```
jlewitt1, line 595: can we move some of these private "list" helpers into a separate file and out of main?

runhouse/resources/hardware/cluster.py
```diff
@@ -1946,3 +1954,65 @@ def _folder_mv(
...
+        status: Optional[Union[str, ResourceServerStatus]] = None,
+    ):
+        """
+        Returns user's runhouse clusters saved in Den. If filters are provided, only clusters that are matching the
+        filters are returned. If not filters are provided, clusters that were active in the last 24 hours are returned.
+
+        Args:
+            show_all (bool, optional): Whether to list all clusters saved in Den (max 50).
+            since (str, optional): Clusters that were active in the specified time period will be returned. Value can be in seconds (65s), minutes (3m), hours (2h) or days (7d).
+            status (str, ResourceServerStatus, optional): Clusters with the provided status will be returned.
```
jlewitt1, line 1975: I think it should be:

```suggestion
            status (Union[str, ResourceServerStatus], optional): Clusters with the provided status will be returned.
```

jlewitt1: actually why can't we only support receiving `ResourceServerStatus`?

runhouse/main.py
```diff
@@ -590,155 +592,129 @@ def status(
...
-        clusters_in_den = []
-    else:
-        clusters_in_den = clusters_in_den_resp.json().get("data")
+    return table
 
-    clusters_in_den_names = [cluster.get("name") for cluster in clusters_in_den]
 
-    if not sky_clusters and not clusters_in_den:
-        console.print("No existing clusters.")
+def _add_clusters_to_output_table(table: Table, clusters: list[Dict]):
```
jlewitt1, line 636: dumb nit, but since we're using `typing` here lets keep it consistent
```suggestion
def _add_clusters_to_output_table(table: Table, clusters: List[Dict]):
```

runhouse/resources/hardware/cluster.py
```diff
@@ -1946,3 +1954,65 @@ def _folder_mv(
...
+    @classmethod
+    def list(
+        cls,
+        show_all: Optional[bool] = False,
+        since: Optional[str] = None,
+        status: Optional[Union[str, ResourceServerStatus]] = None,
+    ):
+        """
+        Returns user's runhouse clusters saved in Den. If filters are provided, only clusters that are matching the
+        filters are returned. If not filters are provided, clusters that were active in the last 24 hours are returned.
```
jlewitt1, line 1970:
```suggestion
        filters are returned. If no filters are provided, clusters that were active in the last 24 hours are returned.
```

runhouse/resources/hardware/cluster.py
```diff
@@ -1946,3 +1954,65 @@ def _folder_mv(
...
+
+        Args:
+            show_all (bool, optional): Whether to list all clusters saved in Den (max 50).
```
jlewitt1, line 1973:
```suggestion
            show_all (bool, optional): Whether to list all clusters saved in Den. Maximum of 50 will be listed. (Default: False). 
```

runhouse/resources/hardware/cluster.py
```diff
@@ -1946,3 +1954,65 @@ def _folder_mv(
...
+            since (str, optional): Clusters that were active in the specified time period will be returned. Value can be in seconds (65s), minutes (3m), hours (2h) or days (7d).
```
jlewitt1, line 1974:
```suggestion
            since (str, optional): Clusters that were active in the specified time period will be returned. Value can be in seconds, minutes, hours or days.
```

I think would be good to show some examples in the docstrings below for how to use the "since" param:

```
    Examples:
        >>> Cluster.list(since="15s")
        >>> Cluster.list(since="3m")
        >>> Cluster.list(since="2h")
        >>> Cluster.list(since="7d")
```

runhouse/resources/hardware/cluster.py
```diff
@@ -1946,3 +1954,65 @@ def _folder_mv(
...
+            status (str, ResourceServerStatus, optional): Clusters with the provided status will be returned.
+        """
+
+        import sky
```
jlewitt1, line 1978: in theory we should be able to support this method + CLI command without even needing to install sky right? maybe we should wrap this in a try/except. wdyt?

runhouse/resources/hardware/utils.py
```diff
@@ -248,3 +252,148 @@ def up_cluster_helper(cluster, capture_output: Union[bool, str] = True):
...
+        if unit == "s" and value < 60:
+            value = 60
+
+        time_duration = value * TIME_UNITS.get(unit, 0)
+
+        if time_duration == 0 and value != 0:
+            logger.warning(
+                f"Filter is not applied, invalid time unit provided ({unit})."
+            )
+            return
```
jlewitt1, line 278: any reason not to also return the default here too?

runhouse/resources/hardware/utils.py
```diff
@@ -248,3 +252,148 @@ def up_cluster_helper(cluster, capture_output: Union[bool, str] = True):
...
+        if last_active_in:
+            cluster_filters["since"] = last_active_in
+
+    if cluster_status:
+
+        if cluster_status.lower() == "down":
+            # added server_down for BC
+            cluster_status = "runhouse_daemon_down"
+
+        elif cluster_status.lower() not in ResourceServerStatus.__members__.keys():
```
jlewitt1, line 301: related to my other note - if we change status to an enum we can delete this validation part

runhouse/main.py
BelSasha: yeah, agreed. I was thinking to do it as part of the cli and main.py revamp task, wdyt?

jlewitt1: if its not such a big lift to do now i think that would be good, just to have main be a little cleaner in the meantime as well

runhouse/main.py
```diff
@@ -590,155 +588,72 @@ def status(
...
+        "--all",
+        help="Get all clusters saved in Den. Up to 50 most recently active clusters will be displayed.",
+    ),
+    since: Optional[str] = typer.Option(
+        None,
+        "--since",
+        help="Time duration to filter on. Minimum allowable filter is 1 minute. You may filter by seconds (s), "
+        "minutes (m), hours (h) or days (s). Examples: 30s, 15m, 2h, 3d.",
+    ),
+    cluster_status: Optional[ResourceServerStatus] = typer.Option(
```
jlewitt1, line 605: this enum has more options than what we describe in the "help" message below, maybe we just create a new separate enum for this param?

runhouse/resources/hardware/cluster.py
```diff
@@ -1946,3 +1954,75 @@ def _folder_mv(
...
+    ###############################
+    @classmethod
+    def list(
+        cls,
+        show_all: Optional[bool] = False,
+        since: Optional[str] = None,
+        status: Optional[ResourceServerStatus] = None,
+    ):
+        """
+        Returns user's runhouse clusters saved in Den. If filters are provided, only clusters that are matching the
```
jlewitt1, line 1969: should distinguish between the sky clusters only saved locally and those in Den 

```suggestion
        Returns clusters saved in Den and locally via Sky. If filters are provided, only clusters that are matching the
```

runhouse/resources/hardware/cluster.py
```diff
@@ -1946,3 +1954,75 @@ def _folder_mv(
...
+    ###############################
+    # Cluster list
+    ###############################
+    @classmethod
+    def list(
+        cls,
+        show_all: Optional[bool] = False,
+        since: Optional[str] = None,
+        status: Optional[ResourceServerStatus] = None,
+    ):
```
jlewitt1, line 1967: I think this returns:
```suggestion
    ) -> Dict[str, List[Dict]] :
```

runhouse/utils.py
```diff
@@ -717,3 +721,68 @@ class StatusColors(str, Enum):
...
+        last_active_at_no_offset = str(last_active_at).split("+")[
+            0
+        ]  # The split is required to remove the offset (according to UTC)
+        rh_cluster["Last Active (UTC)"] = last_active_at_no_offset
+        rh_cluster["Status"] = StatusColors.get_status_color(rh_cluster.get("Status"))
+
+        table = add_cluster_as_table_row(table, rh_cluster)
+
+
+def print_sky_clusters_msg(sky_clusters_amount: int, console):
```
jlewitt1, line 784: very small nit - 

```suggestion
def print_sky_clusters_msg(num_sky_clusters: int, console):
```

also i guess now that this is in a helper, we could be precise about displaying the message for a single live cluster vs multiple live clusters

runhouse/resources/hardware/cluster.py
```diff
@@ -1946,3 +1954,82 @@ def _folder_mv(
...
+
+        try:
+            import sky
+
+            # get sky live clusters
+            sky_live_clusters = [
+                {
+                    "Name": sky_cluster.get("name"),
+                    "Cluster Type": "OnDemandCluster (Sky)",
+                    "Status": sky_cluster.get("status"),
```
jlewitt1, line 1992: just confirming this is a string repr of the value, and not some internal sky object as the "status" value?

runhouse/resources/hardware/utils.py
```diff
@@ -248,3 +258,152 @@ def up_cluster_helper(cluster, capture_output: Union[bool, str] = True):
...
+    # A simple parser for duration like "15m", "2h", "3d"
+    try:
+        unit = duration[-1]
+        value = int(duration[:-1])
+
+        # if the user provides a "--since" time filter than is less than a minute, set it by default to one minute.
+        if unit == "s" and value < 60:
+            value = 60
+
+        time_duration = value * TIME_UNITS.get(unit, 0)
```
jlewitt1, line 278: small nit, but instead of hard coding a constant here can't we use regex, something like? 
`match = re.match(r"(\d+)([smhd])$", value)`

runhouse/resources/hardware/utils.py
```diff
@@ -248,3 +258,152 @@ def up_cluster_helper(cluster, capture_output: Union[bool, str] = True):
...
+    if since:
+        last_active_in: int = parse_time_duration(
+            duration=since
+        )  # return in representing the "since" filter in seconds
+        if last_active_in:
+            cluster_filters["since"] = last_active_in
+
+    if cluster_status:
+
+        if cluster_status.lower() == "down":
```
jlewitt1, line 303: instead of down can we reference the enum, just to make sure we're referencing the right values?

runhouse/main.py
```diff
@@ -590,159 +592,70 @@ def status(
...
-
-    # Add columns to the table
-    table.add_column("Name", justify="left", no_wrap=True)
-    table.add_column("Cluster Type", justify="center", no_wrap=True)
-    table.add_column("Status", justify="left")
-
-    running_clusters = []
-    not_running_clusters = []
+    if not all_clusters:
+        console.print(f"No clusters found in Den for {rns_client.username}.")
```
jlewitt1, line 633: should we make this statement more dynamic? for example if i provide some filter that leads to no results, doesn't necessarily mean there were no clusters found in Den

runhouse/main.py
```diff
@@ -590,159 +592,70 @@ def status(
...
-            else None
-        )
-        last_active_at = (
-            last_active_at.replace(tzinfo=datetime.timezone.utc)
-            if last_active_at
-            else None
-        )
+    if show_all:
+        # if user requesting all den cluster, limit print only to 50 clusters max.
+        clusters_to_print = clusters_to_print[: (min(len(clusters_to_print), 50))]
```
jlewitt1, line 643: can we make 50 a constant? at the very least would allow a user to overwrite it to something larger if they want in the meantime

runhouse/main.py
```diff
@@ -590,159 +592,70 @@ def status(
...
+    since: Optional[str] = typer.Option(
+        None,
+        "--since",
+        help="Time duration to filter on. Minimum allowable filter is 1 minute. You may filter by seconds (s), "
+        "minutes (m), hours (h) or days (s). Examples: 30s, 15m, 2h, 3d.",
+    ),
+    cluster_status: Optional[ClustersListStatus] = typer.Option(
+        None,
+        "--status",
+        help="Cluster status to filter on. Supported filter values: running, terminated (cluster is not live), down (runhouser server is down, but the cluster might be live).",
```
jlewitt1, line 612:
```suggestion
        help="Cluster status to filter on. Supported filter values: running, terminated (cluster is not live), down (Runhouse server is down, but the cluster might be live).",
```

runhouse/main.py
```diff
@@ -590,159 +592,70 @@ def status(
...
 @cluster_app.command("list")
-def cluster_list():
-    """List all Runhouse clusters saved in Den"""
-    import sky
+def cluster_list(
+    show_all: bool = typer.Option(
+        False,
+        "--a",
+        "--all",
+        help="Get all clusters saved in Den. Up to 50 most recently active clusters will be displayed.",
```
jlewitt1, line 601: can we replace 50 here with the constant also?

runhouse/resources/hardware/utils.py
```diff
@@ -248,3 +258,152 @@ def up_cluster_helper(cluster, capture_output: Union[bool, str] = True):
...
+        cluster["Last Active (UTC)"] = cluster.get("Last Active (UTC)").strftime(
+            "%m/%d/%Y, %H:%M:%S"
+        )
+
+    running_clusters = sorted(
+        running_clusters, key=lambda x: x["Last Active (UTC)"], reverse=True
+    )
+
+    for cluster in running_clusters:
+        cluster["Last Active (UTC)"] = cluster.get("Last Active (UTC)").strftime(
```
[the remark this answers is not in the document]
BelSasha: nope, because we pass only den clusters to this function (`get_running_and_not_running_clusters`), and den clusters will always have a "Last Active (UTC)" field.

runhouse/resources/hardware/cluster.py
BelSasha: yeah, good point. added a small change, it will look like this:
image

runhouse/main.py
```diff
@@ -590,159 +593,78 @@ def status(
...
+    if not all_clusters:
+        additional_info_to_msg = (
+            " that match the provided filters."
+            if (show_all or since or cluster_status)
+            else "."
         )
-        clusters_in_den = []
-    else:
-        clusters_in_den = clusters_in_den_resp.json().get("data")
+        no_clusters_msg = f"No clusters found in Den for {rns_client.username}{additional_info_to_msg}"
```
jlewitt1, line 639: maybe easier to do:

```
no_clusters_msg = f"No clusters found in Den for {rns_client.username}
if show_all or since or cluster_status:
   no_clusters_msg += " that match the provided filters"
```

runhouse/main.py
```diff
@@ -590,159 +593,78 @@ def status(
...
+        console.print(no_clusters_msg)
+        print_sky_clusters_msg(len(sky_clusters), console)
```
jlewitt1, line 641: something about passing around the console object feels odd, can we import it instead?

runhouse/resources/hardware/utils.py
```diff
@@ -248,3 +259,155 @@ def up_cluster_helper(cluster, capture_output: Union[bool, str] = True):
...
+        ) if cluster_status == "running" else not_running_clusters.append(cluster_info)
+
+    # Sort clusters by the 'Last Active (UTC)' and 'Status' column
+    not_running_clusters = sorted(
+        not_running_clusters,
+        key=lambda x: (x["Last Active (UTC)"], x["Status"]),
+        reverse=True,
+    )
+    for cluster in not_running_clusters:
+        cluster["Last Active (UTC)"] = cluster.get("Last Active (UTC)").strftime(
```
jlewitt1, line 400: maybe we move this conversion to stftime into a helper since we also do it below?

runhouse/main.py
```diff
@@ -590,159 +593,75 @@ def status(
...
-
-
 @cluster_app.command("list")
-def cluster_list():
-    """List all Runhouse clusters saved in Den"""
-    import sky
+def cluster_list(
+    show_all: bool = typer.Option(
+        False,
+        "--a",
```
jlewitt1, line 600: for a short option i think more intuitive for it to be `-a` instead?

runhouse/utils.py
```diff
@@ -717,3 +726,87 @@ class StatusColors(str, Enum):
...
+    table_title = f"[bold cyan]Clusters for {rns_client.username} (Running: {displayed_running_clusters}/{running_clusters}, Total Displayed: {displayed_clusters}/{total_clusters})[/bold cyan]"
+
+    table = Table(title=table_title)
+
+    if not filters_requested:
+        table.caption = f"[reset]Showing clusters that were active in the last {int(LAST_ACTIVE_AT_TIMEFRAME / HOUR)} hours."
+        table.caption_justify = "left"
+
+    if displayed_clusters == MAX_CLUSTERS_DISPLAY:
+        link_to_clusters_in_den = f"[reset]Full list of clusters can be found in https://www.run.house/resources?folder={rns_client.username}&type=cluster. Please make sure yor as signed as {rns_client.username} before following the link."
```
jlewitt1, line 754: if the user isn't logged it they will be prompted to log in via the UI anyways
```suggestion
        link_to_clusters_in_den = f"[reset]The full cluster list can be viewed at https://www.run.house/resources?folder={rns_client.username}&type=cluster."
```

jlewitt1 (approved):

PR #1236, opened 2024-09-05
updated tokens for dev ci testing

jlewitt1:
Update env vars for the various den accounts used in testing

### DIFF
--- a/.github/workflows/local_tests_den_dev.yaml
+++ b/.github/workflows/local_tests_den_dev.yaml
@@ -21,8 +21,8 @@ jobs:
 
   #     - name: pytest -v --level local tests/test_servers/
   #       env:
-  #         TEST_TOKEN: ${{ secrets.TEST_TOKEN }}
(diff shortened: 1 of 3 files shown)


jlewitt1:
* **#1236** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1236?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

mkandler (approved):
So much more clear!! Is this renaming going to break anything without changing the references to the env vars too?

jlewitt1:
actually that's a good point, would we be able to reuse the same tokens for dev/prod? otherwise we would need to branch in the test code a bit to know which env var to load

mkandler:
I think we'll need different tokens for dev and production since they'll be generated by a different private/public key pair on Den dev v Dev Prod (those should definitely not be the same)

PR #1258, opened 2024-09-12
Don't call sky refresh during ondemand cluster init

carolineechen:


carolineechen:
* **#1258** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1258?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

runhouse/resources/hardware/cluster.py
```diff
@@ -128,36 +128,40 @@ def address(self, addr):
...
             )
             connect_call.start()
             connect_call.join(timeout=5)
             if connect_call.is_alive():
+                return False
+            return True
+
+        if not check_connect_server():
+            if self.__class__.__name__ == "OnDemandCluster":
+                self._update_from_sky_status(dryrun=False)
```
dongreenberg, line 143: Is this called every time we first form the connection, or only if it fails based on the IP in the sky db without refreshing?

carolineechen: we try to form the connection, and if it fails (5 sec timeout) then we do a sky refresh + ping to check if the cluster is available before trying to connect again

PR #1246, opened 2024-09-09
Add more sky resource properties to cluster launched properties

carolineechen:
to add to launched_properties, but not directly from sky - memory, os

PR #1152, opened 2024-08-14
Split up `/call` and `/logs`.

rohinb2:

PR #1268, opened 2024-09-19
cluster cli commands revamp

BelSasha:


BelSasha:
* **#1268** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1268?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

runhouse/main.py
```diff
@@ -129,7 +132,7 @@ def notebook(
     c.notebook()
 
 
-@app.command()
+@cluster_app.command("ssh")
 def ssh(cluster_name: str, up: bool = typer.Option(False, help="Start the cluster")):
```
jlewitt1, line 136:
```suggestion
def ssh(cluster_name: str, up: bool = typer.Option(False, help="SSH into a cluster")):
```

runhouse/main.py
```diff
@@ -129,7 +132,7 @@ def notebook(
...
     """SSH into a cluster created elsewhere (so `ssh cluster` doesn't work out of the box) or not yet up."""
```
jlewitt1, line 137: this docstring appears in the "--help", i think we can condense it a bit
```suggestion
    """SSH into a cluster created elsewhere or not yet up."""
```

runhouse/main.py
```diff
@@ -666,6 +341,96 @@ def cluster_list(
         print_sky_clusters_msg(len(sky_clusters))
 
 
+@cluster_app.command("keep_warm")
+def keep_warm(
+    cluster_name: str,
+    up: Optional[bool] = typer.Option(False, help="Start the cluster"),
```
jlewitt1, line 347: maybe this should be a separate command, since keep warm really implies that the cluster is already up. wdyt?

runhouse/main.py
```diff
@@ -666,6 +341,96 @@ def cluster_list(
         print_sky_clusters_msg(len(sky_clusters))
 
 
+@cluster_app.command("keep_warm")
+def keep_warm(
```
jlewitt1, line 345: can we add a docstring here so it also appears in the "help" command: 
"Keep the cluster warm for given number of minutes after inactivity"

runhouse/main.py
```diff
@@ -129,9 +133,14 @@ def notebook(
...
-@app.command()
-def ssh(cluster_name: str, up: bool = typer.Option(False, help="Start the cluster")):
-    """SSH into a cluster created elsewhere (so `ssh cluster` doesn't work out of the box) or not yet up."""
+###############################
+# Cluster CLI commands
+###############################
+
+
+@cluster_app.command("ssh")
+def ssh(cluster_name: str, up: bool = typer.Option(False, help="SSH into a cluster")):
```
jlewitt1, line 142: similar note as above, maybe up should be a separate command?

runhouse/main.py
```diff
@@ -666,6 +351,182 @@ def cluster_list(
...
+    cluster_name: str = typer.Argument(
+        None,
+        help="The name of cluster to terminate.",
+    )
+):
+    """Terminate cluster if it is not down. No-op if cluster is already down.
+    This only applies to on-demand clusters, and has no effect on self-managed clusters.
+
+    Example:
+        .. code-block:: console:
```
jlewitt1, line 513: i dont think sphinx style code blocks are supported here - its not rending well in the CLI.. prob ok to just include the raw example snippet

runhouse/main.py
```diff
@@ -666,6 +351,182 @@ def cluster_list(
...
+
+            $ runhouse cluster down /sashab/rh-basic-cpu
+    """
+    # TODO: add a wrapper for sky down?
```
jlewitt1, line 517: the runhouse `teardown` already does a `sky down`

runhouse/main.py
```diff
@@ -666,6 +351,182 @@ def cluster_list(
...
+    try:
+        current_cluster = rh.cluster(name=cluster_name)
+        current_cluster.teardown()
```
jlewitt1, line 520: maybe we should add a param here for also removing from configs, which would call `teardown_and_delete` instead of just `teardown`.

runhouse/main.py
```diff
@@ -666,6 +351,182 @@ def cluster_list(
...
+
+    """
+    try:
+        current_cluster = rh.cluster(name=cluster_name)
+        current_cluster.up_if_not()
+    except ValueError as e:
+        if str(e) == f"Resource {cluster_name} not found":
+            console.print("Cluster is not saved in den.")
+        else:
+            console.print(f"Failed to bring up the cluster: {e}")
```
jlewitt1, line 497: what other value error could be raised here?

runhouse/main.py
```diff
@@ -613,12 +289,21 @@ def cluster_list(
...
-    """Load Runhouse clusters"""
+    """Load Runhouse clusters.
+
+    Examples:
+        .. code-block:: console:
+
+            $ runhouse cluster list
+            $ runhouse cluster list --all
+            $ runhouse cluster list --since 10m
+            $ runhouse cluster list --status terminate
```
jlewitt1, line 300:
```suggestion
            $ runhouse cluster list --status terminated
```

runhouse/main.py
```diff
@@ -540,7 +210,13 @@ def status(
         help="Whether to update Den with the status",
     ),
 ):
-    """Load the status of the Runhouse daemon running on a cluster."""
+    """Load the status of the Runhouse daemon running on a cluster.
```
jlewitt1, line 213: wondering if we should be more generic here by just saying "load the status of the cluster". not sure how important the distinction would be to a user that this is specifically checking the status of the daemon, since we make the distinction anyways between a running cluster and one where the daemon is down. wdyt?

runhouse/resources/hardware/cluster.py
```diff
@@ -2033,3 +2034,33 @@ def list(
...
+            logs = obj_store.logs(tail=tail)
+        else:
+            logs = self.call_client_method(
+                "logs",
+                tail=tail,
+            )
+
+        return logs
+
+    def _get_saved_logs_from_den(self):
```
jlewitt1, line 2060: this feels more like a helper than a cluster class method

runhouse/servers/cluster_servlet.py
```diff
@@ -664,11 +664,11 @@ def status(self, send_to_den: bool = False):
     ##############################################
     # Save cluster logs to Den
     ##############################################
-    def _get_logs(self):
+    def _get_logs(self, join_by: str = " "):
```
jlewitt1, line 667: in what scenarios would the `join_by` be different?

runhouse/servers/http/http_server.py
```diff
@@ -789,6 +790,13 @@ def get_status(request: Request, send_to_den: bool = False):
 
         return obj_store.status(send_to_den=send_to_den)
 
+    @staticmethod
+    @app.get("/logs")
+    @validate_cluster_access
+    def get_logs(request: Request, tail: int = DEFAULT_SURFACED_LOG_LENGTH):
```
jlewitt1, line 796: noting that rohin is currently working on a PR that handles logs very similarly, might be worth building on that (we can discuss further too): 
https://github.com/run-house/runhouse/pull/1152/files#diff-3e12105589672c203d48d9b023b833a328411f3a3c8d09f92a84ec63e59e003eR469

runhouse/utils.py
```diff
@@ -800,3 +810,335 @@ def print_sky_clusters_msg(num_sky_clusters: int):
         console.print(
             f"{msg} For more information, please run [bold italic]sky status -r[/bold italic]."
         )
+
+
+def adjust_resource_type(resource_type: str):
```
jlewitt1, line 815: maybe `condense_resource_type` is more accurate? meaning we're abbreviating the full resource path for display purposes right?

runhouse/utils.py
```diff
@@ -800,3 +810,335 @@ def print_sky_clusters_msg(num_sky_clusters: int):
...
+    link_to_den_dashboard = f"{api_server_url}/resources/{cluster_uri}"
+    return link_to_den_dashboard
+
+
+###############################
+# Status helping functions
+###############################
+
+
+def print_cluster_config(cluster_config: Dict):
```
jlewitt1, line 840: these console printing helpers are more specifically cli utils, maybe we move those into a dedicated `cli_utils` file? (specifcally the ones that are only relevant to the CLI)

runhouse/utils.py
```diff
@@ -800,3 +810,335 @@ def print_sky_clusters_msg(num_sky_clusters: int):
...
+
+                    else:
+                        is_func_running: str = ""
+
+                    resource_info_str = resource_info_str + is_func_running
+
+                    console.print(resource_info_str)
+
+
+def print_cloud_properties(cluster_config: dict):
```
jlewitt1, line 1058: caroline recently added a few more properties (https://github.com/run-house/runhouse/pull/1246/files), can we add these too?

runhouse/main.py
```diff
@@ -666,6 +351,182 @@ def cluster_list(
...
+        bring_logs_from_den = True
+
+    else:
+        # we are inside the cluster
+        current_cluster = cluster_or_local  # cluster_or_local = rh.here
+
+    try:
+        if bring_logs_from_den:
+            console.print("Getting the last chunk of logs saved in Den.")
+            get_logs_den_resp = current_cluster._get_saved_logs_from_den()
```
jlewitt1, line 456: for consistency wdyt about having this take the `tail` value, so the requested number of lines (from den) are returned?

runhouse/main.py
```diff
@@ -666,6 +351,182 @@ def cluster_list(
...
+
+    Example:
+        .. code-block:: console:
+
+            $ runhouse cluster logs /sashab/rh-basic-cpu
+
+    """
+
+    cluster_or_local = rh.here
+    bring_logs_from_den = False
```
jlewitt1, line 410: nit - 
```suggestion
    load_from_den = False
```

runhouse/main.py
BelSasha: yeah, the main open question here is how do we handle sky clusters that are up but not saved in den. 

 I think we can handle it as we handle sky clusters in cluster list: if there are sky clusters that are not saved in den, just print a message saying that there are sky clusters that are not saved in den, therefore they are still up. wdyt?

runhouse/main.py
BelSasha: a value error that is not raised by runhouse. This is an edge case.. I'm on the fence deciding if it will be common enough to have this fork in the code. wdyt?

runhouse/servers/cluster_servlet.py
BelSasha: Added it to support getting logs, can be removed now, since we are waiting for Rohin's implementation.

runhouse/utils.py
BelSasha: current output:
image

printing the `num_cpus` and accelerators info, if exists. Not sure if we need to print the total disk size. @jlewitt1 wdyt?

runhouse/main.py
BelSasha: we could do that, but that may require few pings to den, in case that length of the latest logs saved in den is less than the provided tail size. @jlewitt1

jlewitt1: ah right ok then yea lets just keep as is and return the contents of the latest file

runhouse/utils.py
jlewitt1: yea i think thats fine for now

runhouse/main.py
jlewitt1: is it something you've seen come up? if not i think we can remove it

runhouse/main.py
jlewitt1: yep i think that works

runhouse/servers/http/http_client.py
```diff
@@ -279,6 +281,13 @@ def status(self, send_to_den: bool = False):
             req_type="get",
         )
 
+    def logs(self, tail: int = DEFAULT_SURFACED_LOG_LENGTH):
```
jlewitt1, line 284: can we also comment this out for now too?

runhouse/cli_utils.py
```diff
@@ -0,0 +1,475 @@
...
+    print_envs_info(env_servlet_processes, current_cluster)
+
+
+def print_bring_cluster_up_msg(
+    cluster_name: str, msg_prefix="Can't execute the command"
+):
+    from runhouse.main import console
+
+    console.print(
+        f"{msg_prefix} because [reset]{cluster_name} is terminated. To bring it up, run [bold italic green]runhouse cluster up {cluster_name}[/bold italic green]."
```
jlewitt1, line 474: i don't think we always no for sure that the cluster status is terminated when we print this message? 
maybe better to change to "not up"

runhouse/main.py
```diff
@@ -679,6 +350,182 @@ def cluster_list(
         print_sky_clusters_msg(len(sky_clusters))
 
 
+@cluster_app.command("keep-warm")
+def cluster_keep_warm(
+    cluster_name: str,
+    mins: Optional[int] = typer.Option(
+        -1, help="Amount of time (in min) to keep the cluster warm after inactivity."
```
jlewitt1, line 357: can we add "If set to -1, keeps cluster warm indefinitely"

runhouse/main.py
```diff
@@ -679,6 +350,182 @@ def cluster_list(
...
+
+    """
+
+    try:
+        current_cluster = rh.cluster(name=cluster_name)
+        if not current_cluster.is_up():
+            print_bring_cluster_up_msg(cluster_name=cluster_name)
+            return
+        current_cluster.keep_warm(mins=mins)
+        # TODO: keep warm when we are in the cluster?
```
jlewitt1, line 373: i think we should support running this on a cluster, should be easy if we just detect that at the beginning of the command

runhouse/main.py
```diff
@@ -679,6 +350,182 @@ def cluster_list(
...
+    except ValueError as e:
```
jlewitt1, line 374: i think we can simplify this everywhere to: 

```
except ValueError:
  console.print("Cluster not found in Den.")
```

runhouse/main.py
```diff
@@ -679,6 +350,182 @@ def cluster_list(
...
+        if str(e) == f"Resource {cluster_name} not found":
+            console.print("Cluster is not saved in den.")
+            # TODO: do we want to check if the cluster is a live sky cluster, and inform the user how to keep_warm using sky?
+        else:
+            console.print(f"Failed to keep the cluster warm: {e}")
+    except Exception as e:
+        console.print(f"Failed to keep the cluster warm: {e}")
+
+
+# TODO [SB/JL]: uncomment and update once Rohin's change is merged.
```
jlewitt1, line 384: dumb nit but can we move this further down just for readability purposes

jlewitt1: also i think we can just remove all the logic here and just have "pass" instead for now

runhouse/main.py
```diff
@@ -679,6 +350,182 @@ def cluster_list(
...
+):
+    """Bring up the cluster if it is not up. No-op if cluster is already up.
+    This only applies to on-demand clusters, and has no effect on self-managed clusters.
+
+    Example:
+        ``$ runhouse cluster up /sashab/rh-basic-cpu``
+
+    """
+    try:
+        current_cluster = rh.cluster(name=cluster_name)
```
jlewitt1, line 486: i think for this command (+ others where relevant) we should handle this being run on a cluster more gracefully - in this case that would mean not doing anything but printing a relevant message

runhouse/resources/hardware/cluster.py
```diff
@@ -2061,3 +2061,27 @@ def list(
             "sky_clusters": sky_live_clusters,
         }
         return clusters
+
+    ###############################
+    # Cluster logs
+    ###############################
+
+    # TODO [SB/JL]: uncomment and update once Rohin's change is merged.
+    # def logs(self, tail: Optional[int] = DEFAULT_SURFACED_LOG_LENGTH):
```
jlewitt1, line 2070: similar note as above, can remove the logic and just replace with "pass" for now

runhouse/cli_utils.py
```diff
@@ -0,0 +1,522 @@
...
+    cluster_name: str, msg_prefix="Can't execute the command"
+):
+    from runhouse.main import console
+
+    console.print(
+        f"{msg_prefix} because [reset]{cluster_name} is not up. To bring it up, run [bold italic green]runhouse cluster up {cluster_name}[/bold italic green]."
+    )
+
+
+def get_cluster_based_on_rh_here(cluster_name: str):
```
jlewitt1, line 482: maybe `get_cluster_or_local`?

runhouse/main.py
```diff
@@ -1092,3 +820,90 @@ def main(
         print(f"{__version__}")
     elif ctx.invoked_subcommand is None:
         subprocess.run("runhouse --help", shell=True)
+
+
+# TODO [SB/JL]: uncomment and update once Rohin's change is merged.
+# TODO [SB/JL]: once supported, need to add cluster_logs to cli.rst
+# @cluster_app.command("logs")
+# def cluster_logs(
+#     cluster_name: str = typer.Argument(
```
jlewitt1, line 829: we can save this logic elsewhere outside this PR, for the time being can we just replace with a "pass"?

runhouse/cli_utils.py
```diff
@@ -0,0 +1,522 @@
...
+    )
+    cpus_gpus_info_str = f"CPUs: {int(float(num_of_cpus))}"
+    if num_of_gpus > 0:
+        gpu_types_str = gpu_types.pop()
+        for gpu_type in gpu_types:
+            gpu_types_str = gpu_types_str + f", {gpu_type}"
+        cpus_gpus_info_str = (
+            cpus_gpus_info_str + f" | GPUs: {num_of_gpus} (Type(s): {gpu_types_str})"
+        )
+    console.print(f"[reset]{cpus_gpus_info_str}")
```
jlewitt1, line 410: do you have an example for what this looks like rendered in the CLI?

runhouse/cli_utils.py
```diff
@@ -0,0 +1,522 @@
...
+    cluster_config = status_data.get("cluster_config")
+    env_servlet_processes = status_data.get("env_servlet_processes")
+
+    cluster_name = cluster_config.get("name", None)
+
+    if cluster_name:
+        api_server_url = cluster_config.get(
+            "api_server_url", rh.configs.get("api_server_url")
+        )
+        api_server_url = api_server_url.replace(
```
jlewitt1, line 426: what are we doing with the URL here? we could just hard code the link, it's not likely to change

runhouse/cli_utils.py
BelSasha: Screenshot 2024-09-30 at 9 31 32

runhouse/cli_utils.py
BelSasha: we convert the api server url to the runhouse UI link. We could hard code it, but than it will be the same for den prod and den dev. But i guess it is ok, since den-dev is used just in house.

runhouse/main.py
```diff
@@ -112,46 +118,37 @@ def logout():
...
+        return
+    current_cluster.notebook()
 
 
-@app.command()
-def ssh(cluster_name: str, up: bool = typer.Option(False, help="Start the cluster")):
-    """SSH into a cluster created elsewhere (so `ssh cluster` doesn't work out of the box) or not yet up."""
+@cluster_app.command("ssh")
+def cluster_ssh(cluster_name: str):
+    """SSH into a cluster created elsewhere or not yet up.
```
jlewitt1, line 138: "or not yet up" isn't relevant anymore here right? cluster needs to be up for this to work

runhouse/main.py
```diff
@@ -679,6 +311,103 @@ def cluster_list(
...
+
+    current_cluster = get_cluster_or_local(cluster_name)
+
+    try:
+        if not current_cluster.is_up():
+            print_bring_cluster_up_msg(cluster_name=cluster_name)
+            return
+        current_cluster.keep_warm(mins=mins)
+    except ValueError:
+        console.print(f"{cluster_name} is not saved in den.")
```
jlewitt1, line 337: dumb nit - we should be consistent in how we write it
```suggestion
        console.print(f"{cluster_name} is not saved in Den.")
```

runhouse/resources/hardware/utils.py
```diff
@@ -336,17 +336,39 @@ def get_clusters_from_den(cluster_filters: dict):
...
+            return [
+                cluster
+                for cluster in sky_live_clusters
+                if f'/{rns_client.username}/{cluster.get("name")}'
+                not in den_clusters_names
+            ]
+        else:
+            return []
+    except Exception:
+        return []
```
jlewitt1, line 357: maybe we could debug the actual exception here?

runhouse/cli_utils.py
```diff
@@ -0,0 +1,516 @@
...
+
+    """Prints the status of the cluster to the console"""
+    cluster_config = status_data.get("cluster_config")
+    env_servlet_processes = status_data.get("env_servlet_processes")
+
+    cluster_name = cluster_config.get("name", None)
+
+    if cluster_name:
+        ui_link = "https://www.run.house/"
+        cluster_link_in_den_ui = get_resource_link_in_den_ui(
```
jlewitt1, line 424: can we hard code the ui_link inside this function instead of defining here and passing in

runhouse/main.py
```diff
@@ -679,6 +311,149 @@ def cluster_list(
...
+    since: Optional[int] = typer.Option(
+        LogsSince.five_min,
+        "--since",
+        help=f"Get the logs since the given timeframe. Available timeframes (minutes): {LogsSince.supported()}",
+    ),
+):
+    """Load the logs of the Runhouse daemon running on a cluster.
+
+    Examples:
+        ``$ runhouse cluster logs /sashab/rh-basic-cpu``
```
jlewitt1, line 426: Add space here

runhouse/main.py
```diff
@@ -679,6 +311,149 @@ def cluster_list(
...
+@cluster_app.command("logs")
+def cluster_logs(
+    cluster_name: str = typer.Argument(
+        None,
+        help="Name of cluster to get the logs from. If not specified, we will get the logs of the local cluster.",
+    ),
+    since: Optional[int] = typer.Option(
+        LogsSince.five_min,
+        "--since",
+        help=f"Get the logs since the given timeframe. Available timeframes (minutes): {LogsSince.supported()}",
```
jlewitt1, line 420:
```suggestion
        help=f"Get the logs from the provided timeframe (in minutes).",
```

runhouse/cli_utils.py
```diff
@@ -0,0 +1,526 @@
...
+    one_min = 1
+    five_min = 5
+    ten_min = 10
+    one_hour = 60
+    three_hours = 3 * (HOUR / 60)
+    six_hours = 6 * (HOUR / 60)
+    one_day = 24 * (HOUR / 60)
+
+    @classmethod
+    def supported(cls):
```
jlewitt1, line 525: can remove

runhouse/cli_utils.py
```diff
@@ -0,0 +1,526 @@
...
+####################################################################################################
+# Cluster logs utils
+####################################################################################################
+
+
+class LogsSince(int, Enum):
+    one_min = 1
+    five_min = 5
+    ten_min = 10
+    one_hour = 60
```
jlewitt1, line 519: might be confusing to have a mix minutes / hours / days here, maybe just keep everything in munutes?

runhouse/main.py
```diff
@@ -679,6 +311,149 @@ def cluster_list(
...
+):
+    """Load the logs of the Runhouse daemon running on a cluster.
+
+    Examples:
+        ``$ runhouse cluster logs /sashab/rh-basic-cpu``
+
+        ``$ runhouse cluster logs /sashab/rh-basic-cpu --since 60``
+
+    """
+    if since and str(since) not in LogsSince.supported().split(", "):
```
jlewitt1, line 431: can remove this validation once the type has been updated for the option

runhouse/main.py
```diff
@@ -679,6 +311,149 @@ def cluster_list(
...
+        since = LogsSince.five_min
+
+    current_cluster = get_cluster_or_local(cluster_name)
+
+    if not current_cluster:
+        return
+
+    cluster_uri = rns_client.resource_uri(current_cluster.rns_address)
+
+    presigned_file_resp = requests.get(
```
jlewitt1, line 444: maybe just call it "resp"? user doesn't need to know specifics about it coming from a pre-signed url

runhouse/main.py
```diff
@@ -112,46 +118,37 @@ def logout():
...
-def notebook(
-    cluster_name: str, up: bool = typer.Option(False, help="Start the cluster")
-):
+###############################
+# Cluster CLI commands
+###############################
+
+
+@cluster_app.command()
+def notebook(cluster_name: str):
```
jlewitt1, line 127: can remove for now

runhouse/main.py
```diff
@@ -679,6 +311,149 @@ def cluster_list(
...
+
+
+@cluster_app.command("logs")
+def cluster_logs(
+    cluster_name: str = typer.Argument(
+        None,
+        help="Name of cluster to get the logs from. If not specified, we will get the logs of the local cluster.",
+    ),
+    since: Optional[int] = typer.Option(
+        LogsSince.five_min,
```
jlewitt1, line 418:
```suggestion
        LogsSince.five_min.value,
```

runhouse/main.py
```diff
@@ -540,43 +203,12 @@ def status(
...
-            "Start it with `runhouse restart` or specify a remote "
-            "cluster to poll with `runhouse status <cluster_name>`."
-        )
-        raise typer.Exit(1)
-
-    else:
-        # we are inside the cluster
-        current_cluster = cluster_or_local  # cluster_or_local = rh.here
+    Example:
+        ``$ runhouse cluster status /sashab/rh-basic-cpu``
```
jlewitt1, line 209: i think we need space here too

runhouse/main.py
```diff
@@ -679,6 +311,149 @@ def cluster_list(
...
+            )
+    except Exception as e:
+        console.print(f"Failed to terminate the cluster: {e}")
+
+
+@cluster_app.command("logs")
+def cluster_logs(
+    cluster_name: str = typer.Argument(
+        None,
+        help="Name of cluster to get the logs from. If not specified, we will get the logs of the local cluster.",
```
jlewitt1, line 415:
```suggestion
        help="Name of cluster to load the logs from. If not specified, loads the logs of the local cluster.",
```

runhouse/cli_utils.py
```diff
@@ -0,0 +1,523 @@
...
+                f"Could not connect to the server on cluster {cluster_name}. Check that the server is up with "
+                f"[reset][bold italic]`runhouse cluster status {cluster_name}`[/bold italic] or [bold italic]`sky status -r`[/bold italic] for on-demand clusters."
+            )
+            raise typer.Exit(1)
+        return current_cluster
+
+    cluster_or_local = rh.here
+    if cluster_or_local == "file" and not cluster_name:
+        # If running outside the cluster must specify a cluster name
+        console.print("Missing argument `cluster_name`.")
```
jlewitt1, line 493: forget if this was already addressed somewhere else, but just to be safe this can `raise typer.Exit(1)` instead of returning, and then we can remove all these: 

```
if not current_cluster:
   return
```

runhouse/cli_utils.py
```diff
@@ -0,0 +1,523 @@
...
+####################################################################################################
+# Cluster status utils
+####################################################################################################
+
+
+def print_cluster_config(cluster_config: Dict):
+    """
+    Helping function to the `_print_status` which prints the relevant info from the cluster config.
+    """
+    # TODO [SB]: need to modify printing format (colour palette etc).
```
jlewitt1, line 152: still relevant?

runhouse/cli_utils.py
```diff
@@ -0,0 +1,523 @@
...
+            total_cluster_memory = math.ceil(
+                int(env_cpu_info["total_memory"]) / (1024**3)
+            )
+            cpu_memory_usage_percent = round(
+                float(env_cpu_info["used_memory"] / env_cpu_info["total_memory"]),
+                2,
+            )
+            cpu_usage_percent = round(float(env_cpu_info["utilization_percent"]), 2)
+
+            cpu_usage_summery = f"{DOUBLE_SPACE_UNICODE}CPU: {cpu_usage_percent}% | Memory: {memory_usage_gb} / {total_cluster_memory} Gb ({cpu_memory_usage_percent}%)"
```
jlewitt1, line 282:
```suggestion
            cpu_usage_summary = f"{DOUBLE_SPACE_UNICODE}CPU: {cpu_usage_percent}% | Memory: {memory_usage_gb} / {total_cluster_memory} Gb ({cpu_memory_usage_percent}%)"
```

runhouse/main.py
```diff
@@ -679,6 +303,150 @@ def cluster_list(
...
+    Example:
+
+        ``$ runhouse cluster down /sashab/rh-basic-cpu``
+    """
+    current_cluster = get_cluster_or_local(cluster_name)
+
+    try:
+        current_cluster.teardown_and_delete() if remove_configs else current_cluster.teardown()
+    except ValueError:
+        console.print("Cluster is not saved in den, could not bring it down.")
```
jlewitt1, line 396:
```suggestion
        console.print("Cluster is not saved in Den, could not bring it down.")
```

runhouse/main.py
```diff
@@ -679,6 +303,150 @@ def cluster_list(
...
+        None,
+        help="Name of cluster to load the logs from. If not specified, loads the logs of the local cluster.",
+    ),
+    since: Optional[LogsSince] = typer.Option(
+        LogsSince.five.value,
+        "--since",
+        help="Get the logs from the provided timeframe (in minutes).",
+    ),
+):
+    """Load the logs of the Runhouse daemon running on a cluster.
```
jlewitt1, line 421: i think we should be consistent in how we refer to the server / daemon to avoid any confusion. For the server commands we are calling it "server", so maybe we do the same thing for the cluster commands. wdyt?

runhouse/cli_utils.py
BelSasha: It was updated in the server cli commands revamp PR, updating here as well.

runhouse/main.py
BelSasha: yeah good point

runhouse/main.py
```diff
@@ -679,6 +303,196 @@ def cluster_list(
...
+            proceed = typer.prompt(f"Terminating: {cluster_name}. Proceed? [Y/n]")
+        elif remove_all:
+            proceed = typer.prompt(
+                "Terminating all running clusters saved in Den. Proceed? [Y/n]"
+            )
+        else:
+            console.print("Cannot determine which cluster to terminate, aborting.")
+            raise typer.Exit(1)
+
+        if proceed.lower() == "n":
```
jlewitt1, line 411:
```suggestion
        if proceed.lower() in ["n", "no"]:
```

runhouse/main.py
```diff
@@ -679,6 +303,196 @@ def cluster_list(
...
+    """Terminate cluster if it is not down. No-op if cluster is already down.
+    This only applies to on-demand clusters, and has no effect on self-managed clusters.
+
+    Example:
+
+        ``$ runhouse cluster down /sashab/rh-basic-cpu``
+    """
+    if not confirm_deletion:
+        if cluster_name:
+            proceed = typer.prompt(f"Terminating: {cluster_name}. Proceed? [Y/n]")
```
jlewitt1, line 402: if `cluster_name` is not provided this looks a bit off: 
```
Terminating: <typer.models.ArgumentInfo object at 0x118d4cc40>. Proceed? [Y/n]:
```

runhouse/main.py
```diff
@@ -679,6 +303,196 @@ def cluster_list(
...
+            console.print("No running clusters saved in Den")
+            raise typer.Exit(0)
+
+        terminated_clusters: int = 0
+        for running_cluster in running_den_clusters:
+            try:
+                current_cluster = rh.cluster(
+                    name=f'/{rns_client.username}/{running_cluster.get("Name")}'
+                )
+                current_cluster.teardown_and_delete() if remove_configs else current_cluster.teardown()
```
jlewitt1, line 430: we can only call `teardown` and `teardown_and_delete` if they are on-demand clusters, we should check that first otherwise we'll get errors:

```
AttributeError: 'Cluster' object has no attribute 'teardown'
```

runhouse/main.py
```diff
@@ -679,6 +303,196 @@ def cluster_list(
...
+        else:
+            console.print("Cannot determine which cluster to terminate, aborting.")
+            raise typer.Exit(1)
+
+        if proceed.lower() == "n":
+            console.print("Aborted!")
+            raise typer.Exit(0)
+
+    if remove_all:
+        running_den_clusters = Cluster.list(status=ClustersListStatus.running).get(
```
jlewitt1, line 416: im wondering again whether it makes sense to slightly change this API structure - seems odd that if you specify a specific status you still have to extract that status from the output. there's a couple ways we can change this, but maybe we make the output  a list of dicts and add a "status" field to each one, wdyt?

BelSasha: Yeah, we could make `Cluster.list` return just `den_clusters` and `sky_clusters` (live clusters that are not saved in Den), and remove "running_clusters". Both `den_clusters` and `sky_clusters` each will be a list of dicts representing clusters. The dicts we are returning already have a status field :) 
wdyt?

runhouse/main.py
```diff
@@ -679,6 +303,207 @@ def cluster_list(
...
+                continue
+        console.print(f"Successfully terminated {terminated_clusters} clusters.")
+        raise typer.Exit(0)
+
+    current_cluster = get_cluster_or_local(cluster_name)
+
+    try:
+        if isinstance(current_cluster, rh.OnDemandCluster):
+            current_cluster.teardown_and_delete() if remove_configs else current_cluster.teardown()
+        else:
```
jlewitt1, line 448: if it's not an ondemand cluster and `remove_configs` is specified, we can still do that by calling `rns_client.delete_configs(resource= current_cluster)`

runhouse/main.py
```diff
@@ -679,6 +303,207 @@ def cluster_list(
...
+            try:
+                current_cluster = rh.cluster(
+                    name=f'/{rns_client.username}/{running_cluster.get("Name")}',
+                    dryrun=True,
+                )
+                if isinstance(current_cluster, rh.OnDemandCluster):
+                    current_cluster.teardown_and_delete() if remove_configs else current_cluster.teardown()
+                    terminated_clusters += 1
+                else:
+                    console.print(
```
jlewitt1, line 435: similar note as below

runhouse/main.py
```diff
@@ -679,6 +303,207 @@ def cluster_list(
...
+
+        ``$ runhouse cluster up /sashab/rh-basic-cpu``
+
+    """
+    current_cluster = get_cluster_or_local(cluster_name)
+
+    try:
+        current_cluster.up_if_not()
+    except ValueError:
+        console.print("Cluster is not saved in den.")
```
jlewitt1, line 366:
```suggestion
        console.print("Cluster is not saved in Den.")
```

runhouse/main.py
```diff
@@ -679,6 +303,207 @@ def cluster_list(
...
+        "--remove-configs",
+        help="Whether to delete cluster config from Den (default: ``False``).",
+    ),
+    remove_all: bool = typer.Option(
+        False,
+        "-a",
+        "--all",
+        help="Whether to terminate all running clusters saved in Den (default: ``False``).",
+    ),
+    confirm_deletion: bool = typer.Option(
```
jlewitt1, line 389: i think this is more clear about the implications of providing -y, wdyt?
```suggestion
    force_deletion: bool = typer.Option(
```

runhouse/main.py
```diff
@@ -679,6 +303,207 @@ def cluster_list(
...
+            )
+    except ValueError:
+        console.print("Cluster is not saved in Den, could not bring it down.")
+        sky_live_clusters = get_all_sky_clusters()
+        cluster_name = (
+            cluster_name.split("/")[-1] if "/" in cluster_name else cluster_name
+        )
+        if cluster_name in sky_live_clusters:
+            console.print(
+                f"You can bring down the cluster by running [italic bold] sky down {cluster_name} -y"
```
jlewitt1, line 460: small nit, should prob recommend the base command
```suggestion
                f"You can bring down the cluster by running [italic bold] sky down {cluster_name}"
```

runhouse/main.py
```diff
@@ -679,6 +303,207 @@ def cluster_list(
...
+    current_cluster = get_cluster_or_local(cluster_name)
+
+    try:
+        if not current_cluster.is_up():
+            print_bring_cluster_up_msg(cluster_name=cluster_name)
+            return
+        current_cluster.keep_warm(mins=mins)
+    except ValueError:
+        console.print(f"{cluster_name} is not saved in Den.")
+        sky_live_clusters = get_all_sky_clusters()
```
jlewitt1, line 334: thoughts on providing a `refresh` param for the commands that call `sky.status()`?

BelSasha: hmmm, not sure if I'm in favor to add the refresh param...

I think we had some discussion about it and we decided that currently if there are live sky clusters, we'll print the following msg to the console: `There are X live sky clusters that are not saved in Den. For more information, please run [bold italic]sky status -r[/bold italic].` This way, we make the user explicitly run the sky status command with the refresh parameter, which strengthens the value of saving clusters to Den (since we're getting the live status "out of the box" without needing the refresh parameter).

Our main goal is to encourage users to rely on Den as much as possible. By providing the refresh parameter, we introduce a convenient and easy-to-use wrapper around sky CLI commands. This makes it more straightforward to use sky via Runhouse rather than using Den without saving clusters to it.

wdyt?

runhouse/main.py
```diff
@@ -631,30 +253,38 @@ def cluster_list(
...
 
-    if not all_clusters:
+    if not den_clusters:
         no_clusters_msg = f"No clusters found in Den for {rns_client.username}"
         if show_all or since or cluster_status:
             no_clusters_msg += " that match the provided filters"
         console.print(no_clusters_msg)
         if sky_clusters:
-            print_sky_clusters_msg(len(sky_clusters))
+            console.print(SKY_LIVE_CLUSTERS_MSG)
```
jlewitt1, line 282: this variable is pretty specific to `main.py`, i think we can move it out of constants and put here

jlewitt1 (approved):

PR #1276, opened 2024-09-23
separate autostop check into dedicated thread

jlewitt1:


jlewitt1:
* **#1276** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1276?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

BelSasha (approved):

runhouse/servers/cluster_servlet.py
```diff
@@ -330,6 +336,37 @@ async def acheck_cluster_logs(self, interval_size: int):
...
         return logs_den_resp, new_start_log_line, new_end_log_line
 
+    async def aperiodic_autostop_check(self):
+        """Periodically check the autostop of the cluster"""
+        while True:
+            should_update_autostop: bool = self.autostop_helper is not None
+            cluster_config = await self.aget_cluster_config()
+
+            # Note: for now using the same interval for both autostop and status checks
+            interval_size = cluster_config.get(
```
carolineechen, line 346: it is possible interval size in the config is -1 right? and thoughts on just hardcoding this to 60 (s)?

runhouse/servers/cluster_servlet.py
```diff
@@ -330,6 +336,37 @@ async def acheck_cluster_logs(self, interval_size: int):
...
+                "status_check_interval", DEFAULT_STATUS_CHECK_INTERVAL
+            )
+
+            if not should_update_autostop:
+                logger.debug(
+                    f"Not updating autostop, checking again in {interval_size} minutes"
```
carolineechen, line 352: I don't think this debug statement makes sense since `should_update_autostop` as determined currently shouldn't ever change across loops. it's just based on whether or not the cluster is an ondemand cluster.

also noting that `break` here breaks out of the while loop. should be `continue` if we intend to continue running the thread in the background

runhouse/servers/cluster_servlet.py
jlewitt1: yea good point, i think for now easiest to just hard code it until we have a good reason to make it configurable

runhouse/servers/cluster_servlet.py
```diff
@@ -330,6 +336,25 @@ async def acheck_cluster_logs(self, interval_size: int):
 
         return logs_den_resp, new_start_log_line, new_end_log_line
 
+    async def aperiodic_autostop_check(self):
+        """Periodically check the autostop of the cluster"""
+        interval_size = 60
+        while True:
+            should_update_autostop: bool = self.autostop_helper is not None
+            if not should_update_autostop:
```
carolineechen, line 344: what if we move this into the try as follows. this way, we still have the sleep for interval size apply to all cases.

```
try:
    if should_udpate_autostop:
          status, _ = await ...
          ...
except...
finally...
```

carolineechen (approved):

PR #1279, opened 2024-09-24
Update index.rst to solve weird formatting

py-rh:


carolineechen (commented):
there's an issue with the previous bullet points, where there should be a blank line between bullets of different indentations for markdown. I'm not 100% if that will fix the rendering issue but might be worth giving a try before switching completely to html?

ex 
```
* bullet 1

   * nested bullet
   * nested bullet

* bullet 2

   * nested bullet
```

py-rh:
> there's an issue with the previous bullet points, where there should be a blank line between bullets of different indentations for markdown. I'm not 100% if that will fix the rendering issue but might be worth giving a try before switching completely to html?
> 
> ex
> 
> ```
> * bullet 1
> 
>    * nested bullet
>    * nested bullet
> 
> * bullet 2
> 
>    * nested bullet
> ```

They say that ChatGPT will replace developers but they could never replace @carolineechen (I spent a long time debating ChatGPT about why it doesn't work - the extra space below fixed it!)

mkandler (approved):
✨

PR #1286, opened 2024-09-25
add autostop support as an env var

jlewitt1:


jlewitt1:
* **#1286** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1286?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

runhouse/resources/hardware/cluster.py
```diff
@@ -506,6 +506,11 @@ def _sync_default_env_to_cluster(self):
             self._default_env.add_env_var("disable_observability", "True")
             logger.info("Disabling observability on the cluster")
 
+        autostop_interval = os.getenv("RH_AUTOSTOP_INTERVAL")
+        if autostop_interval:
+            self._default_env.add_env_var("RH_AUTOSTOP_INTERVAL", autostop_interval)
+            logger.info(f"Setting cluster autostop to {autostop_interval} seconds")
+
```
carolineechen, line 513: I think we don't need to include this here if the main purpose is for testing and not actually a variable the user will set. We can just add this env var in the test setup instead.

carolineechen (approved):

PR #1300, opened 2024-09-29
server cli cmds revamp

BelSasha:


runhouse/main.py
```diff
@@ -783,11 +794,14 @@ def restart(
...
 
 
-@app.command()
+@server_app.command()
 def stop(
     stop_ray: bool = typer.Option(False, help="Stop the Ray runtime"),
     cleanup_actors: bool = typer.Option(True, help="Kill all Ray actors"),
 ):
+    """
+    Stopping Runhouse server on the current cluster.
```
jlewitt1, line 803: this doesn't have to be on the current cluster right? can we support this + the other server commands locally by specifying a cluster name?

BelSasha: Yeah. So the `restart` and `stop` are supported for both locally on the user laptop and on the cluster.

Regarding the start cmd - not sure how will it differ from restart if we support if both ways. I think we should discuss it further.

runhouse/main.py
```diff
@@ -784,12 +795,24 @@ def restart(
...
+@server_app.command()
 def stop(
+    name: Optional[str] = typer.Option(
+        None, help="A *saved* remote cluster object to restart."
+    ),
     stop_ray: bool = typer.Option(False, help="Stop the Ray runtime"),
     cleanup_actors: bool = typer.Option(True, help="Kill all Ray actors"),
 ):
+    """
+    Stopping Runhouse server on the current cluster.
```
jlewitt1, line 807: for consistency with the other server cmds:
```suggestion
    Stop the Runhouse server on the current cluster.
```

runhouse/main.py
```diff
@@ -807,6 +866,25 @@ def stop(
...
 
 
+@server_app.command("status")
+def server_status(
+    name: Optional[str] = typer.Argument(
+        None, help="A *saved* remote cluster object to restart."
+    ),
+):
+    """Check the HTTP or HTTPS server status on the cluster."""
+    logger.info("Checking the server status.")
```
jlewitt1, line 876: can we make this a debug

runhouse/main.py
```diff
@@ -846,6 +907,23 @@ def stop(
...
+        None, help="A *saved* remote cluster object to restart."
+    ),
+):
+    """Check the HTTP server status on the cluster."""
+    logger.debug("Checking the server status.")
+
+    current_cluster = get_cluster_or_local(cluster_name=cluster_name)
+    try:
+        current_cluster.client.check_server()
+        console.print("Server is up.")
```
jlewitt1, line 922: thoughts on something like this? 
```suggestion
        console.print("f"Server is up and running on host: {current_cluster.client.host} and port {current_cluster.client.port}.")
```

runhouse/main.py
```diff
@@ -719,12 +724,28 @@ def _start_server(
...
 
-@app.command()
-def start(
+###############################
+# Server CLI commands
+###############################
+
+
+@server_app.command("start")
+def server_start(
```
jlewitt1, line 733: wondering if we should handle more gracefully trying to start the server if it's already running - and instead direct the user to run `restart` - for example when running locally: 

```
INFO:__main__:Launching Runhouse API server with den_auth=False and on host=0.0.0.0 and use_https=False and port_arg=32300
INFO:     Started server process [41325]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
ERROR:    [Errno 48] error while attempting to bind on address ('0.0.0.0', 32300): address already in use
INFO:     Waiting for application shutdown.
INFO:     Application shutdown complete.
```

runhouse/main.py
```diff
@@ -892,6 +953,27 @@ def stop(
         subprocess.run(RAY_KILL_CMD, shell=True)
 
 
+@server_app.command("status")
+def server_status(
+    cluster_name: Optional[str] = typer.Argument(
+        None, help="A *saved* remote cluster object to restart."
```
jlewitt1, line 959:
```suggestion
        None, help="A *saved* remote cluster object to restart. If not specified will check the locally running server."
```

runhouse/main.py
```diff
@@ -869,12 +919,23 @@ def restart(
     )
 
 
-@app.command()
-def stop(
+@server_app.command("stop")
+def server_stop(
+    cluster_name: Optional[str] = typer.Option(
+        None, help="A *saved* remote cluster object to restart."
```
jlewitt1, line 925:
```suggestion
        None, help="A *saved* remote cluster object to restart. If not specified will stop the locally running server."
```

runhouse/main.py
```diff
@@ -719,12 +724,28 @@ def _start_server(
...
-def start(
+###############################
+# Server CLI commands
+###############################
+
+
+@server_app.command("start")
+def server_start(
+    cluster_name: Optional[str] = typer.Argument(
+        None, help="A *saved* remote cluster object to restart."
```
jlewitt1, line 735:
```suggestion
        None, help="A *saved* remote cluster to start a Runhouse server. If not specified will start a local server."
```

runhouse/main.py
```diff
@@ -768,17 +811,24 @@ def start(
...
     )
 
 
-@app.command()
-def restart(
-    name: str = typer.Option(None, help="A *saved* remote cluster object to restart."),
+@server_app.command("restart")
+def server_restart(
+    cluster_name: str = typer.Option(
+        None, help="A *saved* remote cluster object to restart."
```
jlewitt1, line 830:
```suggestion
        None, help="A *saved* remote cluster to restart the Runhouse server. If not specified will restart the locally running server."
```

runhouse/main.py
```diff
@@ -64,6 +64,15 @@
     help=f"Cluster related CLI commands. For more info run {ITALIC_BOLD}runhouse cluster --help{RESET_FORMAT}"
 )
 
+app.add_typer(cluster_app, name="cluster")
+
+# creating a cluster app, so we could create subcommands of cluster (i.e runhouse cluster list).
+# Register it with the main runhouse application
+server_app = typer.Typer(
+    help=f"Runhouse server related CLI commands. For more info run {ITALIC_BOLD}runhouse server --help{RESET_FORMAT}"
```
jlewitt1, line 72: The --help message seems duplicative since the "--help" shows up automatically, unless there's another reason to include it? (same for `cluster` too btw)

```suggestion
    help="Runhouse server related CLI commands."
```

runhouse/main.py
```diff
@@ -730,12 +735,28 @@ def _start_server(
...
+    screen: bool = typer.Option(
+        True,
+        help="Start the server in a screen. Only relevant when restarting locally.",
+    ),
     nohup: bool = typer.Option(
-        False, help="Start the server in a nohup if screen is not available"
+        True,
+        help="Start the server in a nohup if screen is not available. Only relevant when restarting locally.",
+    ),
+    resync_rh: bool = typer.Option(
```
jlewitt1, line 757: dont think this is being used

runhouse/main.py
```diff
@@ -903,6 +964,27 @@ def stop(
...
+@server_app.command("status")
+def server_status(
+    cluster_name: Optional[str] = typer.Argument(
+        None, help="A *saved* remote cluster object to restart."
+    ),
+):
+    """Check the HTTP server status on the cluster."""
+    logger.debug("Checking the server status.")
+
+    current_cluster = get_cluster_or_local(cluster_name=cluster_name)
```
jlewitt1, line 976: this command won't work if running `runhouse server status` on a cluster, which i think we should also support

BelSasha: looking good to me, would you mind to verify it on your laptop locally? 
image

runhouse/main.py
```diff
@@ -767,14 +796,42 @@ def start(
...
+    ),
 ):
-    """Start the HTTP or HTTPS server on the cluster."""
+    """Start the HTTP server on the cluster."""
+
+    # If server is already up, ask the user to restart the server instead.
+    server_status_cmd = "runhouse server status"
+    if cluster_name:
+        server_status_cmd = server_status_cmd + f" {cluster_name}"
+    is_server_up = subprocess.run(
```
jlewitt1, line 820: this is only needed for checking local servers no? meaning if cluster name is specificed we can ignore this check

runhouse/main.py
```diff
@@ -767,14 +796,42 @@ def start(
...
+        server_status_cmd, shell=True, capture_output=True, text=True
+    )
+
+    if is_server_up.returncode == 0 and "Server is up" in is_server_up.stdout:
+        console.print(
+            "Runhouse Server is already running. To update it, please run [italic bold]runhouse server restart [reset]with the relevant parameters."
```
jlewitt1, line 826:
```suggestion
            "Local Runhouse server is already running. To restart it, please run [italic bold]runhouse server restart [reset]with the relevant parameters."
```

runhouse/main.py
jlewitt1: yep good for me now too

runhouse/main.py
jlewitt1, line 820: was confused at first bc thought this was a boolean, but really its a subprocess object

```suggestion
    result = subprocess.run(
```

runhouse/main.py
```diff
@@ -767,14 +796,42 @@ def start(
...
+    from_python: bool = typer.Option(
+        False,
+        help="Whether HTTP server started from inside a Python call rather than CLI.",
+    ),
 ):
-    """Start the HTTP or HTTPS server on the cluster."""
+    """Start the HTTP server on the cluster."""
+
+    # If server is already up, ask the user to restart the server instead.
+    server_status_cmd = "runhouse server status"
```
jlewitt1, line 817: is there a way we could do this pythonically instead?

runhouse/main.py
BelSasha: hmm yeah, but in a case the user provides a name, and the server is up, don't we want that the server will be restarted anyway? (Don't have a strong opinion about that)

docs/installation.rst
```diff
@@ -155,7 +155,7 @@ you may find the following `SkyPilot commands <https://skypilot.readthedocs.io/e
 - ``sky status`` - Displays a list of all your clusters by name.
 - ``sky stop <NAME>`` - Stops an instance. This will not fully tear down the cluster in case you need to restart it.
 - ``sky down <NAME>`` - Fully tears down the instance. Best for when you are done using a cluster entirely.
-- ``ssh <NAME>`` - Easily SSH into your cluster. Here you can further examine Runhouse's server with ``runhouse status``.
+- ``ssh <NAME>`` - Easily SSH into your cluster. Here you can further examine Runhouse's server with ``runhouse cluster status``.
```
jlewitt1, line 158: shouldnt this be `runhouse cluster ssh`?

runhouse/resources/hardware/cluster.py
```diff
@@ -502,6 +503,13 @@ def is_up(self) -> bool:
         """
         return self.on_this_cluster() or self._ping()
 
+    def is_server_up(self) -> bool:
```
jlewitt1, line 506: can we make this private? not sure we need to expose this one to users

runhouse/resources/hardware/cluster.py
```diff
@@ -502,6 +503,13 @@ def is_up(self) -> bool:
         """
         return self.on_this_cluster() or self._ping()
 
+    def _is_server_up(self) -> bool:
```
jlewitt1, line 506: how come `is_up` isn't enough here?

BelSasha: because is_up checking if self.address is not None. But sometimes self.address might be equal to an ip ( == the cluster is live on the cloud), but the runhouse server is down.

jlewitt1 (approved):

PR #1311, opened 2024-10-07
update valid cluster status enum

BelSasha:


runhouse/resources/hardware/utils.py
```diff
@@ -49,7 +49,6 @@ class ResourceServerStatus(str, Enum):
     internal_server_error = "internal_server_error"
     runhouse_daemon_down = "runhouse_daemon_down"
     invalid_url = "invalid_url"
-    local_cluster = "local_cluster"
```
jlewitt1, line 52: we might want to have some BC here on `get_status_color`, otherwise will get an attribute error. will open a PR shortly which includes this + some other minor formatting stuff

PR #1318, opened 2024-10-09
add runhouse base image

jlewitt1:


docker/base/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+        echo "Installing latest version"; \
+        pip install --no-cache-dir "runhouse[sky]"; \
+      else \
+        echo "Installing runhouse==${RUNHOUSE_VERSION}"; \
+        pip install --no-cache-dir "runhouse[sky]==${RUNHOUSE_VERSION}"; \
+      fi; \
+    fi
+
+# Keep the container running, allowing SkyPilot to connect and run commands inside the container
+CMD ["bash"]
```
dongreenberg, line 48: Is this really necessary? Feel like it's not standard

jlewitt1: yea fair point, it's not so standard. but we do need to keep the container up and not exit while Sky is doing the provisioning / setup step. prob better to run supervisord as the main process

dongreenberg: I don't really understand - isn't the Skypilot work happening after the container already built? What happens if we try to launch a docker cluster without any line here?

jlewitt1: Without that line (or similar) the container would exit since no process is keeping it alive. It does seem like sky is doing some work to run commands inside the container after it has started - snippet below from the provisioning logs: 

```
Error response from daemon: Container 77772152c4654088ac87318f08620c24aba7aef619fa744ff72b99668971d907 is not running

I 10-09 19:33:56 instance_setup.py:88] _initialize_docker: Retrying in 1.3 seconds, due to Command docker exec sky_container /bin/bash -c 'bash --login -c -i '"'"'source ~/.bashrc; export OMP_NUM_THREADS=1 PYTHONWARNINGS=ignore && (sudo service ssh start)'"'"' failed with return code 1.

I 10-09 19:33:56 instance_setup.py:88] Failed to run docker setup commands

D 10-09 19:33:57 docker_utils.py:154] + command -v docker || echo 'NoExist'

/usr/bin/docker

D 10-09 19:33:58 docker_utils.py:154] + docker inspect -f "{{.State.Running}}" sky_container || true

false

D 10-09 19:33:58 docker_utils.py:154] + docker start sky_container

sky_container

D 10-09 19:33:59 docker_utils.py:154] + docker exec sky_container /bin/bash -c 'bash --login -c -i '"'"'source ~/.bashrc; export OMP_NUM_THREADS=1 PYTHONWARNINGS=ignore && (sudo service ssh start)'"'"'

Error response from daemon: Container 77772152c4654088ac87318f08620c24aba7aef619fa744ff72b99668971d907 is not running

```

PR #1338, opened 2024-10-22
Split cluster keys and ssh properties

carolineechen:


tests/test_resources/test_clusters/test_cluster.py
```diff
@@ -345,12 +351,18 @@ def test_sharing(self, cluster, friend_account_logged_in_docker_cluster_pk_ssh):
         resource_class_name = cluster.config().get("resource_type").capitalize()
         config = cluster.config()
 
+        expected_creds = (
+            "ssh-sky-key"
+            if isinstance(cluster._creds, SSHSecret)
+            else f'{config["name"]}-ssh-secret'
```
jlewitt1, line 357: how come we're not hard coding the ssh-secret name?  (or is this just for testing purposes?)

runhouse/resources/hardware/cluster.py
```diff
@@ -271,6 +274,66 @@ def delete_configs(self, delete_creds: bool = False):
 
         super().delete_configs()
 
+    def _setup_creds(self, ssh_creds: Union[Dict, "Secret"]):
```
jlewitt1, line 277: what about supporting a string here representing the rns address of an existing ssh creds resource?

jlewitt1 (approved):

tests/test_resources/test_clusters/test_cluster.py
carolineechen: just for testing - if the secret has a priv key path itll be the SkySecret `ssh-sky-key`, but if it has a password without a priv key then itll just be a generic secret type, but tied and named specifically to the cluster

runhouse/resources/hardware/cluster.py
carolineechen: I think that makes sense, will add support for it

PR #1366, opened 2024-11-01
Resnet and small den edits

py-rh:


py-rh:
Will do some cleanup for comments and commentary on using with Runhouse, but wanted to open PR for the overall resnet example for review and get feedback.

PR #1393, opened 2024-11-11
Update gcp_secret.py to remove unnecessary default env vars

py-rh:
A) The bug is we try env_vars = {env_vars[k]: self.values[k] for k in self.values} but
self.values is {'account': '', 'client_id': '', 'client_secret': '', 'quota_project_id': '', 'refresh_token': '', 'type': '', 'universe_domain': ''}

while env_vars is {'client_id': 'CLIENT_ID', 'client_secret': 'CLIENT_SECRET'} (from gcp_secret.py _DEFAULT_ENV_VARS )

Hence we get KeyError: 'account'

B) Setting these env vars aren't actually necessary at all for the GCP secrets to work. We are already correctly writing the credentials JSON to ~/.config. The only env variable we might want to set for gcloud auth is the path to the credentials config, but that's not done here.

### DIFF
--- a/runhouse/resources/secrets/provider_secrets/gcp_secret.py
+++ b/runhouse/resources/secrets/provider_secrets/gcp_secret.py
@@ -17,10 +17,6 @@ class GCPSecret(ProviderSecret):
 
     _PROVIDER = "gcp"
     _DEFAULT_CREDENTIALS_PATH = "~/.config/gcloud/application_default_credentials.json"
-    _DEFAULT_ENV_VARS = {
-        "client_id": "CLIENT_ID",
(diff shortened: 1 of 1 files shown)


py-rh:
Curious if we want to retain defaults, what the right defaults should be.

AWS uses env variables for key/secret but GCP actively does not; we're more likely picking up random env_vars named "client_id" than someone setting them just for Runhouse (it wouldn't be automatically set or used by gcloud cli)

Service accounts are actually appropriate to authenticate via json key file, but they have a different set of fields from user account json (it has private_key / private_key_id) so default behavior might be for that

PR #1404, opened 2024-11-13
Fix restart_ray for den launched clusters

mkandler:
Sky status may not be reachable for cluster launched with Den, so fallback to using IPs set directly on the cluster when restarting Ray

### DIFF
--- a/runhouse/resources/hardware/on_demand_cluster.py
+++ b/runhouse/resources/hardware/on_demand_cluster.py
@@ -363,9 +363,12 @@ class OnDemandCluster(Cluster):
         # Find the internal IP corresponding to the public_head_ip and the rest are workers
         internal_head_ip = None
         worker_ips = []
-        stable_internal_external_ips = self._sky_status()[
-            "handle"
-        ].stable_internal_external_ips
+        sky_status = self._sky_status()
+        stable_internal_external_ips = (
+            sky_status["handle"].stable_internal_external_ips
+            if sky_status
+            else self.stable_internal_external_ips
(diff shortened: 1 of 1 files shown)


mkandler:
* **#1404** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1404?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -363,9 +363,12 @@ def _start_ray_workers(self, ray_port, env):
         # Find the internal IP corresponding to the public_head_ip and the rest are workers
         internal_head_ip = None
         worker_ips = []
-        stable_internal_external_ips = self._sky_status()[
-            "handle"
-        ].stable_internal_external_ips
+        sky_status = self._sky_status()
```
carolineechen, line 366: wondering if we can get rid of all this and just do `for internal, external in self.stable_internal_external_ips` below, since it should already be set by the time `_start_ray_workers` is called in restart server.

not sure if there's any edge case though so fine to keep this for the release and test later on if we can skip this extra sky status/populating

carolineechen (approved):
wondering if we can get rid of all this and just do `for internal, external in self.stable_internal_external_ips`, since it should already be set by the time _start_ray_workers is called in restart server.

not sure if there's any edge case though so fine to keep this for the release and test later on if we can skip this extra sky status/populating

PR #1418, opened 2024-11-14
Update cluster list tests

carolineechen:


carolineechen:
* **#1418** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1418?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tests/test_resources/test_clusters/test_cluster.py
```diff
@@ -1007,23 +1006,15 @@ def test_cluster_list_pythonic(self, cluster):
...
-                    if den_cluster.get("Status") == "running"
-                ]
-                if all_clusters
-                else {}
-            )
-            assert len(all_clusters) > 0
-            assert len(running_clusters) > 0
-            assert len(running_clusters) == len(
-                all_clusters
+            all_clusters = Cluster.list(show_all=True).get("den_clusters", {})
```
BelSasha, line 1009: I think that this we should not remove test_cluster_list_pythonic (maybe just re-name it if necessary), because it tests the default behaviour of Cluster.list().

Seems to me we are going to show by default only part of the running clusters by default (based on @carolineechen comments on Josh's [PR](https://github.com/run-house/runhouse/pull/1431)), so IMO we need to have two separate tests:
1. Default behavior of Cluster.list()
2. Tests the behaviour of Cluster.list(show_all=true)

wdyt?

tests/test_resources/test_clusters/test_cluster.py
```diff
@@ -1035,7 +1026,7 @@ def test_cluster_list_pythonic(self, cluster):
 
     @pytest.mark.level("local")
     @pytest.mark.clustertest
-    def test_cluster_list_all_pythonic(self, cluster):
```
BelSasha, line 1038: Similar to the note above, don't think we need to delete this test

tests/test_resources/test_clusters/test_cluster.py
carolineechen: yup makes sense, there was some overlap in my tests but separated to make it clearer. for the pythonic version, test for
* default behavior - contains only running clusters, includes current test cluster
* show_all - contains running + terminated clusters, includes current test cluster
* filter by running, terminated
* filter by time

jlewitt1 (approved):

PR #1422, opened 2024-11-14
Gracefully handle empty IP lists

mkandler:
IP lists can be None in clusters but often assumed that it is an array

### DIFF
--- a/runhouse/resources/hardware/cluster.py
+++ b/runhouse/resources/hardware/cluster.py
@@ -135,7 +135,9 @@ class Cluster(Resource):
 
     @property
     def head_ip(self):
-        return self.ips[0] if isinstance(self.ips, List) else None
+        return (
+            self.ips[0] if (isinstance(self.ips, List) and len(self.ips) > 0) else None
+        )
 
     @property
     def address(self):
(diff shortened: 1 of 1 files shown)


mkandler:
* **#1422** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1422?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

mkandler:
@jlewitt1 cc: this relates to our discussion the other day of setting IPs to`[None]`

Currently running into bugs when launching because of this.

runhouse/resources/hardware/cluster.py
```diff
@@ -637,7 +637,7 @@ def _sync_default_env_to_cluster(self):
         env_vars = _process_env_vars(self._default_env.env_vars)
         setup_cmds = self._default_env.setup_cmds
 
-        for node in self.ips:
+        for node in self.ips or []:
```
jlewitt1, line 640: any issue with just doing this once in the constructor: `self.ips = ips or []`?

also i think with this change we'd also have to update the `head_node` property to avoid an IndexError

mkandler: I think we could change the constructor but then we'd have to enforce that self.ips is never set to `None` in other places (which it can be). Feels like we should do this as a safety measure regardless?

mkandler: and the `head_node` property is fine:
```
@property
def head_ip(self):
    return self.ips[0] if isinstance(self.ips, List) else None
```

mkandler: Actually nvm, that method was problematic too, so it's updated now!

jlewitt1 (approved):

PR #1427, opened 2024-11-15
Remove stopping of autostop when notebook is launched

py-rh:
Also it removes dependence on skypilot for stopping of autostop -- IDK maybe we want to update pausing of autostop to reflect on Den but it might be too complex. 

Just don't launch a cluster with autostop? This is the only place we use it I think.


carolineechen:
pause_autostop w/o local sky works after #1469, we can revisit notebook autostop behavior in the future

PR #1435, opened 2024-11-17
change default autostop to 60 mins

jlewitt1:


jlewitt1:
* **#1435** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1435?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1435?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

runhouse/resources/hardware/cluster_factory.py
```diff
@@ -395,6 +395,12 @@ def ondemand_cluster(
             f"local ~/.rh/config.yaml."
         )
 
+    if autostop_mins == -1:
```
carolineechen, line 398: don't know if we need this if we change the default to 60 min, bc then this means setting `-1` is intentional?

runhouse/rns/login.py
```diff
@@ -146,6 +146,15 @@ def login(
         # via configs.token = token
         configs.set("token", token)
 
+    autostop_mins = configs.defaults_cache.get("default_autostop")
+    if autostop_mins is None:
```
carolineechen, line 150: should we download the config to local first before checking for and updating autostop defaults? for the case of logging in on a new device

runhouse/rns/login.py
```diff
@@ -146,6 +146,15 @@ def login(
...
+        new_autostop_mins = typer.prompt(
+            "Set the number of minutes to auto-terminate clusters after a period of inactivity, "
+            "or press `Enter` to set to 60 minutes. Set to -1 to disable autostop entirely.",
```
carolineechen, line 153: just a suggestion, I feel like could be worded more elegantly but it's kinda tricky. also not sure our stance on static clusters rn either and whether this could be confusing if that's still a use case? 

```suggestion
            "Set the default number of minutes of inactivity after which to auto-terminate clusters. "
            "Press `Enter` to set to 60 minutes, or `-1` to disable autostop entirely.",
```

docs/architecture.rst
```diff
@@ -159,7 +159,8 @@ When a remote object is no longer needed, it can be deallocated from
 the remote compute by calling ``cluster.delete(obj_name)``. This will remove the object from the key-value store and
 free up the memory on the worker. A worker process can similarly be terminated with ``cluster.delete(worker_name)``,
 terminating its activities and freeing its memory. An on-demand cluster can be terminated with ``cluster.teardown()``,
-or by setting its ``autostop_mins``, which will auto-terminate it after a period of inactivity.
+or by setting its ``autostop_mins``, which will auto-terminate it after a period of inactivity. The default autostop
```
carolineechen, line 162:
```suggestion
or by setting its ``autostop_mins``, which will auto-terminate it after that period of inactivity. The default autostop
```

runhouse/rns/login.py
jlewitt1: i guess if we move this check just a little further down we could do that? since if `download_config` is set to True thats what we'd do anyway

runhouse/rns/login.py
jlewitt1: yea couldn't really settle on the best way of wording wording it, and you're right it's really only relevant for on-demand clusters, so thinking we just need to make that more clear in the message

carolineechen (approved):

PR #1436, opened 2024-11-17
Include launcher type in cluster factory check

carolineechen:


carolineechen:
* **#1436** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1436?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1436?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

runhouse/resources/hardware/cluster_factory.py
```diff
@@ -108,6 +108,7 @@ def cluster(
             domain=domain,
             den_auth=den_auth,
             default_env=default_env,
+            launcher_type=launcher_type,
```
jlewitt1, line 111: curious what the motivation is for adding it here? if the user gives a new `launcher_type` value but with a name that is already saved in Den, this would prevent us from reloading an already live cluster and lead to some weird behavior where we try to relaunch it

carolineechen: that's true. at the same time though, if the user gives a new `launcher_type` for a saved cluster, the cluster is reloaded and will return with the old `launcher_type` that was saved. in that case we'd have to add something like `c.launcher_type = launcher_type or c.launcher_type` after loading from name. are there any other properties that fall under this category?

also I'd like to rewrite cluster factories to be a bit more intuitive / less convoluted at some point, we've run into small issues like this before, but will probably tackle this when there's less higher priority things going on

jlewitt1: hmm yea i think this can get a little tricky. one thought - what if we try to reload from name initially, ignoring any other params provided. if the cluster is already saved and conflicting param values were provided, we throw an error instructing the user to teardown and delete that cluster if they want to relaunch with that same name?

carolineechen:
#1575 1575

PR #1441, opened 2024-11-18
_update_from_den_response bugfix

BelSasha:


BelSasha:
* **#1441** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1441?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1441?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

runhouse/resources/hardware/launcher_utils.py
```diff
@@ -180,7 +180,7 @@ def _update_from_den_response(cls, cluster, config: dict):
             "ssh_properties",
             "client_port",
         ]:
-            value = config.get(attribute)
+            value = config.get("cluster_config").get(attribute)
```
jlewitt1, line 183: we should prob leave this as is, and instead parse the response before calling `_update_from_den_response`. is the issue that the config being passed in has two keys (`cluster_config` and `launched_config`)? if so we can just set config=data["cluster_config"] in the `up` method

runhouse/resources/hardware/launcher_utils.py
```diff
@@ -208,7 +208,9 @@ def up(cls, cluster, verbose: bool = True, force: bool = False):
                 payload=payload,
                 cluster_name=payload["cluster_config"].get("name"),
             )
-            cls._update_from_den_response(cluster=cluster, config=data)
+            cls._update_from_den_response(
```
jlewitt1, line 211: actually don't we call this from two places? this addresses the streaming use case, but i imagine the same would apply if verbose=False?

BelSasha: if verbose=False, the data dictionary we receive equals to the cluster_config. It is not a dict with two key. Is it expected or it is also might be related to the `is_up` issues?

jlewitt1:
to be fixed in Den

PR #1443, opened 2024-11-18
remove updates to cluster configs post den launch

jlewitt1:
If we save the cluster object properly in the den launcher immediately after launching, we can save the need to have to update any SSH creds or launched property fields. 

Have a launcher PR up which saves the cluster to Den immediately after it is launched, saving us the need to reload or resave those attributes on the client

### DIFF
--- a/runhouse/resources/hardware/launcher_utils.py
+++ b/runhouse/resources/hardware/launcher_utils.py
@@ -10,7 +10,7 @@ from runhouse.constants import DOCKER_LOGIN_ENV_VARS
 from runhouse.globals import configs, rns_client
 from runhouse.logger import get_logger
 from runhouse.resources.hardware.utils import SSEClient
-from runhouse.rns.utils.api import generate_ssh_keys, load_resp_content, read_resp_data
+from runhouse.rns.utils.api import generate_ssh_keys, load_resp_content
 from runhouse.utils import Spinner
 
 logger = get_logger(__name__)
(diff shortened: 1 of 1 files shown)


jlewitt1:
* **#1443** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1443?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1443?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

runhouse/resources/hardware/launcher_utils.py
```diff
@@ -222,9 +200,7 @@ def up(cls, cluster, verbose: bool = True, force: bool = False):
                 f"Received [{resp.status_code}] from Den POST '{cls.LAUNCH_URL}': Failed to "
                 f"launch cluster: {load_resp_content(resp)}"
             )
-        data = read_resp_data(resp)
         logger.info("Successfully launched cluster")
-        cls._update_from_den_response(cluster=cluster, config=data)
```
mkandler, line 227: This is essential to be able to use the cluster after running this method.

mkandler:
I don't think we can remove this code if you want to be able to work with the cluster after launching. We need to return and set the configuration from the Launcher so that the cluster can be used afterwards. E.g. `cluster.restart_server()` or `.to(cluster)`

PR #1455, opened 2024-11-21
kill servlets in case of an error in cluster status / autostop checks

BelSasha:
CI failures match the failures on main (related to image and default_env changes)

Open items / questions: 
~1. Sending the config.yaml using http request after the cluster servlet is initiated caused the periodic status loops to fail (open [PR](https://github.com/run-house/runhouse/pull/1486)). Reverted this change as a temp solution, just to check the logic of this PR. Once the config.yaml sync will work properly, the temp change will be reverted.~
~2. Currently, if there is an error in the cluster servlet loops, and the log level is not debug, we decrease delay from 60 min to 5 min. Other solution might be track the error, and if it occurs more than twice in consecutive runs then raise the exception and kill the servlet?~
3. What to do with the logs themselves from the servelet? They are a bit messy since they are running inside a ray actor.


runhouse/servers/cluster_servlet.py
```diff
@@ -711,7 +733,13 @@ async def send_cluster_logs_to_den(
...
             prev_end_log_line = 0
 
-        logs_to_den = "\n".join(latest_logs[prev_end_log_line:])
+        # len(latest_logs) may equal to 0 if we are in the process of setting up the cluster servlet.
+        # In this case the server.log might nov have been created yet, therefore self._get_logs() returns an empty string ("").
+        logs_to_den = (
+            "\n".join(latest_logs[prev_end_log_line:])
+            if len(latest_logs) > 0
+            else latest_logs
+        )
```
BelSasha, line 742: Open question about the implementation which addresses the following use case:
When we first initialize the ClusterServlet and the cluster, there is a point where the ~/.rh/server.log file has not been created yet, but the cluster_checks loop is initialized.

At this point:

- The cluster_checks loop attempts to read the logs from the server.log file.
- The log file is not found.
- An exception is caught by the cluster_checks thread.
- The ClusterServlet is terminated, causing some server CI tests to fail.

These tests were not failing previously because we did not terminate the ClusterServlet when this error occurred.

**Current fix**
If the server.log file does not exist, the fix returns an empty string for the logs.
_Pros_:

- Resolves the issue where the server.log file is not created before the first run of the cluster_checks loop.

_Cons:_

- If the server.log file fails to be created later, we will not catch this issue.

**Alternative Solution**
Introduce a short delay (a few seconds) between the creation of the cluster_checks loop and its execution.

_Pros:_

- Ensures the cluster_checks loop starts only after the cluster setup is complete and the server.log file is created.

_Cons:_

- The initial cluster status and log information will be delayed.
- This approach was previously implemented but was removed to eliminate the delay between the loop's creation and execution.

Don't have a strong opinion about any of the solutions, and open to any other solution suggestions.

runhouse/servers/cluster_servlet.py
```diff
@@ -376,8 +379,18 @@ async def aperiodic_autostop_check(self):
                     logger.debug("Successfully updated autostop")
 
             except Exception as e:
+
                 logger.error(f"Autostop check has failed: {e}")
 
+                # killing the cluster servlet only if log level is debug: this indicates that the cluster is created
+                # during test, or by a runhouse team member
+                log_level = os.getenv("RH_LOG_LEVEL")
+                if log_level.lower() == "debug":
```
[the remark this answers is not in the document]
BelSasha: I don't think so, because I added a fix for syncing the `RH_LOG_LEVEL` to the cluster: if there is no `RH_LOG_LEVEL` local env var, we sync the default log level (currently synced to 'INFO'). I tested this logic, and it seems to be working. Nevertheless, I think that we could use here the `DEFAULT_LOG_LEVEL` as well, just to double check that log_level.lower() does not causing a breakage. 

wdyt?

runhouse/resources/hardware/cluster.py
```diff
@@ -602,7 +608,8 @@ def _sync_image_to_cluster(self):
         use the HTTP client.
         """
         env_vars = {}
-        log_level = os.getenv("RH_LOG_LEVEL")
+        # In case the local env var RH_LOG_LEVEL is None, we set the RH_LOG_LEVEL on the cluster to be the DEFAULT_LOG_LEVEL
+        log_level = os.getenv("RH_LOG_LEVEL") or DEFAULT_LOG_LEVEL
```
BelSasha, line 612: @carolineechen any objections / thoughts?

carolineechen: this looks fine

runhouse/servers/cluster_servlet.py
```diff
@@ -693,6 +730,10 @@ def status(self, send_to_den: bool = False):
     # Save cluster logs to Den
     ##############################################
     def _get_logs(self):
+        from pathlib import Path
+
+        if not Path(SERVER_LOGFILE).exists():
```
jlewitt1, line 735: is this expected to happen just initially? if the server doesn't exist at a certain point shouldn't we throw an exception

BelSasha: yeah, the catchy thing is to determine if this is the first run of the logs loop or not. 
We could set a flag during init, and change it after the first run, even if the run failed (will change it in the finally block).

runhouse/resources/hardware/cluster.py
```diff
@@ -776,11 +783,15 @@ def get(self, key: str, default: Any = None, remote=False):
                 remote=remote,
                 system=self,
             )
+            return res
         except KeyError as e:
             if default == KeyError:
                 raise e
             return default
-        return res
+        except ConnectionError:
```
jlewitt1, line 791: are these errors always guaranteed to be  of type `ConnectionError`?

runhouse/servers/cluster_servlet.py
```diff
@@ -91,6 +93,11 @@ async def __init__(
         # will be used when self.gpu_metrics will be updated by different threads.
         self.lock = threading.Lock()
 
+        # will be used in the periodic loop which sends cluster logs to den. Since we are saving logs to the server.log
+        # file asynchronously, there might be a use case where we want to send logs to den but the server.log has not
+        # been created yet, since it is a freshly initialized cluster servlet. In this case, we will return an empty string.
+        self.is_freshly_initialized = True
```
jlewitt1, line 99: maybe `is_log_file_ready`? freshly initialized was a little unclear to me at first

runhouse/resources/hardware/cluster.py
BelSasha: I think so, yes. Because we are sending an HTTP request to the http server on the cluster servlet, but we can't connect to it because the cluster servlet is killed.

tests/utils.py
```diff
@@ -36,14 +37,25 @@ def get_ray_servlet_and_obj_store(env_name):
...
-    if cluster_config:
-        ObjStore.call_actor_method(
-            cluster_servlet, "aset_cluster_config", cluster_config
-        )
+    import logging
+
+    import ray
+
+    # need to initialize ray so the cluster servlet will be initialized.
+    ray.init(
```
jlewitt1, line 50: confused how this was working before this change?

BelSasha: don't think we used `get_ray_cluster_servlet()` before...
It was called only in the `test_cluster_servlet` fixture, which isn't called anywhere in the tests. Think we can delete this fixture for now.

runhouse/resources/hardware/cluster.py
```diff
@@ -2107,10 +2167,15 @@ def remove_conda_env(
...
-        self.call_client_method("get_certificate")
-        logger.info(
-            f"Latest TLS certificate for {self.name} saved to local path: {self.cert_config.cert_path}"
-        )
+        try:
+            self.call_client_method("get_certificate")
+            logger.info(
+                f"Latest TLS certificate for {self.name} saved to local path: {self.cert_config.cert_path}"
+            )
+        except ConnectionError:
```
jlewitt1, line 2175: whats the rationale here for logging the error instead raising an exception in the client?

BelSasha: I thought it will be more user friendly to log an error with a relevant msg, but we could also raise an exception instead, something like:
```
except ConnectionError:
    raise ValueError/RuntimeError(
        "Failed to update periodic cluster status checks: unable to connect to cluster's HTTP client."
    )
```

tests/test_servers/test_servlet.py
```diff
@@ -75,3 +80,29 @@ def test_get_obj_does_not_exist(self, test_servlet):
...
         assert isinstance(error, KeyError)
+
+    @pytest.mark.level("local")
+    def test_failed_cluster_servlet(self):
+
+        invalid_cluster_config = {
+            "api_server_url": "https://api.run.house.invalid",
+            "name": "mocked-cluster",
+        }
+        cluster_servlet = get_ray_cluster_servlet(
```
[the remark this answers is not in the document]
jlewitt1: im having trouble following the logic tho, bc we're loading the cluster_servelet here, and then loading it again a couple lines below. If all we need is to init ray, then why can't we just call ray init here instad?

runhouse/resources/hardware/cluster.py
jlewitt1: the exception is already being raised tho inside of `call_client_method`

tests/test_servers/test_servlet.py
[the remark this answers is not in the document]
BelSasha: yeah, i see your point, make sense. 

(we loading it again via ray later on just to check that is was indeed deleted as expected.)

runhouse/servers/cluster_servlet.py
```diff
@@ -91,6 +95,11 @@ async def __init__(
         # will be used when self.gpu_metrics will be updated by different threads.
         self.lock = threading.Lock()
 
+        # will be used in the periodic loop which sends cluster logs to den. Since we are saving logs to the server.log
+        # file asynchronously, there might be a use case where we want to send logs to den but the server.log has not
+        # been created yet, since it is a freshly initialized cluster servlet. In this case, we will return an empty string.
+        self.is_log_file_ready = False
```
jlewitt1, line 101: this can prob be a private variable

tests/test_servers/test_servlet.py
```diff
@@ -75,3 +79,39 @@ def test_get_obj_does_not_exist(self, test_servlet):
...
+        )
+        try:
+            # waiting for a seconds so the async cluster status check thread will be executed and then fail and kill the clsuter servlet.
+            time.sleep(1)
+            cluster_servlet = ray.get_actor(
+                name="invalid_cluster_servlet", namespace="runhouse"
+            )
+            # if the cluster servlet is not deleted -> it is not none -> not cluster_servlet = False ->
+            # the test fails, because we expect that the cluster servlet will be deleted
+            assert not cluster_servlet
```
jlewitt1, line 111: is this supposed to return an empty object, or raise a ValueError?

runhouse/servers/cluster_servlet.py
```diff
@@ -332,7 +341,9 @@ async def acheck_cluster_status(self, send_to_den: bool = True):
 
         return status, den_resp_status_code
 
-    async def acheck_cluster_logs(self, interval_size: int):
+    async def acheck_cluster_logs(
+        self, interval_size: Optional[int] = DEFAULT_STATUS_CHECK_INTERVAL
```
jlewitt1, line 345: why make this optional if it's always being passed in?

runhouse/resources/hardware/cluster.py
```diff
@@ -793,6 +794,7 @@ def put(self, key: str, obj: Any, process: str = None):
         """
         if self.on_this_cluster():
             return obj_store.put(key, obj, process=process)
+
```
jlewitt1, line 797: can we revert all these other formatting changes in the file to be consistent with main?

tests/test_servers/test_servlet.py
BelSasha: It suppose to raise an ValueError, with the msg we are asserting when we catch the exception (`"Failed to look up actor with name 'invalid_cluster_servlet'. This could because 1. You are trying to look up a named actor you didn't create. 2. The named actor died. 3. You did not use a namespace matching the namespace of the actor."`). If such exception is raised => the assertion of the msg is `True` and the test passes.

If the ValueError is not raised => ray found the `invalid_cluster_servlet` (it is not deleted) => `cluster_servlet != None` => `assert not cluster_servlet` is `false` => the test fails.

runhouse/servers/obj_store.py
```diff
@@ -73,6 +73,8 @@ def __init__(self, *args):
 def get_cluster_servlet(
     create_if_not_exists: bool = False,
     runtime_env: Optional[Dict] = None,
+    cluster_config: Optional[Dict] = None,
```
jlewitt1, line 76: dont love changing the usage of this method solely for the purposes of supporting the unit test .. can we just initialize the cluster servlet by calling ray.remote separately? (or put that speciifc code block in a helper function)

tests/test_servers/test_servlet.py
jlewitt1: what would be the reason why this test behaves differently on each run? like why wouldn't an exception always be raised? 

if we do expect the test to raise the error, we should do: 
```
with pytest.raises(ValueError):
   ...
```

runhouse/servers/obj_store.py
BelSasha: Yeah, makes sense. 

Decided to go with the helper function approach, since we are initializing the NodeServlet and the Servlet in a very similar way, and this approach might be relevant to those initializations as well.

However, I currently created a helping function just for the ClusterServlet init, just to unblock this PR. Creating a generic helper function for all servlets might be a bit trickier (there are some nuances to take into consideration), so I think it will be wiser to do it in a separate PR. 

@jlewitt1 wdyt?

runhouse/utils.py
```diff
@@ -889,3 +890,32 @@ def get_gpu_usage(collected_gpus_info: dict, servlet_type: ServletType):
         gpu_usage["utilization_percent"] = gpu_utilization_percent
 
     return gpu_usage
+
+
+####################################################################################################
+# ray utils
+####################################################################################################
+def init_remote_cluster_servlet_actor(
+    current_ip: Any,
```
jlewitt1, line 899: is `current_ip` ever not a string?

runhouse/servers/cluster_servlet.py
```diff
@@ -566,9 +585,22 @@ async def _aperiodic_gpu_check(self):
...
+
+                # killing the cluster servlet only if log level is debug: this indicates that the cluster is created
+                # during test, or by a runhouse team member
+                log_level = os.getenv("RH_LOG_LEVEL") or DEFAULT_LOG_LEVEL
+                if log_level.lower() == "debug":
+                    kill_actors(
+                        gracefully=False,
+                        actor_name=self.name,
+                    )
+                else:
```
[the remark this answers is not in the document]
BelSasha: how about the following approach:
if we caught and exception and `log_level != "debug"`, this loop will sleep for 1 min (instead of 5 seconds?)

runhouse/servers/cluster_servlet.py
```diff
@@ -378,6 +390,13 @@ async def aperiodic_autostop_check(self):
             except Exception as e:
                 logger.error(f"Autostop check has failed: {e}")
 
+                # killing the cluster servlet only if log level is debug: this indicates that the cluster is created
```
carolineechen, line 393: what if a user sets their RH_LOG_LEVEL env var to debug? do we still intend to kill the servlet then?

runhouse/servers/cluster_servlet.py
```diff
@@ -719,7 +761,13 @@ async def send_cluster_logs_to_den(
             # Likely a sign that the daemon was restarted, so we should start from the beginning
             prev_end_log_line = 0
 
-        logs_to_den = "\n".join(latest_logs[prev_end_log_line:])
+        # len(latest_logs) may equal to 0 if we are in the process of setting up the cluster servlet.
+        # In this case the server.log might nov have been created yet, therefore self._get_logs() returns an empty string ("").
```
carolineechen, line 765:
```suggestion
        # In this case the server.log might not have been created yet, therefore self._get_logs() returns an empty string ("").
```

runhouse/utils.py
```diff
@@ -890,3 +891,32 @@ def get_gpu_usage(collected_gpus_info: dict, servlet_type: ServletType):
...
+            name=servlet_name,
+            get_if_exists=True,
+            lifetime="detached",
+            namespace="runhouse",
+            max_concurrency=1000,
+            resources={f"node:{current_ip}": 0.001},
+            num_cpus=0,
+            runtime_env=runtime_env,
+        )
+        .remote(cluster_config=cluster_config, name=servlet_name)
```
carolineechen, line 920: just wondering what's the desired outcome of passing in cluster_config and name here change?

runhouse/servers/cluster_servlet.py
BelSasha: This a good point, I guess it is not ideal, but I feel like it is the most elegant solution we've came up with. Hopefully we'll cover most of the failures on our own, since we set the log level to 'DEBUG' locally as well as in CI. This way we'll be in a pretty got place before release.

Indeed it is a bit risky to kill the servlet for users outside of our team. The benefit of doing this, however, is that users will probably reach to us with new errors that haven't encountered with. (The servlet errors are printed to the cluster logs regardless to the log level, but the chances the the user will notice and send them to us are pretty low, since the cluster is still up and working).

Another open question is how often will users set the log level to 'DEBUG', but this can be answered only once will have bunch of runhouse users. 

To sum this up - I think that for now its fine to proceed with this solution because:

1. We don't have much traffic / users at the moment.
2. This already found one bug that I fixed.
3. This will allow us to monitor other errors over time, especially now that we've introduced a lot of new changes and refactors. 

I think that we could bring it up for a further discussion at the beginning of the year. wdyt?

runhouse/utils.py
BelSasha: This relates to the discussion with Josh above (cc'ed you). For testing purposes, I wanted to create a mocked invalid cluster servlet that will be killed. To do so, I needed to create a cluster servlet with specific invalid cluster_config, and a specific name. The name is passed so `kill_actors` will kill only the invalid cluster servlet, and not other valid cluster servlets that are being used for other tests. 

It does not affect the main cluster_servlet creation: in this case the cluster_config will be `None` and the name will be `cluster_servlet`, as it was before.

runhouse/servers/obj_store.py
BelSasha: cc: @carolineechen

runhouse/utils.py
```diff
@@ -890,3 +891,32 @@ def get_gpu_usage(collected_gpus_info: dict, servlet_type: ServletType):
...
+
+
+####################################################################################################
+# ray utils
+####################################################################################################
+def init_remote_cluster_servlet_actor(
+    current_ip: str,
+    runtime_env: Optional[Dict] = None,
+    servlet_name: str = "cluster_servlet",
+    cluster_config: Optional[dict] = None,
```
carolineechen, line 903:
```suggestion
    cluster_config: Optional[Dict] = None,
```

runhouse/servers/cluster_servlet.py
carolineechen: sure we can go ahead with this for now, can you update the comment to just reflect that this is the case for debug mode and not necessary just for test/runhouse.

and let's also revisit this when @jlewitt1 is back too

runhouse/utils.py
carolineechen: is there a reason `name=servlet` is included in both `options` and `remote`? also seeing the discussion above, I generally agree with not updating code for the sake of testing - could we leave it as it was, and just copy over the same lines of code for that test (or create a helper function inside test rather than in the core code if it's used more than once in test?)

BelSasha: Yeah, makes sense. Added the helper function to the tests/utils.py. 

The name was passed to `options` to set the name of the ray actor instance. I also passed it to `remote` so we could save the name of the ClusterServelt as an attribute (`self.name`). By doing that, in a case the ClusterServlet should be killed, we'll kill a the specific clusterServlet, and no all actors in `runhouse` namespace. 
Meaning, instead of running:
`kill_actors(gracefully=False, namespace="runhouse")`
we could run 
`kill_actors(gracefully=False, name=self.name)`

I thought that passing the name to the `kill_actors` method is a safer way to kill the cluster servlet (both "in real time" and during tests). Having said that, so far it looks like running kill_actors with the `namespace=runhouse` instead of the `actor_name=self.name` does not causing tests breakage. so I assume that for now could leave the changes to the bare minimum, and pass the name only to options, as we previously did. (We should keep an eye that test won't brake in the future as well).

carolineechen (approved):
can you fix pre-commit before you merge? (edit - I see it's caused by something outside of the PR, feel free to merge)

PR #1459, opened 2024-11-24
reorder teardown sequence

jlewitt1:
Teardown the cluster itself before updating Den. Two main reasons for this: 
- if there's an exception with the teardown proper we shouldn't update the DB 
- previously we were only updating the status of the runhouse daemon, but now we also update the status of the cluster itself. doing so means that if we update the DB first, the teardown will fail bc the cluster is marked as already "stopped"

### DIFF
--- a/runhouse/resources/hardware/on_demand_cluster.py
+++ b/runhouse/resources/hardware/on_demand_cluster.py
@@ -579,8 +579,15 @@ class OnDemandCluster(Cluster):
         Example:
             >>> rh.ondemand_cluster("rh-cpu").teardown()
         """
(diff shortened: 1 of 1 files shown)


jlewitt1:
* **#1459** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1459?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1459?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

PR #1460, opened 2024-11-24
use cluster status for cluster CLI commands

jlewitt1:


jlewitt1:
* **#1460** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1460?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1460?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

runhouse/main.py
```diff
@@ -246,11 +246,10 @@ def cluster_list(
         help="Time duration to filter on. Minimum allowable filter is 1 minute. You may filter by seconds (s), "
         "minutes (m), hours (h) or days (s). Examples: 30s, 15m, 2h, 3d.",
     ),
-    cluster_status: Optional[ClustersListStatus] = typer.Option(
+    cluster_status: Optional[ClusterStatus] = typer.Option(
         None,
         "--status",
-        help="Cluster status to filter on. Supported filter values: 'running', 'terminated' (cluster is not live), "
-        "'down' (Runhouse server is down, but the cluster might be live).",
+        help="Cluster status to filter on.",
```
carolineechen, line 252: is it possible to include the supported values directly in typer?

runhouse/resources/hardware/utils.py
```diff
@@ -410,15 +407,28 @@ def parse_filters(since: str, cluster_status: str):
...
         get_clusters_params.update(cluster_filters)
 
     # If not filters are specified, get only running clusters.
     elif not cluster_filters:
+        # For ondemand clusters, use cluster status (via Sky). for other clusters, use the status of the daemon
         get_clusters_params.update(
-            {"cluster_status": "running", "since": LAST_ACTIVE_AT_TIMEFRAME}
+            {
+                "cluster_status": ClusterStatus.RUNNING,
+                "status": RunhouseDaemonStatus.RUNNING,
```
carolineechen, line 429: should we call it something like "daemon_status" and be more explicit than "status" here since there's multiple statuses? to avoid potential confusion

runhouse/resources/hardware/utils.py
```diff
@@ -410,15 +407,28 @@ def parse_filters(since: str, cluster_status: str):
 def get_clusters_from_den(cluster_filters: dict):
     get_clusters_params = {"resource_type": "cluster", "folder": rns_client.username}
 
+    if "cluster_status" in cluster_filters:
+        # Include the relevant daemon status for the filter based on the cluster status provided
+        if cluster_filters["cluster_status"] == ClusterStatus.RUNNING:
+            cluster_filters["status"] = RunhouseDaemonStatus.RUNNING
```
carolineechen, line 413: daemon doesnt necessarily need to be running if the cluster is up right?

runhouse/main.py
jlewitt1: the enum is included by default like this:
```
  --status [running|terminated|initializing|unknown]
                                  Cluster status to filter on.
```
does that look ok to you?

runhouse/resources/hardware/utils.py
jlewitt1: ya we've been thinking about this, just has a bit of BC implications. @mkandler you think we should make this change now also?

runhouse/resources/hardware/utils.py
jlewitt1: that's true, this is a little tricky tho - bc this is meant to support clusters (at least for now) not launched via the launcher where we only have the daemon status - but given we're moving towards all launching happening via the launcher we can prob remove and rely just on the cluster status

carolineechen (approved):

PR #1478, opened 2024-11-26
Sync user config as a secret

carolineechen:

PR #1486, opened 2024-11-27
Update cluster restart server so periodic cluster status checks won't fail

BelSasha:
Context: the config yaml needs to be on the cluster before we run the restart command, otherwise cluster servlet checks (damon activity, logs, etc.) will fail initially 

Potential options to discuss: 
- revert back to rsyncing instead of requiring an HTTP connection 
- decouple cluster servlet initialization from the server restart 
- polling / waiting mechanism where cluster servelt initialization only starts once the config has been sent to the cluster
- passing in config as server restart params 

**Dec 2nd**:
After talking to @carolineechen and @rohinb2, decided to revert back to rsyncing instead of requiring an HTTP connection (reverting PR #[1478](https://github.com/run-house/runhouse/pull/1478) ).

### DIFF
--- a/runhouse/resources/hardware/cluster.py
+++ b/runhouse/resources/hardware/cluster.py
@@ -4,13 +4,16 @@ import importlib
 import json
 import os
 import re
+import shutil
 import subprocess
+import tempfile
 import threading
 import warnings
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 import requests
+import yaml
 
 from runhouse.resources.hardware.utils import (
     _setup_creds_from_dict,
@@ -1128,6 +1131,30 @@ class Cluster(Resource):
         # Update the cluster config on the cluster
         self.save_config_to_cluster()
 
+        # Save a limited version of the local ~/.rh config to the cluster with the user's
+        # if such does not exist on the cluster
+        if rns_client.token:
+            user_config = {
+                "token": rns_client.cluster_token(resource_address=rns_client.username),
+                "username": rns_client.username,
+                "default_folder": rns_client.default_folder,
+            }
+
+            yaml_path = Path(f"{tempfile.mkdtemp()}/config.yaml")
+            with open(yaml_path, "w") as config_file:
+                yaml.safe_dump(user_config, config_file)
+
+            try:
+                self.rsync(
+                    source=yaml_path,
+                    dest="~/.rh/",
+                    up=True,
+                    ignore_existing=True,
+                )
+                logger.debug("saved config.yaml on the cluster")
+            finally:
+                shutil.rmtree(yaml_path.parent)
+
         restart_cmd = (
             base_cli_cmd
             + (" --restart-ray" if restart_ray else "")
@@ -1171,28 +1198,6 @@ class Cluster(Resource):
 
         self.put_resource(Env(name=DEFAULT_PROCESS_NAME))
         # TODO - image env vars and secrets
-
-        # Save a limited version of the local ~/.rh config to the cluster with the user's hashed token,
-        # if such does not exist on the cluster
-        if rns_client.token:
-            from runhouse.resources.secrets import Secret
-
-            user_config = {
-                "token": rns_client.cluster_token(resource_address=rns_client.username),
-                "username": rns_client.username,
-                "default_folder": rns_client.default_folder,
-            }
-
-            user_config_secret = Secret(name="rh_user_config", values=user_config).to(
-                self
-            )
-            self.call(
-                user_config_secret.name,
-                "_write_to_file",
-                path="~/.rh/config.yaml",
-                format="yaml",
-            )
-            logger.debug("Saved user config to cluster")
 
         return status_codes
 
@@ -1370,6 +1375,7 @@ class Cluster(Resource):
         contents: bool = False,
         filter_options: str = None,
         stream_logs: bool = False,
+        ignore_existing: Optional[bool] = False,
     ):
         """
         Sync the contents of the source directory into the destination.
@@ -1389,6 +1395,8 @@ class Cluster(Resource):
                 an additional directory layer at the destination. (Default: ``False``).
             filter_options (Optional[str], optional): The filter options for rsync.
             stream_logs (Optional[bool], optional): Whether to stream logs to the stdout/stderr. (Default: ``False``).
+            ignore_existing (Optional[bool], optional): Whether the rsync should skip updating files that already exist
+                on the destination. (Default: ``False``).
 
         .. note::
             Ending ``source`` with a slash will copy the contents of the directory into dest,
@@ -1411,6 +1419,7 @@ class Cluster(Resource):
                     contents=contents,
                     filter_options=filter_options,
                     stream_logs=stream_logs,
+                    ignore_existing=ignore_existing,
                 )
             return
 
@@ -1445,6 +1454,8 @@ class Cluster(Resource):
                 source,
                 dest,
             ]  # -a is archive mode, -v is verbose, -z is compress
+            if ignore_existing:
+                cmd += ["--ignore-existing"]
             if filter_options:
                 cmd += [filter_options]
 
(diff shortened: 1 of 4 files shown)


BelSasha:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1486?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1486** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1486?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1486?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1485** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1485?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

jlewitt1 (approved):

runhouse/resources/hardware/cluster.py
```diff
@@ -1127,6 +1128,33 @@ def _start_or_restart_helper(
...
+                self.run(
+                    ["[ -f ~/.rh/config.yaml ]"],
+                    node=self.head_ip,
+                    require_outputs=False,
+                )[0]
+                == 0
+            ):
+                logger.debug("Did not change config.yaml")
+            else:
+                command = f"echo '{user_config}' > ~/.rh/config.yaml"
```
carolineechen, line 1154: we need to change this to a proper rsync rather than the `echo xxx`, which will log and print the token

runhouse/resources/hardware/cluster.py
```diff
@@ -1128,6 +1128,24 @@ def _start_or_restart_helper(
         # Update the cluster config on the cluster
         self.save_config_to_cluster()
 
+        # Save the local ~/.rh config to the cluster with the user's hashed token,
+        # if such does not exist on the cluster
+        if rns_client.token:
```
jlewitt1, line 1133: why are we branching here, i dont think we had this previously? if the token deosn't exist we shouldn't need to sync anything

runhouse/resources/hardware/cluster.py
```diff
@@ -1128,6 +1128,24 @@ def _start_or_restart_helper(
...
+            if (
+                self.run(
```
jlewitt1, line 1135: this runs a subprocss command that will print to the console no? (also we can just use rsync i think)

runhouse/resources/hardware/cluster.py
```diff
@@ -1128,6 +1128,24 @@ def _start_or_restart_helper(
...
+                    node=self.head_ip,
+                    require_outputs=False,
+                )[0]
+                == 0
+            ):
+                logger.debug("Did not change config.yaml")
+            else:
+                source_path = Path("~/.rh/config.yaml").expanduser()
+                dest_path = "~/.rh/config.yaml"
+                self.rsync(source=source_path, dest=dest_path, up=True, is_file=True)
```
jlewitt1, line 1146: this will rsync the whole file, did you look into rsyncing just the portions we want?

runhouse/resources/hardware/cluster.py
```diff
@@ -1366,6 +1362,7 @@ def rsync(
         source: str,
         dest: str,
         up: bool,
+        is_file: bool = False,
```
jlewitt1, line 1365: not sure i understand why we need this?

runhouse/resources/hardware/cluster.py
```diff
@@ -1128,6 +1128,24 @@ def _start_or_restart_helper(
...
+                self.run(
+                    ["[ -f ~/.rh/config.yaml ]"],
+                    node=self.head_ip,
+                    require_outputs=False,
+                )[0]
+                == 0
+            ):
+                logger.debug("Did not change config.yaml")
+            else:
+                source_path = Path("~/.rh/config.yaml").expanduser()
```
BelSasha, line 1144: @jlewitt1 If we want to rsync a condensed config.yaml, it looks like we will have to save it in a temp file (which can be removed after the rsync). 

Pros:
* we save only the necessary fields in the config.yaml we save on the cluster
Cons:
* need to mess a little bit with the user's file system each time he launches a cluster (create a temp file for the condensed config -> rsync it -> delete the temp file+folder)

any thoughts?

runhouse/resources/hardware/cluster.py
BelSasha: @jlewitt1 added a comment below

runhouse/resources/hardware/cluster.py
jlewitt1, line 1144:
```suggestion
                source_path = Path(CONFIG_YAML_PATH).expanduser()
```

runhouse/resources/hardware/cluster.py
BelSasha: actually we had it previously, we created the config.yaml only if there is a rh token. (same goes with the new implementation)

runhouse/resources/hardware/cluster.py
jlewitt1: prob ok to put this in a try / finally block where we delete the temp file after the rsync command is run

runhouse/resources/hardware/cluster.py
```diff
@@ -1128,6 +1129,41 @@ def _start_or_restart_helper(
...
+        # Save a limited version of the local ~/.rh config to the cluster with the user's
+        # if such does not exist on the cluster
+        if rns_client.token:
+            user_config = {
+                "token": rns_client.cluster_token(resource_address=rns_client.username),
+                "username": rns_client.username,
+                "default_folder": rns_client.default_folder,
+            }
+
+            condensed_config_yaml_path = Path(
```
jlewitt1, line 1141: maybe easier if we throw this in a temp dir, and then just nuke the temp dir after?

BelSasha: is it basically what we are doing, no? I'm creating the tmp dir inside `.rh` dir, and call it `condensed_config` instead of just `temp`

Since we are saving the token in the condensed config.yaml, I don't want to save the file outside of the .rh local folder.

BelSasha: @jlewitt1 yeah makes sense, changed it

runhouse/resources/hardware/cluster.py
```diff
@@ -1128,6 +1130,33 @@ def _start_or_restart_helper(
...
+        # Save a limited version of the local ~/.rh config to the cluster with the user's
+        # if such does not exist on the cluster
+        if rns_client.token:
+            user_config = {
+                "token": rns_client.cluster_token(resource_address=rns_client.username),
+                "username": rns_client.username,
+                "default_folder": rns_client.default_folder,
+            }
+
+            condensed_config_yaml_path = (
```
jlewitt1, line 1142: i think we can do something like this in just a couple lines:
```
yaml_path = Path(f"{tempfile.mkdtemp()}/config.yaml")
with open(yaml_path, "w") as config_file:
    yaml.safe_dump(user_config, config_file)
```

my worry is messing with the "~/.rh" folder in any capacity

runhouse/resources/hardware/cluster.py
```diff
@@ -1128,6 +1131,31 @@ def _start_or_restart_helper(
...
+        # if such does not exist on the cluster
+        if rns_client.token:
+            user_config = {
+                "token": rns_client.cluster_token(resource_address=rns_client.username),
+                "username": rns_client.username,
+                "default_folder": rns_client.default_folder,
+            }
+
+            yaml_path = Path(f"{tempfile.mkdtemp()}/config.yaml")
+            yaml_path.parent.mkdir(parents=True, exist_ok=True)
```
jlewitt1, line 1144: doesn't `tempfile.mkdtemp()` already create the directory?

jlewitt1 (approved):

PR #1504, opened 2024-12-02
Fix dupe image argument

py-rh:


carolineechen (commented):
ah my bad, shouldve properly rebased. there's two more duplicate instances in line 474 and 497

carolineechen (commented):
also at the current state, not sure about including image in `alt_options` on line 417 since we haven't properly implemented image saving yet so might lead to weird cluster reloading behavior

carolineechen (approved):
ty!

PR #1520, opened 2024-12-04
account for namespace in cluster ssh command

jlewitt1:


runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -683,10 +683,10 @@ def ssh(self, node: str = None):
             >>> rh.ondemand_cluster("rh-cpu", node="3.89.174.234").ssh()
         """
         if self.provider == "kubernetes":
-            command = f"kubectl get pods | grep {self.name}"
+            namespace_flag = f"-n {self._namespace}" if self._namespace else ""
 
+            command = f"kubectl get pods {namespace_flag} | grep {self.name}"
```
carolineechen, line 688: just a noting that if theres no namespace then this command (and the later one) will have 2 spaces, idk if adding a space at the beginning or end of the flag would be preferable but functionally it's fine

carolineechen (approved):

PR #1541, opened 2024-12-07
Fix server status

carolineechen:
* we can't call `_is_server_up` for `rh.here` since we can't call/create a client or check is_up from on the cluster. instead, checking if an obj store error or connection error is being thrown to detect if server is down, unless there's a better approach for on the cluster (rh.here)
* if server is down, we log and ask the user to run `runhouse cluster status` but that itself will fail if the server is down? or is that not supposed to be the case?

### DIFF
--- a/runhouse/main.py
+++ b/runhouse/main.py
@@ -9,7 +9,6 @@ from typing import Optional
 import ray
 
 import requests
-
 import typer
 import yaml
 from rich.console import Console
@@ -52,6 +51,8 @@ from runhouse.resources.hardware import (
     kill_actors,
 )
 from runhouse.resources.hardware.utils import ClusterStatus
+
+from runhouse.servers.obj_store import ObjStoreError
 
 SKY_LIVE_CLUSTERS_MSG = (
     "Live on-demand clusters created via Sky may exist that are not saved in Den. "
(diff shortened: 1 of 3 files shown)


carolineechen:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1541?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1541** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1541?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1541?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1540** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1540?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

jlewitt1 (approved):

runhouse/main.py
```diff
@@ -978,16 +979,14 @@ def server_status(
     """Check the HTTP server status on the cluster."""
     logger.debug("Checking the server status.")
     current_cluster = get_cluster_or_local(cluster_name=cluster_name)
```
jlewitt1, line 981: realizing this now - but if the cluster isn't found `get_cluster_or_local` will throw an exception, maybe we should put that in a try/catch so if there's a ValueError we handle it more gracefully?

PR #1545, opened 2024-12-08
cli cmds minor cleanup

jlewitt1:


jlewitt1:
* **#1545** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1545?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1545?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

runhouse/main.py
```diff
@@ -390,13 +390,14 @@ def cluster_up(
...
 
-        ``$ runhouse cluster up /sashab/rh-basic-cpu``
+        ``$ runhouse cluster up rh-basic-cpu``
 
     """
     current_cluster = rh.cluster(name=cluster_name, dryrun=True)
 
     try:
         current_cluster.up_if_not()
+        console.print("Cluster is already up.")
```
BelSasha, line 400: How about just `"Cluster is already up."`? Because we might call `runhouse cluster up` on a terminated cluster, and that it will be a bit wired to see` "Cluster is already up"` printed to the console in this case.

runhouse/main.py
```diff
@@ -430,7 +431,7 @@ def cluster_down(
 
     Example:
 
-        ``$ runhouse cluster down /sashab/rh-basic-cpu``
+        ``$ runhouse cluster down rh-basic-cpu``
```
BelSasha, line 434: Maybe we should add `console.print("Cluster is terminated.")` just for consistency?

runhouse/main.py
jlewitt1: oh we prob shouldn't do that, we should be able to know if the cluster is up or not, and then print a message if its already up.. ill change that

runhouse/main.py
jlewitt1: for this one we already log in the teardown APIs, so it will be a bit repetitive i think

runhouse/main.py
```diff
@@ -388,22 +388,26 @@ def cluster_up(
...
+        current_cluster.up()
     except ValueError:
-        console.print("Cluster is not saved in Den.")
+        console.print("Cluster not found in Den.")
     except Exception as e:
         console.print(f"Failed to bring up the cluster: {e}")
 
 
-@cluster_app.command("down")
+# @cluster_app.command("down")
```
BelSasha, line 410: any reason for commenting this out?

jlewitt1: hmm from debugging somehow good catch

PR #1548, opened 2024-12-09
fix cluster list all pythonic test

BelSasha:
Fixing the `cluster list all pythonic test`, issue was found after we started running den clean up cron every midnight.

### DIFF
--- a/tests/test_resources/test_clusters/test_cluster.py
+++ b/tests/test_resources/test_clusters/test_cluster.py
@@ -32,6 +32,7 @@ from runhouse.utils import _process_env_vars
 
 import tests.test_resources.test_resource
 from tests.conftest import init_args
+from tests.fixtures.resource_fixtures import create_folder_path
 from tests.test_resources.test_envs.test_env import _get_env_var_value
 from tests.utils import (
     friend_account,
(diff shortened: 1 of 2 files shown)


BelSasha:
* **#1548** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1548?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1548?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

tests/test_resources/test_clusters/test_cluster.py
```diff
@@ -1018,8 +1019,12 @@ def test_cluster_list_all_pythonic(self, cluster):
             original_username=original_username,
         ):
             # create dummy terminated cluster - set the daemon status to terminated, which will also
-            # update the cluster status
-            terminated_cluster = rh.cluster(name="terminated-cluster", ips=None).save()
+            # update the cluster status. Making sure that its name is in the same format as all testing clusters
+            # (contains timestamp and uuid)
+            terminated_cluster = rh.cluster(
```
jlewitt1, line 1024: any reason we're adding `server_connection_type` and removing `ips` here?

BelSasha: Yeah, so because the `ips` were set to `None`, the `alt_options` we set in the cluster factory method are an empty dictionary. Since we are deleting the resources every 24 hours, we are getting an exception when we try to load the terminated cluster from name => we are catching it and raising it again since there are no `alt_options`. 

Since it is a dummy cluster created just for the purpose of testing, I added `server_connection_type` so the `alt_options` will not be an empty dict. We don't really need to set the values of the ips to be explicitly set to `None` since it is the default behaviour. Therefore we are treating the cluster as terminated, which serves the purpose of this test.

[Code reference
](https://github.com/run-house/runhouse/blob/7caf3898e91c6e01bbb22a824019943fad3314b7/runhouse/resources/hardware/cluster_factory.py#L103)

jlewitt1 (approved):

PR #1591, opened 2024-12-13
Refactor cluster factory

carolineechen:


carolineechen:
* **#1593** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1593?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1592** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1592?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1591** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1591?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1591?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

runhouse/resources/hardware/utils.py
```diff
@@ -711,3 +752,92 @@ def __str__(self):
...
+###################################
+def setup_kubernetes(
+    namespace: Optional[str] = None,
+    kube_config_path: Optional[str] = None,
+    context: Optional[str] = None,
+    **kwargs,
+):
+    if kwargs.get("provider") and not kwargs.get("provider") == "kubernetes":
+        raise ValueError(
+            f"Recieved non kubernetes provider {kwargs.get('provider')} with kubernetes specific "
```
jlewitt1, line 768:
```suggestion
            f"Received non kubernetes provider {kwargs.get('provider')} with kubernetes specific "
```

runhouse/resources/hardware/utils.py
```diff
@@ -711,3 +752,92 @@ def __str__(self):
...
+            "cluster arguments."
+        )
+
+    if (
+        kwargs.get("server_connection_type")
+        and kwargs.get("server_connection_type") != ServerConnectionType.SSH
+    ):
+        raise ValueError(
+            "Runhouse K8s Cluster server connection type must be set to `ssh`. "
+            f"You passed {kwargs.getserver_connection_type}."
```
jlewitt1, line 778:
```suggestion
            f"You passed {kwargs.get('server_connection_type')}."
```

runhouse/resources/hardware/utils.py
```diff
@@ -711,3 +752,92 @@ def __str__(self):
...
+                stderr=subprocess.PIPE,
+                text=True,
+            )
+            logger.debug(process.stdout)
+            logger.info(f"Kubernetes namespace set to {namespace}")
+
+        except subprocess.CalledProcessError as e:
+            logger.info(f"Error: {e}")
+
+    # Q - should the following also only be set up for local launcher?
```
jlewitt1, line 806: we set up something similar in the launcher too, it's slightly different because the kube config is in a temp dir and not in `~/.kube`, so I think better to only have this be for the local flow

runhouse/resources/hardware/utils.py
```diff
@@ -711,3 +752,92 @@ def __str__(self):
...
+    if context and namespace:
+        logger.warning(
+            "You passed both a context and a namespace. Ensure your namespace matches the one in your context.",
+        )
+
+    launcher = kwargs.get("launcher") or configs.launcher
+    if namespace and launcher == "local":
+        # Set the context only if launching locally
+        # check if user passed a user-defined namespace
+        cmd = f"kubectl config set-context --current --namespace={namespace}"
```
jlewitt1, line 790: just in terms of ordering, I think we should set context before namespace (where relevant), bc if the local kube config has multiple contexts we'll need to first have that set before setting the namespace

runhouse/resources/hardware/cluster.py
```diff
@@ -445,6 +445,10 @@ def config(self, condensed: bool = True):
 
         return config
 
+    def _update_values(self, new_values: Dict[str, str]):
```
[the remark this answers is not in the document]
carolineechen: will update the type to `Dict[str, Any]`, I think serialization should happen through config()/from_config() but will double check the comparison of name to config

runhouse/resources/hardware/cluster_factory.py
```diff
@@ -95,260 +105,152 @@ def cluster(
...
+    cluster_args = locals().copy()
+    cluster_args["host"] = kwargs.get("ips") or (
+        [host] if isinstance(host, str) else host
+    )
+    cluster_args["creds"] = cluster_args.pop("ssh_creds")
+    cluster_args.pop("kwargs")
+    cluster_args = {k: v for k, v in cluster_args.items() if v is not None}
+
+    # check for invalid args
+    valid_kwargs = ONDEMAND_COMPUTE_ARGS | KUBERNETES_CLUSTER_ARGS | RH_SERVER_ARGS
```
jlewitt1, line 119: this will break for python < 3.9 no? technically we still support 3.8, tho not sure we need to be

runhouse/resources/hardware/utils.py
```diff
@@ -711,3 +760,90 @@ def __str__(self):
...
+    launcher = kwargs.get("launcher") or configs.launcher
+    if launcher == "local":
+        if context:
+            # check if user passed a user-defined context
+            try:
+                cmd = f"kubectl config use-context {context}"  # set user-defined context as current context
+                subprocess.run(cmd, shell=True, check=True)
+                logger.info(f"Kubernetes context has been set to: {context}")
+            except subprocess.CalledProcessError as e:
+                logger.info(f"Error setting context: {e}")
```
jlewitt1, line 803: wonder if we should actually raise an exception if we fail to set the context or namespace? we don't do it in den either, but this could lead to undesired behavior the user wouldn't want (ex: launching in a different namespace)

runhouse/resources/hardware/cluster_factory.py
```diff
@@ -95,260 +105,152 @@ def cluster(
...
-            host=host,
-            ssh_creds=ssh_creds,
-            server_port=server_port,
-            server_host=server_host,
-            server_connection_type=server_connection_type,
-            ssl_keyfile=ssl_keyfile,
-            ssl_certfile=ssl_certfile,
-            domain=domain,
-            den_auth=den_auth,
+    cluster_args = locals().copy()
```
jlewitt1, line 110: small nit and maybe just me, but was initially confused by the naming here, since `cluster_args` really behaves like kwargs

runhouse/resources/hardware/cluster_factory.py
carolineechen: oh good point, will edit

runhouse/resources/hardware/cluster_factory.py
carolineechen: yea using it as an easy way to get a dict out of the input args to easily pass around and use to check against cluster config, instead of manually creating a dict by passing in each arg as before (easy to miss an arg / forget to update when we add new factory args), but open to other approaches or naming, did you have one in mind?

jlewitt1: ya i like the way its extracted, its cleaner that way.. i think we can leave as is since it's not really "user facing" anyway

jlewitt1 (approved):

PR #1597, opened 2024-12-16
Update cluster ssh

carolineechen:
something broke with the ssh flow of a local (docker) ondemand cluster that isn't saved, with loading the ssh proxy command from a generated sky config. 

can just directly call `ssh <cluster_name>`, sky sets up the ssh config already and it should flow. no longer running into the previous need to call `conda deactivate` for docker images either

### DIFF
--- a/runhouse/main.py
+++ b/runhouse/main.py
@@ -10,7 +10,6 @@ import ray
 
 import requests
 import typer
-import yaml
 from rich.console import Console
 
 import runhouse as rh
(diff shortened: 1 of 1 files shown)


carolineechen:
* **#1597** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1597?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1597?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

runhouse/main.py
```diff
@@ -155,46 +154,21 @@ def cluster_ssh(cluster_name: str):
...
-        import sky
+        try:
+            import sky
 
-        from runhouse import OnDemandCluster
-        from runhouse.constants import DEFAULT_SSH_PORT
-        from runhouse.resources.hardware.utils import _run_ssh_command
-
-        state = sky.status(cluster_names=[cluster_name], refresh=False)
+            state = sky.status(cluster_names=[cluster_name], refresh=False)
```
jlewitt1, line 160: i guess it was already like this before, but running sky status here is really only relevant if the cluster's launcher is set to local no?

jlewitt1: oh i see, this is if we fail to even load it. ya i guess may as well try checking with local sky

carolineechen: yea, if the launcher is den the cluster should be saved/loaded down and run through the first part? but can also make it explicit here

jlewitt1 (approved):

runhouse/main.py
jlewitt1: i guess most explicit would be to exit if a value error is raised and launcher is set to den

carolineechen: actually if we can't load a cluster, we can't determine for sure if it was supposed to be launched with den or local, so I think leaving it probably handles the cases already (theoretically will be saved/loaded. and if for some reason it isn't, it won't be found locally if launched through den, and will raise an error then)

PR #1632, opened 2024-12-30
Fix k8 nightly release tests failures

BelSasha:
Tests that are still fails:
1. `test_autostop_call_updated` - both in CI and locally
2. `test_autostop_function_running` - both in CI and locally. 

Need further investigation, based on Q1 priorities.

### DIFF
--- a/runhouse/resources/hardware/on_demand_cluster.py
+++ b/runhouse/resources/hardware/on_demand_cluster.py
@@ -523,6 +523,9 @@ class OnDemandCluster(Cluster):
                     _, current_context = kubernetes.config.list_kube_config_contexts()
                     self.compute_properties["kube_context"] = current_context["name"]
 
+                self._kube_namespace = self.compute_properties.get("kube_namespace")
+                self._kube_context = self.compute_properties.get("kube_context")
+
     def _update_from_sky_status(self, dryrun: bool = False):
         # Try to get the cluster status from SkyDB
         if self.is_shared:
--- a/tests/test_resources/test_clusters/test_on_demand_cluster.py
+++ b/tests/test_resources/test_clusters/test_on_demand_cluster.py
@@ -261,8 +261,8 @@ class TestOnDemandCluster(tests.test_resources.test_clusters.test_cluster.TestCl
             cluster.compute_properties["internal_ips"] = []
             assert not cluster._ping(retry=False)
 
-        if cluster.compute_properties.get("cloud") == "kubernetes":
-            # kubernetes does not use ips in command runner
+        if not cluster.compute_properties.get("cloud") == "kubernetes":
+            # kubernetes does not use ips in command runner, so this test is relevant only for not k8 clusters.
             cluster.compute_properties["ips"] = ["00.00.000.11"]
             assert not cluster._ping(retry=False)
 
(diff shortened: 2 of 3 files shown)


BelSasha:
* **#1632** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1632?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1632?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tests/test_resources/test_clusters/test_on_demand_cluster.py
```diff
@@ -261,11 +261,6 @@ def test_cluster_ping_and_is_up(self, cluster):
             cluster.compute_properties["internal_ips"] = []
             assert not cluster._ping(retry=False)
 
-        if cluster.compute_properties.get("cloud") == "kubernetes":
```
carolineechen, line 264: was this failing for kubernetes or causing issues with later tests?

carolineechen: if you set `cluster.compute_properties["ips"] = original_ips` after the `assert` in this if statement, would that fix it?

BelSasha: 1. It was failing only for kubernetes
2. The tests fails when we call ` assert not cluster._ping(retry=False)` after we set `cluster.compute_properties["ips"] = ["00.00.000.11"]`.  If we set `cluster.compute_properties["ips"] = original_ips`, the test still fails in this section.

runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -520,6 +520,9 @@ def _populate_connection_from_status_dict(self, cluster_dict: Dict[str, Any]):
                     _, current_context = kubernetes.config.list_kube_config_contexts()
                     self.compute_properties["kube_context"] = current_context["name"]
 
+                self._kube_namespace = self.compute_properties.get("kube_namespace")
```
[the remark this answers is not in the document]
BelSasha: In the constructor they might be set to be `None`, if it is a fresh k8 cluster. Therefore, once it is up, we need to set the actual values for `kube_namespace` and `kube_context` that we get from the compute properties.

jlewitt1 (approved):

tests/test_resources/test_clusters/test_on_demand_cluster.py
carolineechen: do you happen to have the error message? I thought this would be fine for kubernetes since the command runner should use the kube context/namespace rather than ips to ping

BelSasha: Actually there is no error ms, in this case` not cluster._ping(retry=False) == False` => `assert not cluster._ping(retry=False)` fails

carolineechen: ohh wait actually I think this should be `if **not** cluster.compute_properties.get("cloud") == "kubernetes":` - for non k8s clusters, checking that if you set the ips to something incorrect that ping w/o retry fails (and this will work for kubernetes bc it doesnt rely on ips)

carolineechen (approved):

tests/test_resources/test_clusters/test_on_demand_cluster.py
BelSasha: @jlewitt1

PR #1637, opened 2025-01-03
Parallelize Image Setup, Fix local pip package install, move Ray/Runhouse install into Image

dongreenberg:


dongreenberg:
* **#1637** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1637?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1637?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

runhouse/resources/hardware/cluster.py
```diff
@@ -1082,47 +1062,22 @@ def _start_or_restart_helper(
...
-                and (local_rh_package_path.parent / "setup.py").exists()
-            )
-
-            if installed_editable_locally:
-                logger.debug("Runhouse is installed locally in editable mode.")
-                resync_rh = True
-            else:
-                # We only want this to be set if it was installed editable locally
-                local_rh_package_path = None
+        if resync_rh:
```
carolineechen, line 1072: do we not need to keep the `if rsync_rh is not None` to do detection of if we need to sync a local runhouse?

runhouse/resources/hardware/cluster.py
```diff
@@ -1082,47 +1062,22 @@ def _start_or_restart_helper(
...
-                resync_rh = True
+        image_secrets, image_env_vars = self._sync_image_to_cluster(parallel=parallel)
 
         if resync_rh:
-            self._sync_runhouse_to_cluster(
-                _install_url=_rh_install_url,
-                local_rh_package_path=local_rh_package_path,
-            )
-            logger.debug("Finished syncing Runhouse to cluster.")
+            # remove the runhouse install so it doesn't get reinstalled on restart
```
carolineechen, line 1079: should this be `if not rsync_rh`?

runhouse/resources/hardware/cluster.py
dongreenberg: We need to add a force argument here which takes in resync_rh, and if local runhouse package is editable then force=True by default (otherwise false by default).

carolineechen: `force_sync_local`

runhouse/resources/hardware/cluster.py
dongreenberg: No - this is specifically if we had rsync_rh enabled and therefore added some items to the image. Here we're cleaning them up by deleting them.

PR #1646, opened 2025-01-08
save condensed cluster config on the cluster

BelSasha:


BelSasha:
* **#1646** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1646?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1646?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

runhouse/resources/hardware/cluster.py
```diff
@@ -224,14 +225,28 @@ def _save_config_to_cluster(
         node: str = None,
     ):
         config = self.config(condensed=False)
+        config["has_cuda"] = is_gpu_cluster()
```
jlewitt1, line 228: we should be computing this on the cluster and not client side. also why do we need to add to the cluster config instead of storing as a cluster servlet private variable?

runhouse/resources/hardware/cluster.py
```diff
@@ -224,14 +225,28 @@ def _save_config_to_cluster(
...
+            "has_cuda",
+            "resource_subtype",
+            "den_auth",
+            "server_port",
+            "server_connection_type",
+            "server_host",
+            "autostop_mins",
+            "domain",
+        ]
+        if config.get("resource_subtype", None) == "OnDemandCluster":
```
jlewitt1, line 241: confused about this - a generic cluster still has compute properties how come its limited to OnDemand?

runhouse/resources/hardware/cluster.py
```diff
@@ -1015,6 +1028,10 @@ def status(self, send_to_den: bool = False):
                 "run `rh.configs.enable_observability()` and restart the server (`cluster.restart_server()`)."
             )
 
+        # adding the full cluster_config, because the returned config is a condensed version of the cluster_config
+        # saved locally on the cluster.
+        status["cluster_config"] = self.config()
```
jlewitt1, line 1033: cant we compute this before, and then just extract the useful keys? instead of doing it twice?

tests/test_servers/test_servlet.py
```diff
@@ -82,9 +82,11 @@ def test_get_obj_does_not_exist(self, test_servlet):
         error = deserialize_data(resp.data["error"], "pickle")
         assert isinstance(error, KeyError)
 
+    @pytest.mark.skip(
```
jlewitt1, line 85: this seems like an issue, why would this rely on the api server value in the cluster config?

runhouse/resources/hardware/cluster.py
BelSasha: yeah, agreed about computing it on the client side.

We add it to the cluster config so it will be easier to construct the cluster status CLI output. We could easily determine if a cluster is a GPU or a CPU cluster by simply calling `cluster_config.get("has_cuda")`, when the cluster_config is the one returned by the `cluster.status()` method. If we want to determine if a cluster is a CPU or GPU, but store `has_cuda` as a servlet private variable, we should implement one of the following:

1. Create a private variable in the cluster object as well, and set it to: `cluster_obj.has_cude = is_gpu_cluster()`.
2. Run `is_gpu_cluster()`during the status cli output construction. 
Pros (1+2): easy to get its value during status cli output construction.
Cons (1+2): need to call is_gpu_cluster() twice; both in the cluster servlet and on the client side. 
3. Create an http endpoint which retrieve the `cluster_servlet._has_cuda` value.
Pros: Calling `is_gpu_cluster()` only once.
Cons: Creating an endpoint for a very specific usecase.

@jlewitt1 wdyt?

runhouse/resources/hardware/cluster.py
BelSasha: This code section is being executed during cluster restart, but before we actually restarting the server (meaning before running `runhouse server restart`). In this case, cluster_config of a "plain" cluster does not have the "compute_properties" field yet, and on demand does have it (because we already launched the cluster using sky, so we could extract the compute properties from sky DB).

runhouse/resources/hardware/cluster.py
BelSasha: hmm I'm not sure we can do that. 
When we call cluster.config() for the first time, we are doing it to save on the cluster only the useful keys. When we compose the cluster status in the cluster servlet, the condensed `cluster_config` is returned in the output (because it is the one saved on the cluster servlet). Therefore, if we want `cluster.status()` to return the full cluster_config, we have to compute it once again and replace the condensed config with the full one.

runhouse/resources/hardware/cluster.py
BelSasha: change naming: `has_cuda` -> `is_gpu`

jlewitt1 (approved):

PR #1668, opened 2025-01-12
fix not-cluster nightly release test fails

BelSasha:


BelSasha:
* **#1668** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1668?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1668?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tests/test_resources/test_data/test_folder.py
```diff
@@ -145,7 +145,7 @@ def test_cluster_folder_put_and_get_serialization_methods(
     def test_send_folder_to_dest(self, folder, dest):
         _check_skip_test(folder, dest)
 
-        system = "here" if dest.system == "file" else dest.system
+        system = dest.system
```
jlewitt1, line 148: whats incorrect about the line we previous had?

BelSasha: code reference: 
```
    def to(
        self,
        system: Union[str, Cluster],
        path: Optional[Union[str, Path]] = None,
    ):
        """Copy the folder to a new filesystem or cluster.
        Currently supported: ``here``, ``file``, ``gs``, ``s3``, or a cluster.

        Args:
            system (str or Cluster): Filesystem or cluster to move the folder to.
            path (str, optional): Path to move the folder to.

        Example:
            >>> local_folder = rh.folder(path="/my/local/folder")
            >>> s3_folder = local_folder.to("s3")
        """
        if system == "here":
            current_cluster_config = _current_cluster(key="config")
            if current_cluster_config:
                system = Cluster.from_config(current_cluster_config)
            else:
                system = self.DEFAULT_FS
            path = str(Path.cwd() / self.path.split("/")[-1]) if path is None else path

        if isinstance(system, Cluster):
            if not path:
                # Use a default path on the cluster (in ~/.cache/runhouse)
                dest_path = Folder.default_path(
                    rns_address=self.rns_address, system=system
                )
            else:
                # Destination path on the cluster should be a relative path
                dest_path = (
                    relative_file_path(file_path=path)
                    if Path(path).is_absolute()
                    else path
                )

            # rsync the folder contents to the cluster's destination path
            logger.debug(f"Syncing folder contents to cluster in path: {dest_path}")
            return self._to_cluster(system, path=dest_path)

        path = str(
            path or Folder.default_path(self.rns_address, system)
        )  # Make sure it's a string and not a Path

        system_str = getattr(
            system, "name", system
        )  # Use system.name if available, i.e. system is a cluster
        logger.info(
            f"Copying folder from {self.fsspec_url} to: {system_str}, with path: {path}"
        )

        # to_local, to_cluster and to_data_store are also overridden by subclasses to dispatch
        # to more performant cloud-specific APIs
        system = _get_cluster_from(system)
        if system == "file":
            return self._to_local(dest_path=path)
        elif system in ["s3", "gs"]:
            return self._to_data_store(system=system, data_store_path=path)
        else:
            raise ValueError(
                f"System '{system}' not currently supported as a destination system."
            )
```

It caused a bug when we were trying to sync s3_folder to local_folder. 
since `local_folder.system == file`, we set system to be `"here"`.
Later, when we called `s3_folder.to(system="here")`, according to the above code, we set the system to be `self.DEFAULT_FS`, which in s3_folder case is `"s3"`. Therefore, the `folder.to` method is trying to sync the s3_folder to another s3 folder in an un-existing bucket (which makes sense, since we are trying to sync it to a local folder). Therefore, we can an s3 runtime error (`RuntimeError: fatal error: An error occurred (NoSuchBucket) when calling the ListObjectsV2 operation: The specified bucket does not exist`).

If we are leaving the system to be dest.system == "file", the s3_folder is synced to the local folder, as expected.

BelSasha: Another fix to this might be leave `system = "here" if dest.system == "file" else dest.system`, but in the folder.to() change `system = self.DEFAULT_FS to be system = "file"`. 

Not sure which solution is less hacky tbh..

jlewitt1 (approved):

PR #1676, opened 2025-01-14
remove unused group endpoints

jlewitt1:


jlewitt1:
* **#1676** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1676?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1676?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

runhouse/resources/secrets/secret.py
```diff
@@ -21,10 +21,7 @@
 
 class Secret(Resource):
     RESOURCE_TYPE = "secret"
-
     USER_ENDPOINT = "user/secret"
```
carolineechen, line 24: if we don't plan on adding another endpoint, we can further modify `secrets/utils.py` functions to get rid of the `endpoint` params and directly use the constant instead

jlewitt1: its possible we may add an org one at some point, but for now we can treat as irrelevant. wdym tho by removing the endpoint params here?

carolineechen: we have 3 signatures using endpoint like `def load_config(name: str, endpoint: str = USER_ENDPOINT):` in `secrets/utils.py`. we can just make `def load_config(name: str)` and use the constant `USER_ENDPOINT` instead of `endpoint` within the function

but depending on if org endpoint will likely be added and fit in there, fine if we want to remove them now or leave it

carolineechen (approved):

PR #1679, opened 2025-01-14
[WIP] option to list shared resources in cluster list command

jlewitt1:


jlewitt1:
* **#1679** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1679?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1679?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

runhouse/resources/hardware/cluster.py
```diff
@@ -2465,7 +2468,7 @@ def list(
             >>> Cluster.list(show_all=True)
         """
         cluster_filters = (
-            parse_filters(since=since, cluster_status=status)
+            parse_filters(since=since, cluster_status=status, show_shared=show_shared)
             if not show_all
             else {"all": "all"}
         )
```
carolineechen, line 2474: a little bit confused, do we not need to add logic to `get_clusters_from_den` to handle shared clusters?

and question on behavior - if `show_shared=True`, will that only show the shared clusters that satisfy the other filters, or will it show both owned + shared satisfying the constraints? can both `show_all` and `show_shared` be True?

jlewitt1: oh i see what you mean, yea that's a little unclear.. do you think the least confusing would be to only allow one of those filters? otherwise it would be a little arbitary in terms of which we decide to show

carolineechen: I think either of the following could work as long as documented clearly, I just wasn't too sure where this logic was being implemented. right now it looks like the second case?
* only one of `show_all` and `show_shared`
* can allow both, ordered by last used (?) up to the max

PR #1683, opened 2025-01-14
update cluster creds sharing

jlewitt1:


jlewitt1:
* **#1683** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1683?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1683?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

runhouse/resources/hardware/launcher_utils.py
```diff
@@ -434,6 +432,23 @@ def keep_warm(cls, cluster, mins: int):
...
+            secret = rh.provider_secret("sky")
+            return secret
+
+        except ValueError:
+            # Create a new key pair required by Sky
+            path, _ = generate_ssh_keys()
+            secret = rh.provider_secret(
+                provider="sky", path=path, name=SSH_SKY_SECRET_NAME
+            )
+            secret.save()
```
carolineechen, line 448: I think `secret.save()` saves the resource but you need `.write()` to write it to file system

runhouse/resources/hardware/launcher_utils.py
```diff
@@ -434,6 +432,23 @@ def keep_warm(cls, cluster, mins: int):
             set_cluster_autostop_cmd = _cluster_set_autostop_command(mins)
             cluster.run_bash_over_ssh([set_cluster_autostop_cmd], node=cluster.head_ip)
 
+    @classmethod
+    def load_creds(cls):
+        """Loads the SSH credentials resource required for the launcher."""
+        try:
+            secret = rh.provider_secret("sky")
```
carolineechen, line 439: should this default to the user's default ssh key (which doesn't _need_ to be sky? and then try loading a sky secret if they don't have a default

runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -197,6 +197,15 @@ def image_id(self) -> str:
...
         return None
 
+    @property
+    def creds_values(self) -> Dict:
+        if self._creds is None:
+            if self.launcher == LauncherType.DEN:
+                self._creds = DenLauncher.load_creds()
+            elif self.launcher == LauncherType.LOCAL:
+                self._creds = LocalLauncher.load_creds()
+        return self._creds.values
```
carolineechen, line 207: we did `return {**self._creds.values, **self.ssh_properties}` for consistency with prior behavior. I think it's fine to change it to just return creds values (since creds arent shared but ssh properties are), but we should update `creds_values` for static clusters to match this behavior too

runhouse/resources/hardware/launcher_utils.py
```diff
@@ -332,6 +304,32 @@ def teardown(cls, cluster, verbose: bool = True):
...
+        default_ssh_key = rns_client.default_ssh_key
+        if default_ssh_key:
+            try:
+                secret = rh.Secret.from_name(default_ssh_key)
+                return secret
+            except ValueError:
+                pass
+
+        # try loading the default SSH key
+        secret = rh.provider_secret("ssh", path="~/.ssh/id_rsa")
```
carolineechen, line 319: not 100% sure the intended behavior here, but I think this can just be `rh.provider_secret("ssh")`?
* if a user saved `rh.provider_secret("ssh", path="some_other_path")`, then just calling "ssh" should reload from den. including the path will create a new secret object (I think cluster may be the only one that lets you reload from Den when you provide other arguments)
* if not saved in den, `rh.provider_secret("ssh")` is equivalent to `rh.provider_secret("ssh", path="~/.ssh/id_rsa")` since that is the default path used to look for `provider="ssh"`

so what you have is correct if we're only looking for default ssh key locally, but not if we additionally want to check den for a saved ssh secret

jlewitt1: ya good point, intention was to load the default which could be any arbitrary path set by the user

runhouse/resources/hardware/launcher_utils.py
jlewitt1: sky specifically requires the key to be in that path, but don't think we have to create a dedicated sky secret for it. re-using the default and copying into that path makes sense

runhouse/resources/hardware/cluster.py
```diff
@@ -208,13 +204,6 @@ def check_connect_server():
             raise ConnectionError(f"Check server failed: {e}.")
         return self._http_client
 
-    @property
-    def creds_values(self) -> Dict:
```
jlewitt1, line 212: we were previously bundling these values with the `ssh_properties`, but if we decouple them doesn't look like there's a need for us to store the actual values in a separate dedicated property

runhouse/resources/hardware/cluster.py
```diff
@@ -338,22 +334,32 @@ def _setup_creds(self, ssh_creds: Union[Dict, "Secret", str]):
...
             creds, ssh_properties = _setup_creds_from_dict(ssh_creds, self.name)
             self._creds = creds
             self.ssh_properties = ssh_properties
 
+    def _setup_default_creds(self):
+        from runhouse.resources.secrets import Secret
+
+        default_ssh_key = rns_client.default_ssh_key
+        if default_ssh_key is None:
+            return None
```
carolineechen, line 359: is this okay for it to return none in this case or should this be raising an error or warning?

runhouse/resources/hardware/launcher_utils.py
```diff
@@ -356,6 +318,29 @@ def teardown(cls, cluster, verbose: bool = True):
...
+        if not default_ssh_key:
+            raise ValueError(
+                "No default SSH key found in the local Runhouse config, "
+                "please set one by running `runhouse login`"
+            )
+        try:
+            # Note: we still need to load them down locally to use for certain cluster operations (ex: rsync)
+            secret = rh.Secret.from_name(default_ssh_key)
+            # Ensure this specific keypair is written down locally
+            secret.write(overwrite=True)
```
carolineechen, line 335: a bit hesitant on having overwrite be True, maybe throw an error instead that a secret exists in the location locally?

or if it's easy to wire up diff local vs saved file names, could generate a new unique file location locally (e.g. `~/.rh/ssh/cluster-name-ssh-key`). not sure if that'll lead to complications elsewhere tho (but also don't want to repeat saving the same creds under diff names)

runhouse/resources/hardware/cluster.py
```diff
@@ -646,6 +641,38 @@ def keep_warm(self):
...
+        if rns_address is None:
+            return False
+
+        # If shared the creds used for launching will be attributed to another user
+        creds: str = (
+            self._resource_string_for_subconfig(self._creds, True)
+            if hasattr(self, "_creds") and self._creds
+            else None
+        )
+        if isinstance(creds, dict):
```
carolineechen, line 657: from the line above, creds should be either a string or None right? and not enter the dict or secret check

runhouse/resources/hardware/launcher_utils.py
jlewitt1: hmm ya i see your point, the ore requiremnt here is that the key pair used in launch matches what's saved down locally in the sky path (~/.ssh/sky-key). waht aobut checking if the path exists locally, and only if it doesn't exist then writing those values down?

runhouse/resources/hardware/cluster.py
jlewitt1: good catch, i think that's right

runhouse/resources/hardware/cluster.py
jlewitt1: for static clusters i think that's ok, but will add a warning

runhouse/resources/hardware/launcher_utils.py
carolineechen: yea that sounds good, actually looking at `_write_to_file` for `SSHSecret1 we don't support `overwrite` for ssh anyways, and it currently throws a warning if there's a mismatch, we want to catch/raise the warning here or even make it an error

jlewitt1: oh cool, so in that case having that method raise an error and letting that propagate back seems reasonable

runhouse/resources/hardware/cluster.py
```diff
@@ -646,6 +645,32 @@ def keep_warm(self):
...
+
+        if not creds or isinstance(creds, dict):
+            # specifying custom creds (for non on-demand clusters)
+            return False
+
+        cluster_folder = rns_client.base_folder(rns_address)
+        cluster_creds_folder = rns_client.base_folder(creds)
+
+        if cluster_folder != rns_client.username and (
+            creds and cluster_creds_folder != rns_client.username
```
carolineechen, line 668: would `creds` ever equal `rns_client.username` or is just checking the folders sufficient? in `creds and cluster_creds_folder != rns_client.username`

runhouse/resources/hardware/utils.py
```diff
@@ -50,6 +50,10 @@ class LauncherType(str, Enum):
     LOCAL = "local"
     DEN = "den"
 
+    @classmethod
+    def strings(cls):
```
carolineechen, line 54: is this used anywhere?

jlewitt1: ya in the factory for validating the launcher val provided:
```
    if launcher not in LauncherType.strings():
        raise ValueError(
            f"Invalid launcher type '{launcher}'. Must be one of {LauncherType.strings()}."
        )
```

carolineechen (approved):

PR #1684, opened 2025-01-15
fix ping-and-up test for den launched clusters

BelSasha:


BelSasha:
* **#1684** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1684?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1684?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -526,16 +526,37 @@ def _populate_connection_from_status_dict(self, cluster_dict: Dict[str, Any]):
...
+            resp = requests.get(cluster_uri, headers=rns_client.request_headers())
+            if resp.status_code != 200:
+                logger.warning(
+                    f"Received [{resp.status_code}] from Den GET '{cluster_uri}': Failed to load cluster status"
+                )
+                return
+            cluster_den_status: list = read_resp_data(resp)
+
+            if not cluster_den_status:
+                logger.warning("Failed to update cluster info")
```
BelSasha, line 550: Decided to go with `"Failed to update cluster"` info since we are updating cluster info (such as compute properties etc), and not just cluster status.

runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -526,16 +526,37 @@ def _populate_connection_from_status_dict(self, cluster_dict: Dict[str, Any]):
...
             # If the cluster is shared can ignore, since the sky data will only be saved on the machine where
             # the cluster was initially upped
             return
 
-        if self.launcher == "local":
+        if self.launcher == LauncherType.LOCAL:
             cluster_dict = self._sky_status(refresh=not dryrun)
             self._populate_connection_from_status_dict(cluster_dict)
+        if self.launcher == LauncherType.DEN:
+            cluster_uri = rns_client.format_rns_address(self.rns_address or self.name)
```
jlewitt1, line 540: if we're using the den launcher it always have an rns address, is there a reason we have the `or self.name` here?

runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -526,16 +526,37 @@ def _populate_connection_from_status_dict(self, cluster_dict: Dict[str, Any]):
                 self._kube_namespace = self.compute_properties.get("kube_namespace")
                 self._kube_context = self.compute_properties.get("kube_context")
 
-    def _update_from_sky_status(self, dryrun: bool = False):
+    def _update_from_status(self, dryrun: bool = False):
         # Try to get the cluster status from SkyDB
         if self.is_shared:
```
jlewitt1, line 531: not sure if this check should apply for den launched clusters, since the data updated from status has to do with compute properties and not the creds themselves

runhouse/resources/hardware/on_demand_cluster.py
BelSasha: Yeah, the first time we call `self._update_from_status(dryrun=True)` is during the cluster construction, before we save the cluster (relevant for both den and local launched clusters), therefore we don't have `rns_address` yet. 
([Code reference](https://github.com/run-house/runhouse/blob/97d564c7c68b9253ab14d081b24e684ce5fa553a/runhouse/resources/hardware/on_demand_cluster.py#L147))

jlewitt1: so in that case isn't it enough to rely on the RNS address, and if it's none not hit den?

BelSasha: yeah, good point, agreed

runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -542,6 +542,45 @@ def _setup_default_creds(self):
         if self.launcher == LauncherType.DEN:
             return DenLauncher.load_creds()
 
+    def _update_from_status(self, dryrun: bool = False):
+
+        if self.launcher == LauncherType.LOCAL:
+            # Try to get the cluster status from SkyDB
+            if self._is_shared:
```
jlewitt1, line 549: should this shared check apply for den launcher too?

BelSasha: yeah makes sense, updated

PR #1685, opened 2025-01-15
runhouse server dependency updates

jlewitt1:


jlewitt1:
* **#1685** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1685?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1685?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

runhouse/main.py
```diff
@@ -935,6 +933,14 @@ def server_stop(
 
         ``$ runhouse server stop rh-cpu``
     """
+    try:
+        import ray
```
carolineechen, line 937: this is only in added in `server_stop`, but shouldn't ray also be required for other server commands in this file too, like `server_start`, `server_status`?

jlewitt1: i guess that depends more on whether the server is locally running or not, technically for a server on a cluster i dont think we would, but maybe we should just enforce it anyway

carolineechen (approved):

PR #1689, opened 2025-01-16
add info msg to cli output if logs are not available

BelSasha:


BelSasha:
* **#1689** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1689?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1689?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

runhouse/main.py
```diff
@@ -501,8 +501,14 @@ def cluster_logs(
         headers=rns_client.request_headers(),
     )
 
+    if (
+        resp.status_code == 404
+        and resp.json().get("detail") == "No logs found for the requested resource"
```
jlewitt1, line 506: do we need to parse the output, don't we know if its a 404 that the logs arent readfy yety?

runhouse/main.py
```diff
@@ -501,8 +501,11 @@ def cluster_logs(
         headers=rns_client.request_headers(),
     )
 
+    if resp.status_code == 404:
+        console.print("Cluster logs are not available yet.")
+        return
     if resp.status_code != 200:
-        console.print("Failed to load cluster logs.")
+        console.print(f"Failed to load cluster logs: {resp.json().get('detail')}.")
```
jlewitt1, line 508: is there any additional info here in the response we would want to expose to a user?

BelSasha: well it feels like it will be a bit easier to debug it, since the user might just report us the response msg back to us so well get a sense of what went wrong before going though the logs.  But I see your point why we should not print it to the user, removing.

jlewitt1 (approved):

PR #1711, opened 2025-01-23
fix cli discrepencies

BelSasha:


BelSasha:
* **#1711** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1711?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1711?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

runhouse/main.py
```diff
@@ -157,6 +157,9 @@ def cluster_ssh(
     try:
         c = cluster(name=cluster_name)
         if isinstance(c, rh.OnDemandCluster):
+            if c.is_initializing():
```
jlewitt1, line 160: can't we check `cluster._cluster_status` instead of having a dedicated endpoint?

runhouse/cli_utils.py
```diff
@@ -497,9 +497,15 @@ def get_cluster_or_local(cluster_name: str = None):
 
     if cluster_name:
         try:
-            current_cluster = rh.cluster(name=cluster_name, dryrun=True)
-        except ValueError:
-            console.print("Cluster not found in Den.")
+            current_cluster: rh.Cluster = rh.cluster(name=cluster_name, dryrun=True)
+            if (
+                isinstance(current_cluster, rh.OnDemandCluster)
+                and current_cluster._cluster_status == "initializing"
```
jlewitt1, line 503: what discrepency is this solving for? also were you able to reproduce this? (meaning reloading down a cluster with the _cluster_status set to initializing)?

BelSasha: hmm its a semi-discrepancy in my opinion. Without this change we'll get `Cluster <cluster_name> is not up. If it's an on-demand cluster, you can run runhouse cluster up <cluster_name> to bring it up automatically.`

It's not totally wrong, because the cluster is indeed is not up yet, but I think it might be more accurate to print the the cluster is currently being initialized (no need to run `runhouse cluster up`)

wdyt?

runhouse/main.py
```diff
@@ -157,6 +157,9 @@ def cluster_ssh(
     try:
         c = cluster(name=cluster_name)
         if isinstance(c, rh.OnDemandCluster):
+            if c._cluster_status == "initializing":
+                console.print(f"[reset]{cluster_name} is being initialized.")
```
jlewitt1, line 161: We could say "Cluster is currently initializing. Please wait for it to finish, "
                "or run `runhouse cluster up {cluster_name} -f` to abort the initialization and relaunch." 
And then we can also add a force argument to the up command (to make it identical to the python API).

wdyt?

runhouse/cli_utils.py
jlewitt1: that makes sense, but it's not clear to me if the clsuter status property will ever actually be updated to the "initializing" value. where in the flow (den + local) is this supposed to happen?

runhouse/main.py
BelSasha: yeah, sounds good!

runhouse/cli_utils.py
```diff
@@ -494,9 +494,15 @@ def get_cluster_or_local(cluster_name: str = None):
...
+            ):
+                console.print(f"[reset]{cluster_name} is being initialized.")
```
jlewitt1, line 502: could add the same messasge as below here no? but more generally, I wonder if there's a need for `get_cluster_or_local`. seems like we're applying it to a bunch of different commands, but not sure it's always necessary. for example with `keep_warm` we also check if the cluster is up, which we're already doing here.

runhouse/cli_utils.py
BelSasha: Locally-launched clusters:

1. If it is a new cluster (not saved in den), the constructor is calling `self._update_from_sky_status()` -> `self._populate_connection_from_status_dict()` ->  `self._update_cluster_status_from_sky_status() `, that could set self._cluster_status to` ClusterStatus.INITIALIZING`
2. If it a saved cluster: if the cluster is terminated and we call `self.up_if_not()` / `self.up()` -> `self._update_from_sky_status()` (rest of the flow is the same as in section 1)

Den-launched clusters:
When the launcher is initializing the cluster, it saves in the DB that the cluster status is "initializing". Therefore, if we create cluster from name (can happen if we call `rh cluster status` ->  `get_cluster_or_local()`  -> `cluster = rh.cluster(<cluster_name>) `=> in the constructor, there is the following call `self._cluster_status = kwargs.get("cluster_status")`, where `cluster_status == "initializing"`

runhouse/cli_utils.py
BelSasha: I think the main purpose of the `get_cluster_or_local` method is to enable running the cli commands on the cluster, without providing the cluster name. For example, we ssh into a cluster, and we could run `rh cluster status`, and the status will be printed to the console. I agree we could try and refactor this method, so it won't contain unnecessary and duplicate logic. wdyt?

jlewitt1: ya makes sense, i think it just needs to be adjusted depending on the use case. bc we don't always want to the same behavior for each command (ex: exiting if theres an errro). opened a spearate PR for that

jlewitt1 (approved):

runhouse/cli_utils.py
```diff
@@ -490,6 +492,19 @@ def get_local_or_remote_cluster(cluster_name: str = None, exit_on_error: bool =
                 raise typer.Exit(1)
             raise e
 
+        if isinstance(current_cluster, rh.OnDemandCluster):
+
+            # in case we called current_cluster.up() on a local cluster, we need to update the cluster_status,
+            # because its being updated properly only if we call up_if_not()
+            if current_cluster.launcher == LauncherType.LOCAL:
+                current_cluster._fetch_sky_status_and_update_cluster_status()
```
jlewitt1, line 500: maybe we should be more specific here on when to call this, I imagine we don't need to run a refresh anytime this is called?

runhouse/main.py
```diff
@@ -360,7 +365,13 @@ def cluster_up(
     cluster_name: str = typer.Argument(
         None,
         help="The name of cluster to bring up.",
-    )
+    ),
+    force: bool = typer.Option(
+        False,
+        "-f",
+        "--force",
+        help="Whether to abort the initialization and relaunch.",
```
jlewitt1, line 373:
```suggestion
        help="Whether to up the cluster regardless of its current initialization status",
```

runhouse/resources/hardware/on_demand_cluster.py
```diff
@@ -436,7 +436,7 @@ def _populate_connection_from_status_dict(self, cluster_dict: Dict[str, Any]):
         sky_status = cluster_dict["status"]
         self._update_cluster_status_from_sky_status(sky_status)
 
-        if sky_status in [SkyClusterStatus.UP, SkyClusterStatus.INIT]:
```
jlewitt1, line 439: do we not have the compiute properties at init stage?

runhouse/cli_utils.py
```diff
@@ -490,6 +492,22 @@ def get_local_or_remote_cluster(cluster_name: str = None, exit_on_error: bool =
...
+            # because its being updated properly only if we call up_if_not()
+            if (
+                current_cluster.launcher == LauncherType.LOCAL
+                and current_cluster._cluster_status != ClusterStatus.RUNNING
+            ):
+                current_cluster._fetch_sky_status_and_update_cluster_status()
+
+            if current_cluster._cluster_status == ClusterStatus.INITIALIZING:
+                console.print(
+                    f"[reset]{cluster_name} is being initialized. Please wait for it to finish, or run runhouse cluster up {cluster_name} -f to abort the initialization and relaunch."
```
jlewitt1, line 507: can we format the command like we do with the others? `[reset][bold italic]`cmd`[/bold italic]`

runhouse/cli_utils.py
```diff
@@ -490,6 +492,22 @@ def get_local_or_remote_cluster(cluster_name: str = None, exit_on_error: bool =
...
 
+        if isinstance(current_cluster, rh.OnDemandCluster):
+
+            # in case we called current_cluster.up() on a local cluster, we need to update the cluster_status,
+            # because its being updated properly only if we call up_if_not()
+            if (
+                current_cluster.launcher == LauncherType.LOCAL
+                and current_cluster._cluster_status != ClusterStatus.RUNNING
+            ):
+                current_cluster._fetch_sky_status_and_update_cluster_status()
```
jlewitt1, line 503: in this specific case shouldn't we be running sky with refresh=True?

BelSasha: yeah, good point

PR #552, opened 2024-03-05
Change mapper test to correctly check timing.

rohinb2:

PR #1712, opened 2025-01-23
by default, cluster list returns all clusters owned by the user

BelSasha:
depends on [#552](https://github.com/run-house/runhouse-den/pull/552) den PR

### DIFF
--- a/runhouse/cli_utils.py
+++ b/runhouse/cli_utils.py
@@ -93,10 +93,7 @@ def create_output_table(
         table.caption_justify = "left"
 
     if displayed_clusters == MAX_CLUSTERS_DISPLAY:
-        link_to_clusters_in_den = (
-            f"[reset]The full list of clusters can be viewed "
-            f"at https://www.run.house/resources?folder={rns_client.username}&type=cluster."
-        )
+        link_to_clusters_in_den = "[reset]The full list of clusters can be viewed at https://www.run.house/resources?type=cluster"
         if table.caption:
             table.caption += f"\n{link_to_clusters_in_den}"
         else:
--- a/runhouse/resources/hardware/utils.py
+++ b/runhouse/resources/hardware/utils.py
@@ -440,7 +440,7 @@ def parse_filters(since: str, cluster_status: Union[str, ClusterStatus]):
 
 
 def get_clusters_from_den(cluster_filters: dict, force: bool):
-    get_clusters_params = {"resource_type": "cluster", "folder": rns_client.username}
+    get_clusters_params = {"resource_type": "cluster", "owned": True}
 
     if (
         "cluster_status" in cluster_filters
@@ -528,8 +528,7 @@ def cast_last_active_timestamp(clusters: List[Dict[str, Any]]):
 def get_running_and_not_running_clusters(clusters: list):
     up_clusters, down_clusters = [], []
     for den_cluster in clusters:
-        # Display the name instead of the full Den address
-        cluster_name = den_cluster.get("name").split("/")[-1]
+        cluster_name = den_cluster.get("name")
         cluster_type = den_cluster.get("data").get("resource_subtype")
         cluster_status = den_cluster.get("cluster_status")
         cluster_status = cluster_status or ClusterStatus.UNKNOWN.value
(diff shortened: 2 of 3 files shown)


BelSasha:
* **#1712** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1712?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1712?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

runhouse/cli_utils.py
```diff
@@ -93,10 +93,10 @@ def create_output_table(
...
 
     if displayed_clusters == MAX_CLUSTERS_DISPLAY:
-        link_to_clusters_in_den = (
-            f"[reset]The full list of clusters can be viewed "
-            f"at https://www.run.house/resources?folder={rns_client.username}&type=cluster."
-        )
+        # TODO: The following link will lead to the dashboard where extra clusters will be displayed:
+        # all self + org cluster will be displayed, even org clusters that are not owned by the user.
+        # Need to discuss if this a desired behavior, because in for cluster list we show only user-owned clusters.
+        link_to_clusters_in_den = "[reset]The full list of clusters can be viewed at https://www.run.house/resources?type=cluster"
```
jlewitt1, line 99: would we be able to update the url to include the owned param ?

BelSasha: we could add it, but i think it might require some changes to the UI as well, because right now if i post `https://www.run.house/resources?resource=cluster&owned=True` in my browser, I see that not necessarily owned by me (shared with me / part of my org but owned by other org members)

cc: @mkandler

mkandler: I don't know if we want to show resources by owned on /resource page necessarily. Though I think it'd be helpful to include owner info in the table. Would have to update the query results to include a bit more info on the owners (besides ObjectID)

jlewitt1: i think for now its prob fine to link to the UI the way we have it set up, its easier for the user to navigate as opposed to displaying one massive table in the CLI

BelSasha: @jlewitt1 do you mean just `https://www.run.house/resources?type=cluster`? or the way we had it before? `https://www.run.house/resources?type=cluster&folder={rns_client.username}`

jlewitt1: i think `https://www.run.house/resources?type=cluster` works, that way the user can also see org resources by default

runhouse/resources/hardware/utils.py
```diff
@@ -528,8 +528,15 @@ def cast_last_active_timestamp(clusters: List[Dict[str, Any]]):
 def get_running_and_not_running_clusters(clusters: list):
     up_clusters, down_clusters = [], []
     for den_cluster in clusters:
-        # Display the name instead of the full Den address
-        cluster_name = den_cluster.get("name").split("/")[-1]
+        # If the cluster is in the /<username> folder, we display the name instead of the full Den address.
+        # Otherwise, we display the full Den address, to emphasize that the cluster was created as part of an
+        # organization.
+        cluster_rns_address = den_cluster.get("name")
+        cluster_name = (
```
jlewitt1, line 535: could this lead to some unexpected behavior? (ex: a user decides to put their username in the cluster name for an org). any harm in just displaying the full RNS address for all clusters?

BelSasha: yeah good point. 
I guess we could print the whole RNS address for all clusters but then:

1. The table in the cli output might be a bit wide, in case there are long cluster names
2. we already have the table header which informs that the clusters are associated with a specific user name (`Clusters for <username> (Running: 0/0, Total Displayed: 7/7)`), so i guess i wanted to emphasize on the clusters that are part of an org. 

A fix for the unexpected behavior is to check if the username is the part of the folder part of the `rns_address`, but it requires a bit more parsing:
```
cluster_rns_address = den_cluster.get("name")
folder, cluster_name = cluster_rns_address.split("/")[0], cluster_rns_address.split("/")[1]
cluster_name = (
    cluster_name
    if rns_client.username in folder
    else cluster_rns_address
)
```

wdyt? @jlewitt1

BelSasha: output with the whole rns_address:
image

jlewitt1: seems ok to keep the full rns address as is, also keeps things consistent with the way we display it in the UI

jlewitt1 (approved):

PR #1739, opened 2025-02-03
collect cpu metrics of multinode cluster

BelSasha:


BelSasha:
* **#1739** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1739?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1739?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

runhouse/rns/rns_client.py
```diff
@@ -25,16 +25,22 @@
 logger = get_logger(__name__)
 
 
-# This is a copy of the Pydantic model that we use to validate in Den
+# This is a copy of the Pydantic models that we use to validate in Den
+class WorkerResourceUsage(BaseModel):
+    ip: Optional[str] = None  # for local cluster tests
```
jlewitt1, line 30: similar comment as Den, can we make this required and use fake IP for testing?

runhouse/servers/cluster_servlet.py
```diff
@@ -647,6 +663,13 @@ async def astatus(self, send_to_den: bool = False) -> Tuple[Dict, Optional[int]]
         import psutil
 
         config_cluster = copy.deepcopy(self._cluster_config)
+        compute_properties = config_cluster.get("compute_properties", None)
+        ips = compute_properties.get("ips") if compute_properties else None
+        internal_ips = (
+            compute_properties.get("internal_ips") if compute_properties else None
+        )
+        is_multinode = len(ips) > 1 if ips else False
+        head_ip = ips[0] if ips else None
```
jlewitt1, line 672: if there are no IPs mybe we raise an error? is there a use case where this would expected to be None?

runhouse/servers/cluster_servlet.py
```diff
@@ -692,15 +712,65 @@ async def astatus(self, send_to_den: bool = False) -> Tuple[Dict, Optional[int]]
...
             for k in relevant_memory_info.keys()
         }
 
+        head_worker_resource_usage = {
+            "ip": head_ip,
+            "server_cpu_utilization": cpu_utilization_head_node,
+            "server_memory_usage": memory_usage_head_node,
+        }
+
+        workers = [head_worker_resource_usage]
```
jlewitt1, line 736: maybe call this "workers_resource_usage"?

runhouse/servers/cluster_servlet.py
```diff
@@ -692,15 +712,65 @@ async def astatus(self, send_to_den: bool = False) -> Tuple[Dict, Optional[int]]
...
+
+        # if the cluster is multinode, calculate the cpu_utilization and memory usage of the nodes that are not the head node.
+        if is_multinode and ips:
+            for ip_index in range(1, len(ips)):
```
jlewitt1, line 740: assuming we validate that ips are not None, i think we can remove the if statement and just do: 
`for ip in ips[1:]:`

runhouse/servers/cluster_servlet.py
```diff
@@ -692,15 +712,65 @@ async def astatus(self, send_to_den: bool = False) -> Tuple[Dict, Optional[int]]
...
+                    require_outputs=True,
+                )
+                worker_cpu = float(worker_cpu[1].strip())
+                worker_mem_cmd = 'python3 -c "import psutil; print(psutil.virtual_memory()._asdict())"'
+                worker_mem = await obj_store.arun_bash_command_on_node_or_process(
+                    command=worker_mem_cmd,
+                    node_ip=worker_internal_ip,
+                    require_outputs=True,
+                )
+                worker_mem = ast.literal_eval(
```
jlewitt1, line 759: could we use json.loads() or shlex.split() here? ( depending on the format this comes back in)

runhouse/servers/cluster_servlet.py
```diff
@@ -692,15 +712,65 @@ async def astatus(self, send_to_den: bool = False) -> Tuple[Dict, Optional[int]]
...
         # Fields: `total`, `available`, `percent`, `used`, `free`, `active`, `inactive`, `buffers`, `cached`, `shared`, `slab`
         # according to psutil docs, percent = (total - available) / total * 100
-        memory_usage = psutil.virtual_memory()._asdict()
+        # we now it is memory usage of the head node because it is calculated on the cluster_servlet which is
+        # initialized on the head node
+        memory_usage_head_node = psutil.virtual_memory()._asdict()
 
-        memory_usage = {
-            relevant_memory_info[k]: memory_usage[k]
+        memory_usage_head_node = {
```
jlewitt1, line 725: this might be slightly cleaner? 
```
memory_usage = {v: memory_usage[k] for k, v in relevant_memory_info.items()}
```

runhouse/servers/cluster_servlet.py
```diff
@@ -692,15 +712,65 @@ async def astatus(self, send_to_den: bool = False) -> Tuple[Dict, Optional[int]]
...
+            relevant_memory_info[k]: memory_usage_head_node[k]
             for k in relevant_memory_info.keys()
         }
 
+        head_worker_resource_usage = {
```
jlewitt1, line 730: shoudln't this (+ `worker_resource_usage`) be  a `WorkerResourceUsage` object instead of a dict?

runhouse/servers/cluster_servlet.py
BelSasha: actually we need the ip index, in order to match between the internal and external IPs: when we call `obj_store.arun_bash_command_on_node_or_process` we need to provide the internal ip, but we store the external IP in the status output (had a discussion about it on the Den PR). 

if we'll do `for ip in ips[1:]:` we could use `ips.index(ip)` but it is redundant parsing in my opinion.

runhouse/servers/cluster_servlet.py
```diff
@@ -424,6 +427,12 @@ async def aperiodic_autostop_check(self):
                     logger.debug("Successfully updated autostop")
 
             except Exception as e:
+                if (
+                    "Failed to look up actor with name" in str(e)
```
jlewitt1, line 431: how come we're only checking for this specific exception? this is only relevant if hte other nodes are not yet ready?

runhouse/servers/cluster_servlet.py
```diff
@@ -662,6 +677,17 @@ async def astatus(self, send_to_den: bool = False) -> Tuple[Dict, Optional[int]]
         import psutil
 
         config_cluster = copy.deepcopy(self._cluster_config)
+        compute_properties = config_cluster.get("compute_properties", None)
+        ips = (
```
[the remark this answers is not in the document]
BelSasha: for BYO clusters we don't have `compute_properties`, therefore the IPs i this case are stored as there own field in the cluster config.

runhouse/servers/cluster_servlet.py
BelSasha: yeah, if the nodes are ready but we could not find the actor we are looking for, this means the actor died and there is an actual problem we need to investigate. In this case we are raising the exception that the actor could not be found.

runhouse/servers/cluster_servlet.py
```diff
@@ -81,6 +81,8 @@ async def __init__(
         # been created yet, since it is a freshly initialized cluster servlet. In this case, we will return an empty string.
         self._is_log_file_ready = False
 
+        self._are_multinode_servlets_ready = False
```
jlewitt1, line 84: whats the signfiicance of this variable if its not a multinode cluster? (we also have the `is_multinode` variable)

runhouse/servers/cluster_servlet.py
```diff
@@ -677,45 +703,89 @@ async def astatus(self, send_to_den: bool = False) -> Tuple[Dict, Optional[int]]
...
+        # if the cluster is multinode, calculate the cpu_utilization and memory usage of the nodes that are not the head node.
+        if is_multinode and ips:
+            for ip_index in range(1, len(ips)):
+                worker_internal_ip = internal_ips[ip_index]
+                worker_ip = ips[ip_index]
+                cpu_util_cmd = (
+                    'python3 -c "import psutil; print(psutil.cpu_percent(interval=0))"'
+                )
+                # todo [SB]: check that output status code is 0? (worker_cpu[0] == 0). same for next cmds.
+                worker_cpu = await obj_store.arun_bash_command_on_node_or_process(
```
jlewitt1, line 761: i assume this won't take that long to do sequentially, but instead of running these commands in parallel across each node we could do it in parallel

runhouse/servers/cluster_servlet.py
BelSasha: Since the creation on the node_servlets are async, there might be a case when the head node trying to collect the status from those nodes before they are ready, and then we are gonna get a `Failed to look up actor` exception. If this is the first round of cluster status checks, this exception is acceptable, and it should not be logged / cause any breakage. This variable helps us in deciding whether to ignore the exception or not.

`is_multinode` is just a flag that helps parsing the status data, and a nice-to-know cluster information.  (If we would like to know whether the cluster is multi-node we could just get the `is_multinode` value, instead of checking the amount of cluster IPs.

jlewitt1 (approved):

BelSasha:
### Merge activity

* **Feb 18, 4:28 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/1739).

PR #1740, opened 2025-02-03
Collect gpu metrics for multinode

BelSasha:
CI is passing after additional changes made in the stack, see #1752

### DIFF
--- a/runhouse/servers/cluster_servlet.py
+++ b/runhouse/servers/cluster_servlet.py
@@ -37,8 +37,8 @@ from runhouse.servers.http.auth import AuthCache
 from runhouse.servers.http.http_utils import CreateProcessParams
 from runhouse.utils import (
     ColoredFormatter,
-    get_gpu_usage,
     get_pid,
+    parse_gpu_usage,
     ServletType,
     sync_function,
 )
@@ -615,7 +615,7 @@ class ClusterServlet:
                         total_memory = memory_info.total  # in bytes
                         used_memory = memory_info.used  # in bytes
                         free_memory = memory_info.free  # in bytes
-                        utilization_percent = util_info.gpu / 1.0  # make it float
+                        utilization_percent = float(util_info.gpu)  # make it float
 
                         # to reduce cluster memory usage (we are saving the gpu_usage info on the cluster),
                         # we save only the most updated gpu usage. If for some reason the size of updated_gpu_info is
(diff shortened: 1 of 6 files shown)


runhouse/servers/node_servlet.py
```diff
@@ -1,18 +1,56 @@
...
-    async def __init__(self):
+    async def __init__(self, **kwargs):
         # Still need the object store to communicate with ClusterServlet and other actors
         await obj_store.ainitialize(
             servlet_name=None,
             has_local_storage=None,
             setup_cluster_servlet=ClusterServletSetupOption.GET_OR_FAIL,
         )
 
+        self.name = kwargs.get("name", None)
```
jlewitt1, line 34: any reason to pass this in as kwargs instead of a dedicated argument?

jlewitt1: maybe for clarity we name this class property `node_name`? bc there's cluster `name` which we also reference

runhouse/utils.py
```diff
@@ -895,14 +899,42 @@ def get_gpu_usage(collected_gpus_info: dict, servlet_type: ServletType):
...
         gpu_usage["gpu_count"] = len(gpus_indices)
-        gpu_usage["utilization_percent"] = gpu_utilization_percent
+        gpu_usage[
+            "utilization_percent"
+        ] = gpu_utilization_percent  # value can be from 0 to 100
 
     return gpu_usage
+
+
+def parse_gpu_usage(collected_gpus_info: dict, servlet_type: ServletType):
```
jlewitt1, line 924: `collected_gpus_info` is a list no?

runhouse/servers/cluster_servlet.py
```diff
@@ -657,21 +657,50 @@ def _periodic_gpu_check(self):
...
+                worker_gpu_usage = obj_store.call_actor_method(
+                    node_servlet, "get_gpu_metrics", send_to_den
+                )
+                worker_resource_usage = workers_usage[ip_index]
+                worker_resource_usage["gpu_usage"] = worker_gpu_usage
+                updated_workers_usage.append(worker_resource_usage)
 
-        # will be useful for multi-node clusters.
-        cluster_gpu_usage["server_pid"] = server_pid
+        updated_workers_usage = parse_gpu_usage(
```
jlewitt1, line 699: how come we parse the usage after collecting all of them instead of inside the for loop?

runhouse/servers/cluster_servlet.py
```diff
@@ -786,21 +815,16 @@ async def astatus(self, send_to_den: bool = False) -> Tuple[Dict, Optional[int]]
                 }
                 workers_resource_usage.append(worker_resource_usage)
 
-        # get general gpu usage
-        server_gpu_usage = (
-            self._get_node_gpu_usage(self.pid)
+        # get cluster gpu usage (if its multinode cluster, each we get gpu usage per node)
```
jlewitt1, line 818:
```suggestion
        # get cluster gpu usage (if it's a multinode cluster, we get gpu usage per node)
```

runhouse/servers/cluster_servlet.py
```diff
@@ -657,21 +657,50 @@ def _periodic_gpu_check(self):
...
+        head_collected_gpus_info = copy.deepcopy(self.gpu_metrics)
+
+        if (
+            not is_multinode
+            and head_collected_gpus_info is None
+            or not head_collected_gpus_info[0]
+        ):
+            return updated_workers_usage
+
+        # adding the gpu usage of the head node to the cpu head node usage, and make it the first element of the workers
```
jlewitt1, line 674: a little confused about this comment and what we're doing here - adding gpu usage to the cpu usage? what does that mean?

runhouse/utils.py
BelSasha: If we are parsing a process gpu usage, it will be a dictionary, and if we parse the workers gpu usage, it will be list (containing gpu usage of all cluster nodes)

BelSasha: but this method is not longer needed, since we are parsing the GPU usage per node / process, before returning it. (see comment below)

runhouse/servers/cluster_servlet.py
BelSasha: yeah, good point, changed the parsing to run inside the loop.

runhouse/servers/cluster_servlet.py
BelSasha: oh, yeah, indeed a confusing comment. updated to: `# Updating the head worker's resource usage to include its GPU usage.`

jlewitt1 (approved):

BelSasha:
### Merge activity

* **Feb 18, 4:28 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/run-house/runhouse/1740).

PR #1751, opened 2025-02-06
config handling updates

jlewitt1:


runhouse/rns/defaults.py
```diff
@@ -26,7 +26,7 @@ class Defaults:
     # TODO [DG] default sub-dicts for various resources (e.g. defaults.get('cluster').get('resource_type'))
     BASE_DEFAULTS = {
         "default_folder": "~",
-        "default_provider": "cheapest",
+        "default_provider": None,
```
jlewitt1, line 29: thought here is that since this is only a local launcher feature we turn it off by default

runhouse/rns/login.py
```diff
@@ -194,6 +194,51 @@ def login(
         return token
 
 
+def _set_local_config_defaults(local_config: dict, den_config: dict):
```
jlewitt1, line 197: since we save the base defaults in den, thinking we save them locally as well. otherwise this will create mismatches when there aren't really (between local and den configs)

runhouse/rns/login.py
```diff
@@ -194,6 +194,51 @@ def login(
...
+    """Set base default values in the local config to match those that will be saved in Den if not explicitly set by
+    the user."""
+    if not local_config.get("username") and den_config:
```
carolineechen, line 200: could we just wrap this in a for loop?

```
for property in ["username", "default_folder",. ...]:
    if not local_config.get(property) and den_config:
         value = den_config.get(property, configs.BASE_DEFAULTS[property])
         configs.set(property, value)
```

carolineechen (approved):

PR #1770, opened 2025-02-16
raising DiskOutOfSpace exception

BelSasha:


BelSasha:
* **#1770** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1770?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1770?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

runhouse/utils.py
```diff
@@ -369,7 +369,7 @@ def run_command_with_password_login(
         command_run.logfile_read = sys.stdout
 
     # If CommandRunner uses the control path, the password may not be requested
-    next_line = command_run.expect(["assword:", pexpect.EOF])
```
jlewitt1, line 372: i think this was intentional to use more fuzzy pattern matching, but could be more robust to use regex matching, something like `r"[Pp]assword:"`

runhouse/resources/hardware/utils.py
```diff
@@ -290,6 +297,46 @@ def detect_cuda_version_or_cpu(cluster: "Cluster" = None, node: Optional[str] =
...
+    disk_usage = cluster.run_python(
+        commands=["import psutil", "print(psutil.disk_usage('/')._asdict())"],
+        node=node,
+        stream_logs=False,
+        conda_env_name=cluster.conda_env_name,
+    )[0]
+    if node == "all":
+        # If we sync the object to all cluster nodes, we get a list of disk_usage dictionaries, one per node.
+        # Therefore, we need to iterate through each dictionary to ensure that every node has enough disk space.
+        is_disk_sufficient = any(
```
carolineechen, line 328: this should be `all` not `any` for checking each node

runhouse/resources/hardware/cluster.py
```diff
@@ -1486,6 +1488,15 @@ def rsync(
         if not src_node and up and not Path(source).expanduser().exists():
             raise ValueError(f"Could not locate path to sync: {source}.")
 
+        # before syncing the object, making sure there is enough disk space for it. If we not preform this check in
```
carolineechen, line 1491:
```suggestion
        # before syncing the object, making sure there is enough disk space for it. If we don't perform this check in
```

runhouse/__init__.py
```diff
@@ -38,6 +38,8 @@
 )
 from runhouse.utils import sync_function
 
+from .exceptions import InsufficientDisk
```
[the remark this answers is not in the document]
BelSasha: Well the motivation was so the user could easily catch this exception. For example:
```
import runhouse as rh
cluster = rh.cluster(name="my-cluster")
try:
    # some logic
except rh.InsufficientDisk:
   # handle the exception 
```

Having said that, we could catch the exception by `runhouse.exceptions.InsufficientDisk`. Therefore removing the exposure of the exception.

carolineechen (approved):

PR #1799, opened 2025-03-03
add custom name handling to ssh secrets

jlewitt1:


mkandler:
Nice, this feels much more intuitive/clear compared to my quick fix

mkandler (approved):

runhouse/resources/hardware/cluster.py
```diff
@@ -398,8 +399,8 @@ def from_name(
             _resolve_children=_resolve_children,
         )
         if cluster and cluster._creds and not dryrun:
+            from runhouse import SSHSecret
```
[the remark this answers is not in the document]
carolineechen: not sure if `SSHSecret` needs to be user facing, and we also generally import internally from more nested files rather from the main runhouse package, though not sure if there's a strong reason for that

runhouse/resources/secrets/provider_secrets/ssh_secret.py
```diff
@@ -46,6 +47,24 @@ def from_config(config: dict, dryrun: bool = False, _resolve_children: bool = Tr
...
+    @classmethod
+    def from_name(
+        cls,
+        name,
+        provider: str = None,
+        load_from_den: bool = True,
+        dryrun: bool = False,
+        _resolve_children: bool = True,
+    ):
+        name = cls.ensure_user_prefix(name)
```
carolineechen, line 59: I think we also need to add this into the `__init__` for SSHSecret for consistency. since it's possible to construct a secret outside of the `from_name` pipeline

carolineechen: ProviderSecret's from_name actually uses `return provider_class(name=name, provider=provider, dryrun=dryrun)` rather than `xxx.from_name`.

To test, `rh.provider_secret("ssh").config()` should have name of the given format but is empty currently

runhouse/resources/hardware/cluster.py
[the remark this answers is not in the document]
jlewitt1: yea good point, doesnt need to be user facing. can just use the import method you gave above

runhouse/resources/secrets/provider_secrets/ssh_secret.py
jlewitt1: yea i added that bit to the init as well

PR #1804, opened 2025-03-05
rsync accept node_idx and 'head' as node arg

BelSasha:


runhouse/resources/hardware/utils.py
```diff
@@ -277,6 +277,22 @@ def _setup_creds_from_dict(ssh_creds: Dict, cluster_name: str):
...
+    elif node == "head":
+        return ips[0]
+
+    elif isinstance(node, str) and node.isnumeric():
+        node = int(node)
+
+    if isinstance(node, int) and node < len(ips):
+        return ips[node]
+
+    raise ValueError(f"Node {node} is unsupported, could not get it's IP.")
```
carolineechen, line 293:
```suggestion
    raise ValueError(f"Node {node} is unsupported, could not get its IP.")
```

carolineechen (approved):

PR #1815, opened 2025-03-11
add validation to reqs arg in cluster install pkgs method

BelSasha:


BelSasha:
* **#1815** <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1815?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/run-house/runhouse/1815?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

runhouse/resources/hardware/cluster.py
```diff
@@ -832,6 +832,8 @@ def install_packages(
             >>> cluster.pip_install(reqs=["accelerate", "diffusers"])
             >>> cluster.pip_install(reqs=["accelerate", "diffusers"], conda_env_name="my_conda_env", override_remote_version=True)
         """
+        reqs = [reqs] if isinstance(reqs, str) else reqs
```
carolineechen, line 835: could we update the type checking in the function arg too? ` Union[str, List[str]]`

BelSasha: Also added support for passing a single `rh.Package` instance (and not a list of `rh.Package` instances).

carolineechen (commented):
actually do you think we could hold off on this change? there's a couple things
* with the addition of `pip_install`, `conda_install`, etc, it may make sense to eventually turn `install_packages` internal facing, in which case we might not need to update the signature
* on that note, if we want to support single packages it probably makes more sense to do it inside those more specific functions, which can be more vague (vs `install_packages` does indicate plural)
* I don't think we need to support `Package` input type for the individual functions, those should be abstracted away internally and push users towards just using strings rather than constructing packages. 
* and given the new direction, maybe we'll end up adjusting these APIs too so I think we can keep this in mind but no need to implement it now?

PR #1824, opened 2025-05-22
Cleanup of examples

py-rh:
* Improved text 
* Image vs image
* GPUs

### DIFF
--- a/examples/dask-preprocessing-and-training/lightgbm_dask.py
+++ b/examples/dask-preprocessing-and-training/lightgbm_dask.py
@@ -129,23 +129,23 @@ if __name__ == "__main__":
     # ## Define compute with multiple nodes
     num_nodes = 3
 
-    # The environment for the remote compute
-    img = kt.images.dask().pip_install(
+    # The environment for the remote compute, and the compute.
+    # Importantly, we use .distribute("dask") to start the Dask cluster and indicate this will be used with Dask
+
+    img = kt.Image(image_id="ghcr.io/dask/dask").pip_install(
         [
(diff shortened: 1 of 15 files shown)


examples/dask-preprocessing-and-training/lightgbm_dask.py
```diff
@@ -129,23 +129,23 @@ def predict(self, X):
     # ## Define compute with multiple nodes
     num_nodes = 3
 
-    # The environment for the remote compute
-    img = kt.images.dask().pip_install(
+    # The environment for the remote compute, and the compute.
+    # Importantly, we use .distribute("dask") to start the Dask cluster and indicate this will be used with Dask
```
carolineechen, line 133:
```suggestion
    # Use .distribute("dask") to start the Dask cluster and indicate this will be used with Dask
```

carolineechen (commented):
more general question on docs and examples and explanations - how in depth do we want to explain what's going on behind the scenes and sprinkling in value prop or other use cases, vs presenting it as a pretty self explanatory, easy to follow piece of code? I don't have the answer but as we start building out docs, I think we should be intentional with how we're using each collateral, as standalone or in the bigger picture and make sure it serves its purpose.

for instance, in the latter, we can use clean concise examples where the code speaks for itself, and other docs like the home page (sales value prop) or api docs (figuring out what kind of things are also supported) supplement examples. whereas fitting more info in does provide a bit more content and would make sense if examples was the core piece of docs that we share. asking bc currently it feels a bit in between to me

PR #1845, opened 2025-10-31
Update kt deploy reload flow

carolineechen:
we should only instantiate compute during `kt deploy <file>` cli workflow. otherwise, when importing, leave compute empty as it will be reloaded `from_name` when the module is called. this avoid unnecessary computation (esp once template building gets moved to compute init) or potentially out of date/mismatched compute objects that don't match 1-to-1 with the existing deployed compute.

### DIFF
--- a/python_client/kubetorch/cli.py
+++ b/python_client/kubetorch/cli.py
@@ -443,6 +443,7 @@ def kt_deploy(
     @kt.compute in the file or module."""
     from kubetorch.resources.compute.utils import _collect_modules
 
+    os.environ["KT_CLI_DEPLOY_MODE"] = "1"
     to_deploy, target_fn_or_class = _collect_modules(target)
 
     if not target_fn_or_class:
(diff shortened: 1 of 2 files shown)


carolineechen:
* **#1888** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/1888?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1846** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/1846?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1845** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/1845?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/run-house/kubetorch/1845?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

dongreenberg (approved):
The code looks fine, but can you explain why we need this a bit more, and why we wouldn't use the same codepath to load a resource by name if we're not inside `kt deploy`?

PR #1864, opened 2025-11-05
convert cluster scoped logs RBAC to namespaced scoped

jlewitt1:


jlewitt1:
* **#1874** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/1874?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1864** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/1864?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/run-house/kubetorch/1864?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

charts/kubetorch/templates/log-streaming/logs-deployment.yaml
```diff
@@ -34,37 +35,86 @@ spec:
...
+    app: kubetorch-logging
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: kubetorch-logging
+  namespace: {{ .Release.Namespace }}
+rules:
+  - apiGroups: [""]
+    resources: ["pods", "services", "endpoints", "configmaps"]
```
mkandler, line 54: Shouldn't these also be in the namespaces loop? Or do we only need this for the kubetorch namespace not default etc? What's the reason for moving these to a different service account that's added in a new way?

charts/kubetorch/templates/log-streaming/logs-deployment.yaml
```diff
@@ -34,37 +35,86 @@ spec:
...
+    namespace: {{ .Release.Namespace }}
 {{- end }}
+
 {{- if (index .Values "opentelemetry-collector").enabled }}
 ---
-# RBAC for existing OpenTelemetry Collector Agent
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: kubetorch-opentelemetry-collector
```
mkandler, line 77: Why didn't we have to create this ServiceAccount before?

mkandler (commented):
Curious about the changes in logging, also I think the old ClusterRoles will remain after a helm upgrade, so you'll definitely want to test this out on a fresh cluster if you haven't yet

charts/kubetorch/templates/log-streaming/logs-deployment.yaml
jlewitt1: the otel subchart helm dependency automatically creates it by default, but it does that using a clusterrole and clusterrolebinding which is cluster wide

charts/kubetorch/templates/log-streaming/logs-deployment.yaml
jlewitt1: Those are the minimum permissions for the OTEL collector that don't really need to be cluster-wide - OTEL only needs to watch the namespaces it’s configured to collect logs from

charts/kubetorch/templates/configmaps/kubetorch-config.yaml
```diff
@@ -4,7 +4,6 @@ metadata:
   name: kubetorch-config
   namespace: {{ .Release.Namespace }}
 data:
-  DEFAULT_SERVICE_NAMESPACE: "{{ .Values.kubetorchConfig.environment.DEFAULT_SERVICE_NAMESPACE | default "default" }}"
```
jlewitt1: noting this also doesn't seem to be used anywhere

charts/kubetorch/values.yaml
```diff
@@ -1,5 +1,4 @@
 kubetorchConfig:
-  environment:
```
jlewitt1: also noting don't think we need this anymore

charts/kubetorch/templates/log-streaming/logs-deployment.yaml
mkandler: Doesn't that need to be all the deployment namespaces though, not just the installation one? I think we might need to loop `.Values.kubetorchConfig.deployment_namespaces` to do that. E.g. if deployment_namespaces is `['kubetorch', 'default']` we'd need permissions for both, not just kubetorch namespace

jlewitt1: the logging service is only deployed in the kubetorch namespace at the moment, only the rsync service needs to be per namespace. or is this in refenrce to somethning else?

mkandler: Even though it's deployed in the kubetorch namespace, does it not need RBAC permissions for services (pods, etc) we've launched in other namespaces? Is that not how logs are fetched?

jlewitt1: im pretty sure that traffic goes over the cluster network — not through the K8s API, so RBAC isn't needed. meaning it never calls the API server to list or watch Pods elsewhere. also this is definitely working in testing, but will dig more to confirm this is the case

BelSasha (approved):

PR #1872, opened 2025-11-06
more compute intensive hello world example

jlewitt1:


jlewitt1:
* **#1872** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/1872?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/run-house/kubetorch/1872?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

README.md
```diff
@@ -11,21 +11,38 @@ Since Kubetorch has no local runtime or code serialization, you can access large
...
+    for i in range(num_iterations):
+        a = np.random.rand(matrix_size, matrix_size)
+        b = np.random.rand(matrix_size, matrix_size)
+
+        product = a @ b
+        iteration_sum = np.sum(product)
+        total_sum += iteration_sum
+
+        print(f"Iteration {i+1}: sum={iteration_sum:.3f}")
+        time.sleep(1)
```
carolineechen: down for a slightly more involved hello world that numpy to showcase the compute/image, but I do think the content here in the for loop is not super necessary and the focus should be on the kt portion. thoughts on cutting it down more simply? e.g.

```
for i in range(num_iterations):
     total_sum += np.rand()
     print(f"Iteration {i+1}: sum={iteration_sum:.3f}")
 
     time.sleep(1)
```

carolineechen (approved):
left a comment on potentially simplifying the numpy function. but don't feel super strongly if you think leaving it as is showcases the use case better

README.md
jlewitt1: i agree the code isn't so meaningful, the only benefit is that it allows us to stream utilization metrics back.. so i'm a little torn

README.md
```diff
@@ -13,19 +13,23 @@ Since Kubetorch has no local runtime or code serialization, you can access large
 ```python
 import kubetorch as kt
 
-def hello_world():
-    return "Hello from Kubetorch!"
+def simulate_workload(num_iterations: int = 5, vector_size: int = 10000):
+    import numpy as np
+
+    return sum(np.dot(np.random.rand(vector_size), np.random.rand(vector_size)) for _ in range(num_iterations))
```
jlewitt1, line 19: @carolineechen better?

carolineechen: yup, can we also pass in new (kw)args to the `remote_fn` call below to also show that the function takes in args just as the original python function does?

PR #1882, opened 2025-11-10
Update kt config behavior

carolineechen:
fix overwriting local username in tests
- don't write down config during port forward
- don't write down config based on local values in cli tests, but based on original file

in the process, update kt config write/set/unset behavior to only modify a specific key if provided, rather than keys potentially set through env vars

### DIFF
--- a/python_client/kubetorch/globals.py
+++ b/python_client/kubetorch/globals.py
@@ -98,7 +98,7 @@ def _ensure_pf(service_name: str, namespace: str, remote_port: int, health_endpo
             cluster_config = wait_for_port_forward(proc, local_port, health_endpoint=health_endpoint)
             if isinstance(cluster_config, dict):
                 config.cluster_config = cluster_config
-                config.write()
+                config.write(values={"cluster_config": cluster_config})
         else:
             # Minimal TCP wait (no HTTP probe)
             deadline = time.time() + 10
--- a/python_client/kubetorch/cli.py
+++ b/python_client/kubetorch/cli.py
@@ -341,8 +341,8 @@ def kt_config(
             raise typer.Exit(1)
 
         try:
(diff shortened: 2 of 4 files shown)


carolineechen:
* **#1882** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/1882?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/run-house/kubetorch/1882?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python_client/kubetorch/globals.py
```diff
@@ -98,7 +98,6 @@ def _ensure_pf(service_name: str, namespace: str, remote_port: int, health_endpo
             cluster_config = wait_for_port_forward(proc, local_port, health_endpoint=health_endpoint)
             if isinstance(cluster_config, dict):
                 config.cluster_config = cluster_config
```
carolineechen, line 100: @mkandler should we be writing `cluster_config` into `.kt/config.yaml` here or is it fine to just have it set locally in the config global? calling `config.write()` here was causing username to be overwritten in tests/more generally, I updated the `kt.write()` to be able to take in a specific value and override only that (ignoring any env vars taking precedence), so we could call `kt.write(values={"cluster_config": cluster_config})` but not sure if this is necessary or desired behavior?

mkandler: I think for this to work reliably we do want to write the cluster_config after it's set in this case, since it's referenced in places like the compute init method. There's probably a better approach overall for maintaining those values locally, but probably need to keep a `kt.write(values={"cluster_config": cluster_config})` for now

PR #1892, opened 2025-11-11
notebook cli cmd and function support

jlewitt1:


jlewitt1:
* **#1899** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/1899?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1892** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/1892?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/run-house/kubetorch/1892?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python_client/kubetorch/cli.py
```diff
@@ -1698,6 +1698,243 @@ def kt_volumes(
...
+    compute_kwargs = {"namespace": namespace}
+
+    if cpus:
+        compute_kwargs["cpus"] = cpus
+    if memory:
+        compute_kwargs["memory"] = memory
+    if gpus:
+        compute_kwargs["gpus"] = gpus
+    if inactivity_ttl:
+        compute_kwargs["inactivity_ttl"] = inactivity_ttl
```
carolineechen: nit: the default `None` for these `if args:` is the same as the compute default so this could just be 

```
    compute_kwargs = {
        "namespace": namespace,
        "cpus": cpus,
        "memory": memory,
        "gpus": gpus,
        "inactivity_ttl": inactivity_ttl,
    }
```

python_client/kubetorch/cli.py
```diff
@@ -1698,6 +1698,243 @@ def kt_volumes(
...
+    original_port = local_port
+    for i in range(5):
+        if is_port_available(local_port):
+            break
+        local_port += 1
+    else:
+        console.print(f"\n[red]Ports {original_port}-{original_port + 4} are all in use.[/red]")
+        raise typer.Exit(1)
+    if local_port != original_port:
+        console.print(f"[yellow]Port {original_port} already in use, using port {local_port} instead.[/yellow]")
```
carolineechen:
```suggestion
    try:
        local_port = find_available_port(local_port, max_tries=5)
        if local_port != original_port:
            console.print(f"[yellow]Port {original_port} already in use, using port {local_port} instead.[/yellow]")
    except RuntimeError:
        console.print(f"\n[red]Ports {original_port}-{original_port + 4} are all in use.[/red]")
        raise typer.Exit(1)
```

python_client/kubetorch/cli_utils.py
```diff
@@ -953,3 +953,10 @@ def create_table_for_output(columns: List[set], no_wrap_columns_names: list = No
             table.add_column(name, style=style)
 
     return table
+
+
+def notebook_placeholder():
+    """Placeholder function to launch notebook service"""
+    import time
+
+    time.sleep(3600)  # Keep alive for port forwarding
```
carolineechen, line 962: does this mean the notebook will only stay alive for 1 hr?

jlewitt1: yea but i think ok? bc this is for a more dummy use case where the user doesnt specify any existing service to launch into, we're just creating a new one with this placeholder to stiil create the notebook instance

python_client/kubetorch/cli.py
```diff
@@ -1675,6 +1675,245 @@ def kt_volumes(
...
+        if mismatches:
+            console.print("[yellow]Cannot reuse existing notebook due to mismatched parameters:[/yellow]")
+            for key, existing, requested in mismatches:
+                display_existing = existing if existing is not None else "<default>"
+                console.print(f"  - [bold]{key}[/bold]: existing = '{display_existing}', requested = '{requested}'")
+            console.print(
+                f"\n[yellow]Delete the existing notebook service ([bold]`kt teardown {service_name} -y`[/bold]) "
+                "or create a new one with a different name.[/yellow]"
+            )
+            return
```
carolineechen, line 1803: I think this should be moved after we call `.to` and define the new compute (`compute = remote_fn.compute`, updated in the PR on top of this). currently this basically just doing
```
compute = kt.Compute(**kwargs)
for key, val in kwargs:
     assert compute.key = val
```
which should never fail

jlewitt1: oh yep good call

PR #1907, opened 2025-11-14
expand python version support

jlewitt1:


python_client/pyproject.toml
```diff
@@ -28,7 +29,7 @@ include = [
 ]
 
 [tool.poetry.dependencies]
-python = ">=3.9,<=3.13"
+python = ">=3.9,<=3.15"
```
carolineechen: we can also just do `"^3.9"` to indicate >= 3.9 and not set an upper bound

PR #1944, opened 2025-12-03
make volume mount path settable and add as reload param

jlewitt1:


jlewitt1:
* **#1944** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/1944?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/run-house/kubetorch/1944?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python_client/kubetorch/resources/volumes/volume.py
```diff
@@ -132,8 +153,20 @@ def from_name(
         create_if_missing: bool = False,
         namespace: str = None,
         core_v1: client.CoreV1Api = None,
+        mount_path: str = None,
     ) -> "Volume":
-        """Get existing volume or optionally create it"""
+        """Get existing volume or optionally create it
```
carolineechen: using `from_name` with a `create_if_missing` seems a little unintuitive + is `False` in all of our internal code, is there a strong use case to keep and support this? also seeing we're setting 5Gi for the size if creating it, is this a standard size or would we want to add that as an arg or document in `create_if_missing` docstring if continuing to support it

jlewitt1: yea fair point, i think creating a new one on reload is prob unintuitive

PR #2009, opened 2025-12-21
Allow custom annotations for kubetorch service account

lyonsy:


jlewitt1 (approved):
Removed the chart version bump since we’re likely rolling this into an upcoming major release

PR #2015, opened 2025-12-22
Optional queuing with kueue

mkandler:


mkandler:
* **#2015** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2015?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2015?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python_client/kubetorch/resources/compute/compute.py
```diff
@@ -167,6 +172,7 @@ def __init__(
         self._volumes = volumes
         self._logging_config = logging_config or LoggingConfig()
         self._manifest = None
+        self._queue_name = queue_name
```
carolineechen: I think we can get rid of setting `_queue_name` variable since we have the `queue_name` property that extracts or sets the value from the manifest - to avoid any discrepancies between the two and treat the manifest value as the source of truth

mkandler: Good call, improved the logic to check the manifest in the BYO case too!

carolineechen (approved):

PR #2048, opened 2025-12-28
More updates to tests

BelSasha:


BelSasha:
* **#2048** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2048?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2048?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

charts/kueue/kueue-resources.yaml
```diff
@@ -34,12 +34,16 @@ spec:
   #   AKS: kubernetes.azure.com/accelerator: nvidia
   #   Generic: nvidia.com/gpu.present: "true"
   nodeLabels:
-    nvidia.com/gpu.present: "true"
+    type: "gpu"
+  tolerations:
   # Optional: Add tolerations if your GPU nodes have taints
-  # tolerations:
```
jlewitt1, line 39: lets keep these as optional by default?

charts/kueue/kueue-resources.yaml
```diff
@@ -34,12 +34,16 @@ spec:
   #   AKS: kubernetes.azure.com/accelerator: nvidia
   #   Generic: nvidia.com/gpu.present: "true"
   nodeLabels:
-    nvidia.com/gpu.present: "true"
```
jlewitt1, line 37: why cant we leave this whole file as it was previously?

BelSasha: because otherwise the service won't be scheduled a gpu. The generic nodeLabel is not nvidia.com/gpu.present is not the generic label, for GKE clusters at least.

jlewitt1: but thats for our specifirc use case? the pooint being for oss we prob don't want to be so prescriptive

BelSasha: Hmm i see your point, i guess uses could change the labels and tolerations according to their gpu node pools configuration. Reverting.

python_client/tests/test_controller.py
```diff
@@ -469,7 +469,7 @@ def test_integration_full_stack():
     try:
         # Deploy with resources - (pass Secret object instead of string to avoid provider inference)
         remote_fn = kt.fn(get_env_var, name=service_name).to(
-            kt.Compute(cpus=".01", gpu_anti_affinity=True, secrets=[secret], volumes=[vol])
+            kt.Compute(cpus=".01", gpu_anti_affinity=True, secrets=[secret], volumes=[vol], launch_timeout=300)
```
BelSasha, line 472: increased the launch timeout, so it'll be created in case the service triggered the autoscaler to created a new node.

PR #2052, opened 2025-12-29
add more control over autoscaler pod eviction behavior for distributed jobs

jlewitt1:
Distributed training jobs (PyTorchJob, TFJob, etc.) were ocassionally failing with `WorkerMembershipChanged` errors when the cluster autoscaler evicted pods during execution. Pod eviction breaks quorum and fails the entire job.

Added `compute.safe_to_evict` property that gets/sets the `cluster-autoscaler.kubernetes.io/safe-to-evict` annotation for the workload. 
 
When a user sets `compute.safe_to_evict = False`:
  - Always sets annotation on primary pod template (works for any kind)
  - Also sets on worker replica if present (training jobs get both Master and Worker annotated)

### DIFF
--- a/python_client/kubetorch/resources/compute/compute.py
+++ b/python_client/kubetorch/resources/compute/compute.py
@@ -1289,6 +1289,47 @@ class Compute:
                             del self.pod_spec["affinity"]
 
     @property
+    def safe_to_evict(self) -> Optional[bool]:
+        """Whether pods can be evicted by the cluster autoscaler.
+
+        Returns:
+            True if pods can be evicted, False if protected from eviction,
+            None if not explicitly set.
+        """
+        annotations = self.pod_template.get("metadata", {}).get("annotations", {})
+        value = annotations.get("cluster-autoscaler.kubernetes.io/safe-to-evict")
+        if value is None:
+            return None
+        return value.lower() == "true"
+
+    @safe_to_evict.setter
+    def safe_to_evict(self, value: bool):
+        """Set whether pods can be evicted by the cluster autoscaler.
+
+        Args:
+            value: True to allow eviction (e.g., for checkpointed training that
+                   can resume after interruption), False to prevent eviction
+                   (e.g., for jobs where pod loss would fail the entire run).
+        """
+        annotation_value = "true" if value else "false"
+
+        # Set on primary pod template
+        template_metadata = self.pod_template.setdefault("metadata", {})
+        annotations = template_metadata.setdefault("annotations", {})
+        annotations["cluster-autoscaler.kubernetes.io/safe-to-evict"] = annotation_value
+
+        # Also set on worker replica pod template if present (ex: for training jobs)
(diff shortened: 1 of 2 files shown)


jlewitt1:
* **#2052** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2052?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2052?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python_client/kubetorch/resources/compute/compute.py
```diff
@@ -1554,6 +1586,11 @@ def replicas(self, value: int):
                 }
             distributed_config["quorum_workers"] = value
             self.service_manager.set_replicas(self._manifest, value, distributed_config=distributed_config)
+
+            # Prevent cluster autoscaler from evicting pods during distributed training.
```
carolineechen: should this be set outside of the `TrainJobServiceManager` check, to apply to any case where distributed_config is set (including knative and ray)

jlewitt1: im not sure we want to blindly apply it to all distributed use cases, was intended to be only for training jobs with quorum requirements

carolineechen: then in the `distributed_config` setter should we only apply it in the train job case?

python_client/kubetorch/resources/compute/compute.py
```diff
@@ -1288,6 +1288,33 @@ def gpu_anti_affinity(self, value: bool):
...
+        return value.lower() == "true"
+
+    @safe_to_evict.setter
+    def safe_to_evict(self, value: bool):
+        """Set whether pods can be evicted by the cluster autoscaler.
+
+        Args:
+            value: True to allow eviction, False to prevent eviction.
+                   For distributed training jobs, this defaults to False when
+                   distributed_config is set, to protect quorum stability.
```
carolineechen: "distributed training jobs" here is a bit vague if we intended it to only refer to train job type (vs distributed training in knative or ray). though I don't think it really makes sense to have a default value for a setter. maybe just remove this part, or can mention the use case when user would want to set it to True/False

python_client/kubetorch/resources/compute/compute.py
jlewitt1: yea good point.. ill update to only automatically set for TrainJob kinds (PyTorchJob, TFJob, etc.) when setting distributed_config and when setting replicas. Ray, Knative, etc. i think best to be left to their own scaling/eviction behavior

jlewitt1: actually i think im going to remove this, and keep the default behavior we had before. would prefer we dont change the compute behavior without explicit user intent

carolineechen (approved):

PR #2085, opened 2026-01-06
Add kt apply cli command

carolineechen:
Add `kt apply` cli command for deploying and iteration on kubetorch services from an existing manifest and optional dockerfile instructions. kubetorch server start command is added to manifest using `Compute.from_manifest`, and the original command (from the manifest or dockerfile) is run using `kt.App`.

Additionally, add `Image.from_dockerfile()` image construction option and add `image` argument to `Compute.from_manifest`.

Note: only basic kubetorch server compatible dockerfile commands are supported as part of the image. extension + broader support tbd

### DIFF
--- a/python_client/kubetorch/cli.py
+++ b/python_client/kubetorch/cli.py
@@ -3,6 +3,7 @@ import importlib
 import inspect
 import json
 import os
+import shlex
 import signal
 import subprocess
 import sys
@@ -12,6 +13,8 @@ from datetime import datetime
 from pathlib import Path
 from typing import List
 from urllib.parse import urlparse
+
+import yaml
 
 from kubetorch.resources.compute.utils import (
     ControllerRequestError,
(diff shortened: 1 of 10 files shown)


carolineechen:
* **#2243** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2243?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2228** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2228?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2227** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2227?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2213** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2213?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2194** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2194?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2085** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2085?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2085?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2128** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2128?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2127** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2127?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

python_client/kubetorch/cli.py
```diff
@@ -1179,6 +1181,133 @@ def kt_run(
...
 
 
+@app.command("apply")
+def kt_apply(
+    manifest_path: str = typer.Argument(..., help="Path to K8s manifest YAML file"),
+    dockerfile: str = typer.Option(None, "--dockerfile", "-d", help="Path to Dockerfile for image setup"),
+    name: str = typer.Option(None, "--name", "-n", help="Service name (overrides manifest name)"),
+):
+    """
+    Apply a kubetorch service from a k8s manifest file and optional Dockerfile.
```
dongreenberg:
```suggestion
    Apply a Kubernetes resource from a YAML manifest and optional Dockerfile via Kubetorch fast deployment.
```

python_client/kubetorch/cli.py
```diff
@@ -1179,6 +1181,133 @@ def kt_run(
...
+
+    Automatically injects kubetorch server into the pod, applies the manifest, runs Dockerfile instructions
```
dongreenberg: syncing dockerfile dependencies?

python_client/kubetorch/cli.py
```diff
@@ -1179,6 +1181,133 @@ def kt_run(
...
+    if original_cmd:
+        run_cmd = original_cmd
+    elif dockerfile_cmd:
+        console.print(f"[dim]Using command from Dockerfile: {dockerfile_cmd}[/dim]")
+        run_cmd = dockerfile_cmd
+    else:
+        console.print("[yellow]Note: No command found in manifest or Dockerfile, server will run idle.[/yellow]")
+        run_cmd = "sleep infinity"
+
+    app_instance = App(
```
dongreenberg, line 1267: Ports?

python_client/kubetorch/resources/callables/module.py
```diff
@@ -588,10 +588,14 @@ def _get_rsync_dirs_and_dockerfile(self, install_url, use_editable, init_args):
...
-            source_file = Path(f"{self.pointers[0]}/{self.pointers[1]}.py")
-            rsync_dirs = [str(source_file)]
-            logger.info(f"No project markers found, syncing file {source_file}")
+            if self.pointers[1]:
+                # Use the source file (.py) instead of directory
+                source_file = Path(f"{self.pointers[0]}/{self.pointers[1]}.py")
+                rsync_dirs = [str(source_file)]
+                logger.info(f"No project markers found, syncing file {source_file}")
+            else:
+                # No project markers and no specific file - sync cwd
```
dongreenberg: I don't think we should be syncing cwd here - just what the dockerfile tells us to sync (it's the user's responsibility to include all required local code in COPY commands, there's no expectation that we automatically package code). I think that also means we don't need to pass the faux pointers into the kt.App above.

python_client/kubetorch/resources/compute/compute.py
```diff
@@ -2361,6 +2367,15 @@ def _image_setup_and_instructions(self, rsync: bool = True):
 
         logger.debug("Writing out image instructions.")
 
+        # Prepend raw dockerfile content if present, excluding CMD/ENTRYPOINT.
```
dongreenberg: Not sure we need to do this here, but I'm thinking the dockerfile itself can actually be the internal representation of the image. Meaning instead of holding our partial representations of the user commands, we just store a dockerfile instance var and build it up as the user adds commands. And then from_dockerfile just starts from a particular point before layering on other commands (and can't have an image_id, because the FROM is already populated).

python_client/kubetorch/resources/images/image.py
```diff
@@ -92,10 +92,131 @@ def __init__(
...
+                "If you need these features, build your image with `docker build` "
+                "and reference the built image in your manifest instead."
+            )
+
+        # Extract base image
+        image_id = None
+        for line in content.split("\n"):
+            line = line.strip()
+            if line.startswith("FROM "):
+                parts = line[5:].split()  # Handle image:tag format
```
dongreenberg: Why do we extract out the tag? We need the whole line for image_id anyway.

python_client/tests/test_cli.py
```diff
@@ -1595,3 +1595,189 @@ async def test_logs_cli_single_pod_wrong_name(remote_logs_fn):
...
+    dockerfile_cmd = "python -c print('hello from dockerfile')"
+
+    dockerfile_content = f"""FROM {dockerfile_base_image}
+RUN pip install requests
+CMD ["python", "-c", "print('hello from dockerfile')"]
+"""
+
+    # Create manifest with different image and name to test override behavior
+    service_name = f"{kt.config.username}-apply-dockerfile"
+    manifest_content = f"""apiVersion: apps/v1
```
dongreenberg: I'd rather test with a Pod manifest because it's something we don't already natively support in KT - I do this in the BYO compute tests too.

python_client/tests/test_cli.py
```diff
@@ -1595,3 +1595,189 @@ async def test_logs_cli_single_pod_wrong_name(remote_logs_fn):
...
+
+        assert expected_service_name in pool_names, f"Service {expected_service_name} not found in pools: {pool_names}"
+
+    finally:
+        os.unlink(manifest_path)
+
+
+@pytest.mark.level("minimal")
+@pytest.mark.asyncio
+async def test_kt_apply_knative():
```
dongreenberg: Any particular reason we need this one?

python_client/kubetorch/cli.py
```diff
@@ -1179,6 +1181,120 @@ def kt_run(
...
+    except ValueError:
+        pod_template_path = ["spec", "template"]
+
+    current = manifest
+    for key in pod_template_path:
+        if isinstance(key, int):
+            current = current[key] if len(current) > key else {}
+        else:
+            current = current.get(key, {})
+    containers = current.get("spec", {}).get("containers", [])
```
dongreenberg, line 1238: Can we just search for the first instance of containers + command + args by traversing the dictionary, rather than try to infer the exact spec path? Meaning traverse through the manifest until we see "containers", then if it contains a list, traverse through the list until we see command and/or args. Does that make sense? Also, we should all users to disable this swap functionality if they want to just start the server on their own.

python_client/kubetorch/cli.py
```diff
@@ -1179,6 +1181,120 @@ def kt_run(
...
+
+    # Construct Image from Dockerfile if provided
+    image = None
+    dockerfile_cmd = None
+    if dockerfile:
+        image = Image.from_dockerfile(dockerfile)
+        dockerfile_cmd = image.dockerfile_command
+
+    compute = Compute.from_manifest(manifest=manifest, image=image)
+    compute.freeze = True  # Don't sync over git workdir, only paths specified in the dockerfile
```
dongreenberg: Doesn't this also disable dockerfile rebuilding on the cluster? Maybe there's another way to explicitly disable workdir syncing in .to.

python_client/tests/test_cli.py
```diff
@@ -1595,3 +1595,199 @@ async def test_logs_cli_single_pod_wrong_name(remote_logs_fn):
...
+            f.write(test_file_content)
+
+        # Create a Python script to run
+        script_path = os.path.join(tmpdir, "run.py")
+        with open(script_path, "w") as f:
+            f.write(
+                """import time
+with open('/app/test_data.txt') as f:
+    print(f.read())
+time.sleep(10)
```
dongreenberg: Can we just make all these strings into files in the assets folder?

python_client/tests/test_cli.py
```diff
@@ -1595,3 +1595,199 @@ async def test_logs_cli_single_pod_wrong_name(remote_logs_fn):
...
+RUN pip install requests
+COPY test_data.txt /app/test_data.txt
+COPY run.py /app/run.py
+CMD ["python", "/app/run.py"]
+"""
+
+        service_name = f"{kt.config.username}-apply-dockerfile"
+
+        manifest_content = f"""apiVersion: apps/v1
+kind: Deployment
```
dongreenberg: Can we do this with just a Job or pod so we know it doesn't depend on being a builtin? And same for the below?

dongreenberg (approved):

python_client/kubetorch/cli.py
carolineechen: this is pending the flexible BYO work as I think there will likely be overlap with these cases. I think for the scope of this PR just to have kt apply work for the currently supported kt manifest types. once BYO is supported, I can go ahead and extend that work to kt apply and test that in a new PR as well

PR #2111, opened 2026-01-10
Auto-enable distributed execution for training jobs and RayClusters

jlewitt1:


jlewitt1:
* **#2111** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2111?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2111?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python_client/tests/test_byo_manifest.py
```diff
@@ -731,6 +731,8 @@ async def test_byo_manifest_with_endpoint_url():
             "ports": [{"port": user_port, "targetPort": DEFAULT_KT_SERVER_PORT}],
         },
     }
+    # Delete existing service if it exists (cleanup from previous test runs)
+    controller.delete_service(namespace=namespace, name=user_service_name, ignore_not_found=True)
```
carolineechen: is this not being deleted by test teardown as a byo service? we should try to resolve the error there rather than add this line here, or at least add this at the end of this test with a `finally` block to delete it after the test runs, rather than when the test is rerun again

jlewitt1: yea its bc the service is created seaparely outside of kubetorch, but you're right we should have a proper try/ finally block for this, will add

carolineechen (approved):

PR #2113, opened 2026-01-11
Aggregate controller delete call

BelSasha:


BelSasha:
* **#2136** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2136?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2113** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2113?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2113?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python_client/kubetorch/resources/compute/utils.py
```diff
@@ -249,7 +248,7 @@ def load_configmaps(
 
 # ----------------- Resource Deletion Utils ----------------- #
 def delete_configmaps(
```
jlewitt1: are we calling this anywhere?

python_client/tests/test_cli.py
```diff
@@ -78,12 +58,7 @@ def validate_logs_fn_service_info(list_output: str, service_name: str, compute_t
     assert re.search(service_expected_info, list_output, re.DOTALL)
 
 
-def validate_teardown_output(teardown_output: str, service_name: str, force_delete: bool = False):
-
-    assert "Force deleting resources..." if force_delete else "Deleting resources..." in teardown_output
-    assert f"✓ Deleted deployment {service_name}" in teardown_output
-    assert f"✓ Deleted cached data for {service_name}" in teardown_output
-
+def validate_service_no_in_kt_list(service_name: str):
```
jlewitt1:
```suggestion
def validate_service_not_in_kt_list(service_name: str):
```

python_client/tests/utils.py
```diff
@@ -519,3 +522,34 @@ def service_deployer_with_logs(service_name: str):
 
 def strip_ansi(s: str) -> str:
     return ANSI_RE.sub("", s)
+
+
+def remote_fn_for_teardown():
```
jlewitt1, line 527: why do we need these new fixtures?

python_client/kubetorch/cli.py
```diff
@@ -1490,160 +1501,146 @@ def kt_teardown(
...
+            teardown_all=teardown_all,
+            username=config.username if teardown_all else None,
+            exact_match=exact_match,
+        )
+        services_names = []
+        pools_to_delete = []
+        for service_type, service_names in services.items():
+            if service_type != "pools" and service_names:
+                services_names = services_names + list(service_names)
+            if service_type == "pools":
```
jlewitt1: we dont need to be surfacing pools here, thats more for our own internal management of the users launched compute

python_client/kubetorch/cli.py
```diff
@@ -1490,160 +1501,146 @@ def kt_teardown(
...
-        if not confirm:
-            console.print("[yellow]Teardown cancelled[/yellow]")
-            raise typer.Exit(0)
+        # Confirmation prompt for multiple services
+        if not yes:
+            if service_count >= 1:
+                console.print("The following resources will be deleted:")
+                for service_name in services_names:
+                    console.print(f" • [reset]{service_name}")
+            if byo_service_pools:
```
jlewitt1: can remove also

python_client/kubetorch/resources/callables/module.py
```diff
@@ -830,31 +830,29 @@ def teardown(self):
...
-            service_name=self.service_name,
-            namespace=self.compute.namespace,
-        )
-        if configmaps:
-            logger.info(f"Deleting {len(configmaps)} configmap{'' if len(configmaps) == 1 else 's'}")
-            delete_configmaps(
-                configmaps=configmaps,
-                namespace=self.compute.namespace,
+        try:
+            delete_result = delete_resources_for_services(
```
jlewitt1: is this function doing anything? cant we call `controller_client.delete_services()` directly?

python_client/kubetorch/resources/callables/module.py
```diff
@@ -830,31 +830,29 @@ def teardown(self):
...
-            service_name=self.service_name,
-            namespace=self.compute.namespace,
-        )
+            msg = f"Successfully force deleted {self.service_name}"
+
+            # BYO (selector-based) compute mode:
+            # The user applied the Kubernetes manifest themselves (e.g., via kubectl, Helm, or ArgoCD).
+            # Kubetorch did not create or own the K8s resources, so teardown only removes
+            # Kubetorch controller state and associated metadata — not the underlying pods/deployments/services
+            byo_deleted_services = delete_result.pop("byo_deleted_services")
```
jlewitt1, line 837: not sure i understand what we're doing with BYO compute here?

jlewitt1: cant we just have a single print here? im losing track of all the helpers that we're calling

python_client/kubetorch/cli.py
```diff
@@ -1490,160 +1501,146 @@ def kt_teardown(
...
-                        except Exception as e:
-                            if not http_not_found(e):
-                                console.print(f"[red]Failed to delete orphaned pod {pod_name}: {e}[/red]")
-            except Exception as e:
-                console.print(f"[red]Failed to list orphaned pods: {e}[/red]")
+            for service_type in deleted_services:
+                deleted_resources = deleted_services.get(service_type)
 
-    console.print("\n[green]Teardown completed successfully[/green]")
+                service_type = service_type[:-1]
```
jlewitt1: this seems fragile - how would knative be handled for example?

python_client/kubetorch/cli.py
```diff
@@ -1490,160 +1501,146 @@ def kt_teardown(
...
+                service_deletion_msg = f"{force_deleted_msg_prefix} [reset]{service_type}"
+
+                for resource in deleted_resources:
+                    console.print(f"{service_deletion_msg} [blue]{resource}[/blue]")
+                    resource_data_store_keys = deleted_data_store_keys.get(resource)
+                    if resource_data_store_keys:
+                        console.print(f'✓ Deleted cached data for {resource}"')
```
jlewitt1:
```suggestion
                        console.print(f'✓ Deleted cached data for {resource}')
```

python_client/kubetorch/cli.py
```diff
@@ -1490,160 +1501,146 @@ def kt_teardown(
...
+                console.print("The following resources will be deleted:")
+                for service_name in services_names:
+                    console.print(f" • [reset]{service_name}")
+            if byo_service_pools:
+                console.print("The following BYO resources will be deleted:")
+                for byo_service in byo_service_pools:
+                    console.print(f" • [reset]{byo_service}")
+
+        # Confirmation prompt for single service
+        if not yes and not force:  # if --force is provided, we don't need additional confirmation
```
jlewitt1: aren't we already inside not yes and not force block?

python_client/kubetorch/cli.py
```diff
@@ -1490,160 +1501,146 @@ def kt_teardown(
...
+                        console.print(f'✓ Deleted cached data for {resource}"')
+
+            console.print("\n[green]Teardown completed successfully[/green]")
+
+    except Exception as e:
+        if isinstance(e, ControllerRequestError):
+            handle_controller_delete_error(service_name=name, controller_error=str(e), console=console)
+            raise typer.Exit(1)
+        else:
+            console.print(f"[red] Failed to run `kt teardown`: {e}[/red]")
```
jlewitt1, line 1623: we should typer Exit here

python_client/kubetorch/cli.py
```diff
@@ -1490,160 +1501,146 @@ def kt_teardown(
...
-        configmaps = service_info["configmaps"]
-        service_type = service_info.get("type", "knative")
-        service_types.add(service_type)
-
-        delete_resources_for_service(
-            configmaps=configmaps,
-            name=name,
-            service_type=service_type,
+    try:
+        username = config.username if not prefix else None
```
jlewitt1: shouldnt it be `username = config.username if teardown_all else None` since `--all` is the only use case where we need to filter by username? this logic passes username when you're just deleting a single serice by name and no prefix is set

python_client/kubetorch/cli.py
```diff
@@ -1490,160 +1501,146 @@ def kt_teardown(
...
+        delete_result = delete_resources_for_services(
+            services=services,
```
jlewitt1: would it be possible to always pass in names, and not a dict? what benefit do we get by having all this different variable types sent to the controller?

python_client/kubetorch/globals.py
```diff
@@ -526,9 +526,47 @@ def get_service(self, namespace: str, name: str, ignore_not_found=False) -> Dict
...
+        params = {
+            "name": name,
+            "force": force,
+            "prefix": prefix,
+            "teardown_all": teardown_all,
+            "username": username,
+            "exact_match": exact_match,
+        }
+        params = {k: v for k, v in params.items() if v is not None}
+        return self.delete(f"/api/v1/namespaces/{namespace}/services", ignore_not_found=ignore_not_found, json=params)
```
jlewitt1: this isn't calling the new endpoint, its just calling the proxy endpoint (and same above)

python_client/tests/utils.py
BelSasha: to run teardown tests on stand-alone resources that are not being used by any other test

python_client/kubetorch/cli.py
BelSasha: The main reason I allowed passing a dict is that the it is the return type of `/teardown/list`. It returns a dict because it shares most of the logic with the `/discover` endpoint.  
We call `/teardown/list` when `--yes` is not provided. In this case: 
1. If we pass a list of names instead of a dict, it means we gonna have to run additional discover logic on the controller, even though we run it on the OSS side. `/teardown/list` was designed to get us all the information we need about the resources that we need to delete. 
2. If we case the dict to a list of names, we gonna have to make a lot of unnecessary parsing. Moreover, we already have some sort of parsing logic those resources in the controller (which i tried to minimize as much as possible). So i don't think we should avoid parsing when we can.

jlewitt1:
### Merge activity

* **Jan 15, 10:04 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/run-house/kubetorch/2113).
* **Jan 15, 10:04 AM UTC**: @jlewitt1 merged this pull request with [Graphite](https://app.graphite.com/github/pr/run-house/kubetorch/2113).

PR #2114, opened 2026-01-12
Makes DCGM exporter config cloud-agnostic for BYOC users

cezarc1:
values.yaml currently configures the DCGM Prometheus scrape to only 
search gke-managed-system namespace, which is correct and [documented by GKE](https://docs.cloud.google.com/kubernetes-engine/docs/how-to/dcgm-metrics).

However, this breaks GPU metrics collection on [AKS](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/microsoft-aks.html)/[EKS](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/amazon-eks.html) where the GPU Operator deploys DCGM to `gpu-operator` namespace.

This PR updates the DCGM scrape config to search all common DCGM namespaces 
(should now work on GKE/AKS/EKS):
- `gke-managed-system` (GKE managed DCGM)
- `gpu-operator` (NVIDIA GPU Operator default)
- `nvidia-gpu-operator` (GPU Operator alternate)

## Before

```bash
$ helm template test ./charts/kubetorch | grep job_name
- job_name: 'self'
- job_name: 'kubelet-cadvisor'
- job_name: 'pods'
```

## After

```bash
$ helm template test ./charts/kubetorch | grep job_name
- job_name: 'self'
- job_name: 'kubelet-cadvisor'
  job_name: dcgm-exporter
- job_name: 'pods'
```


charts/kubetorch/README.md.gotmpl
```diff
@@ -0,0 +1,18 @@
+{{ template "chart.header" . }}
```
py-rh: Don't need to add this file, we don't use Helm-Docs currently for the readme

cezarc1: Done.

jlewitt1 (approved):

PR #2127, opened 2026-01-13
Store image contents as dockerfile lines

carolineechen:
Directly store image contents from setup steps as a dockerfile line string, computed in the image class rather than the compute class at launchtime. Separately store list of rsync kwargs for launchtime rsyncing. + subsequent refactoring from these changes

Also add support for `Image.from_dockerfile()`

### DIFF
--- a/python_client/kubetorch/resources/compute/compute.py
+++ b/python_client/kubetorch/resources/compute/compute.py
@@ -19,8 +19,8 @@ from kubetorch.logger import get_logger
 from kubetorch.provisioning.autoscaling import AutoscalingConfig
 from kubetorch.provisioning.utils import pod_is_running, SUPPORTED_TRAINING_JOBS
 from kubetorch.resources.callables.utils import find_locally_installed_version
-from kubetorch.resources.compute.utils import _get_sync_package_paths, _run_bash
-from kubetorch.resources.images.image import Image, ImageSetupStepType
+from kubetorch.resources.compute.utils import _run_bash
+from kubetorch.resources.images.image import Image
 from kubetorch.resources.secrets.kubernetes_secrets_client import KubernetesSecretsClient
 from kubetorch.resources.volumes.volume import Volume
 from kubetorch.serving.utils import is_running_in_kubernetes, load_template
@@ -1135,7 +1135,7 @@ class Compute:
             return self.image.python_path
 
         container = self._container()
-        if "env" in container:
+        if container and "env" in container:
             for env_var in container["env"]:
                 if env_var["name"] == "KT_PYTHON_PATH" and "value" in env_var:
                     return env_var["value"]
@@ -2359,7 +2359,7 @@ class Compute:
     ):
         """Pip install reqs onto compute pod(s)."""
         reqs = [reqs] if isinstance(reqs, str) else reqs
-        python_path = self.image.python_path if self.image else "python3"
+        python_path = self.python_path if self.python_path else "python3"
         pip_install_cmd = f"{python_path} -m pip install"
         try:
             result = self.run_bash("cat .kt/kt_pip_install_cmd 2>/dev/null || echo ''", node=node)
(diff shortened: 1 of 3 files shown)


carolineechen:
* **#2243** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2243?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2228** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2228?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2227** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2227?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2213** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2213?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2194** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2194?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2085** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2085?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2128** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2128?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2127** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2127?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2127?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python_client/kubetorch/resources/images/image.py
```diff
@@ -86,34 +61,34 @@ def __init__(
...
-    def _setup_step_from_config(step: Dict):
-        """Convert setup step config (dict) to ImageSetupStep object"""
-        step_type = step["step_type"]
-        kwargs = step["kwargs"]
-        return ImageSetupStep(
-            step_type=ImageSetupStepType(step_type),
-            **kwargs,
-        )
+    @property
+    def contents(self) -> str:
```
carolineechen, line 83: @dongreenberg is this sufficient or do you want it in string form or an option to output to a file? or different naming (dockerfile, to_dockerfile)

python_client/kubetorch/resources/compute/compute.py
```diff
@@ -2392,86 +2392,32 @@ def _image_setup_and_instructions(self, rsync: bool = True):
...
-                for key, val in step.kwargs.get("env_vars").items():
-                    # single env var per line in the dockerfile
-                    instructions += f"ENV {key} {val}\n"
-            if step.kwargs.get("force") and step.step_type != ImageSetupStepType.PIP_INSTALL:
-                instructions += " # force"
-            instructions += "\n"
-
-        return instructions
+            lines.append(f"ENV KT_PYTHON_PATH {self.image.python_path}")
+        lines.extend(self.image._dockerfile_contents)
```
dongreenberg: Feels a little roudabout - shouldn't these lines just be in like a property or get_dockerfile() method in [Image.py](http://Image.py) ?

python_client/kubetorch/resources/images/image.py
dongreenberg: This is fine!

python_client/kubetorch/resources/compute/compute.py
carolineechen: The reason for this is that server_image and python_path might be set outside of the image, but we should still include it in this dockerfile. For instance if an image doesn't have an image_id, the compute still has a server_image (the default one we provide), or python path can also be directly specified through an env var workaround, setting `KT_PYTHON_PATH` through the compute env vars. 

I think they aren't strictly necessary to include in this dockerfile we send over as they're used prior to the http server image steps (image is pulled beforehand + kt python path is used for installation and setup, don't believe it's used within the http server) but figured for consistency we can include them even if it's not part of or different from the image class value.

dongreenberg (approved):
As an aside, I can't see exactly where this is, but when we send a fn or cls to compute, it would be nice to actually properly append the workdir as a COPY command to the dockerfile, rather than implicitly upload it on the side. Then we could let the upload of the workdir flow through the usual mechanism, not a separate upload call, and the dockerfile would actually be buildable as the service.

PR #2132, opened 2026-01-14
Improves Prometheus scrape config defaults for remote write and Knative

cezarc1:
Two improvements to prometheus metrics collection:

  1. **Enables, by default, Prometheus Knative metrics**: Enables queue-proxy metrics export by adding a `observability` config to KnativeServing CR and port 9091. See https://knative.dev/docs/serving/observability/metrics/collecting-metrics/

  2. **Remote write behavior changes**: Enables `retry_on_http_429` to prevent sample loss during rate limiting. Sets cAdvisor scrape interval to 30s (industry standard) from 3s to reduce sample ingestion rate.


py-rh:
Discussed and aligned on the configs in the yaml for user-side platform team configuration rather than being feature complete to all forwarding configs

PR #2133, opened 2026-01-14
Support BYO manifests with arbitrary K8s resource types

jlewitt1:


jlewitt1:
* **#2211** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2211?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2133** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2133?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2133?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python_client/kubetorch/provisioning/templates/pod_template.yaml
```diff
@@ -20,7 +20,7 @@ affinity:
               operator: DoesNotExist
 {% endif %}
 
-{% if node_selector is not none %}
+{% if node_selector is not none and node_selector %}
```
jlewitt1, line 23: to ignore empty dict

python_client/kubetorch/provisioning/utils.py
```diff
@@ -371,12 +371,21 @@ def nested_merge(user_dict, kt_dict):
...
+    """
     resource_type = resource_type.lower()
     if resource_type not in RESOURCE_CONFIGS:
-        raise ValueError(
-            f"Unknown resource type: {resource_type}. " f"Supported types: {', '.join(RESOURCE_CONFIGS.keys())}"
-        )
+        # Use generic config for BYO manifests - these are handled via the dynamic client on the controller side
+        # Note: most K8s workloads (StatefulSet, DaemonSet, Job, ReplicaSet) use spec.template
+        return {
+            "pod_template_path": ["spec", "template"],
```
dongreenberg, line 384: Isn't this settable by the user when they create the manifest? If it's a more complex manifest there could be nesting and the pod template could be deeper down

jlewitt1: oh yea i see your point, that is a gap. We could have a layered approach: 
1. Check `RESOURCE_CONFIGS` for known types → use that                                                                                                                                                                                   
2. If `pod_template_path` param (new param) is specified → use that (explicit override)                                                                                                                                                               
3. Try auto-detection → find pod-like structure                                                                                                                                                                       
4. Fall back to `[“spec”, “template”]` if nothing found  

just added that new param, wdyt?

dongreenberg: Ya should just be an explicit path provided by the user for now

dongreenberg: Ya should just be an explicit path provided by the user for now

python_client/tests/test_byo_manifest.py
```diff
@@ -525,6 +525,180 @@ def test_from_manifest_getters_setters():
...
+    Uses a Deployment with an explicit pod_template_path to verify the override
+    mechanism works end-to-end. When pod_template_path is provided, kubetorch
+    skips merging defaults and uses the manifest as-is.
+    """
+    from .conftest import KUBETORCH_IMAGE
+
+    pool_name = f"{kt.config.username}-byo-path-override"
+    namespace = kt.globals.config.namespace
+
+    # Deployment manifest with explicit pod_template_path
```
dongreenberg: Can we make this a resource type we don't explicitly support, and that doesn't accept the typical spec.template path? JobSet is one that comes to mind, but maybe there's a resource type built into K8s that would work.

dongreenberg: Oh maybe just a pod manifest would work actually, because it doesn't have "spec.template", it's just at "spec"

python_client/tests/test_byo_compute.py
```diff
@@ -278,7 +278,6 @@ def test_byo_compute_run_bash(byo_pod_compute):
     # Deploy a simple function to get a reference to the compute
     remote_fn = kt.fn(summer, name=service_name).to(compute)
 
-    # Verify function works
```
dongreenberg, line 281: 🦾

python_client/kubetorch/resources/compute/compute.py
```diff
@@ -612,32 +640,41 @@ def logging_config(self, value: LoggingConfig):
...
+        return current
 
     @pod_spec.setter
     def pod_spec(self, value: dict):
         """Set the pod spec in the manifest."""
-        template_path = self.service_manager.get_pod_template_path()
+        template_path = self._get_pod_template_path()
+        if not template_path:
+            # Unknown resource type - cannot set pod spec
+            return
```
dongreenberg: Shouldn't this be an error instead of failing silently?

python_client/tests/test_byo_manifest.py
jlewitt1: i added a new test in a follow on PR in this stack which uses a jobset (and in the process uncovers some addiitonal changes).

PR #2140, opened 2026-01-15
Fix `sys.path` for scripts in subdirectories to enable sibling package imports

jlewitt1:
- Fixes import failures when scripts are located in subdirectories and need to import from sibling packages at the project root                                                                                                        
  - Adds project root path to `sys.path` and `PYTHONPATH` on remote pods, not just the script's directory      

Corresponding controller PR: https://github.com/run-house/kubetorch-internal/pull/1022

### DIFF
--- a/python_client/kubetorch/serving/http_server.py
+++ b/python_client/kubetorch/serving/http_server.py
@@ -264,6 +264,8 @@ class ControllerWebSocket:
             os.environ["KT_CLS_OR_FN_NAME"] = module_info["cls_or_fn_name"]
         if module_info.get("file_path"):
             os.environ["KT_FILE_PATH"] = module_info["file_path"]
+        if module_info.get("project_root"):
+            os.environ["KT_PROJECT_ROOT"] = module_info["project_root"]
         if module_info.get("callable_type"):
             os.environ["KT_CALLABLE_TYPE"] = module_info["callable_type"]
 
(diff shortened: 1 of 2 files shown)


jlewitt1:
* **#2140** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2140?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2140?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python_client/kubetorch/serving/http_server.py
```diff
@@ -975,11 +977,29 @@ def patch_sys_path():
...
+    project_root = os.environ.get("KT_PROJECT_ROOT")
+    if project_root:
+        abs_project_root = str(Path(project_root).expanduser().resolve())
+        if abs_project_root not in sys.path:
+            sys.path.insert(0, abs_project_root)
+            logger.debug(f"Added project root {abs_project_root} to sys.path")
+
     # Maybe needed for subprocesses (e.g. distributed) to find the callable's module
     # Needed for distributed subprocesses to find the file path
     existing_path = os.environ.get("PYTHONPATH", "")
```
carolineechen, line 991: `existing_path` here is a string, so checking if `project_root`/`kt_file_path` is a substring of `existing_path` might not be entirely bug proof (e.g. if a subdir of `project_root` was in the path, but not the dir we intended to set or another dir's prefix matched)

```suggestion
    `existing_path_list = existing_path.split(os.pathsep) if existing_path else []`
```

python_client/kubetorch/serving/http_server.py
```diff
@@ -975,11 +977,29 @@ def patch_sys_path():
         sys.path.insert(0, abs_path)
         logger.debug(f"Added {abs_path} to sys.path")
 
+    # Add project root to sys.path to enable imports from sibling directories
+    # This allows scripts in subdirectories (e.g., experimental/script.py) to import
+    # from sibling packages at the project root (e.g., from utils.module import func)
+    project_root = os.environ.get("KT_PROJECT_ROOT")
+    if project_root:
+        abs_project_root = str(Path(project_root).expanduser().resolve())
```
carolineechen: nit for clarity/reuse later
```suggestion
    abs_project_root = str(Path(project_root).expanduser().resolve()) if project_root else None
    if abs_project_root and abs_project_root not in sys.path:
          ...
```

python_client/kubetorch/serving/http_server.py
```diff
@@ -975,11 +977,29 @@ def patch_sys_path():
...
+            logger.debug(f"Added project root {abs_project_root} to sys.path")
+
     # Maybe needed for subprocesses (e.g. distributed) to find the callable's module
     # Needed for distributed subprocesses to find the file path
     existing_path = os.environ.get("PYTHONPATH", "")
+    paths_to_add = []
     if os.environ["KT_FILE_PATH"] not in existing_path:
-        os.environ["PYTHONPATH"] = f"{abs_path}{os.pathsep}{existing_path}" if existing_path else abs_path
+        paths_to_add.append(abs_path)
+    if project_root and project_root not in existing_path:
```
carolineechen: we should probably check for absolute path (`abs_path`/`abs_project_root`) rather than relative path here to be more explicit, especially if we switch existing_path to be a list instead of a string. (as a string it will likely have caught the relative substring fine but not necessarily always)

jlewitt1: yea good point

PR #2151, opened 2026-01-18
Add empty limits to Helm chart to avoid breaking the schema

BelSasha:
Even though we don't want to set values to `limits`, we need to keep those keys in the helm chart. Otherwise the schema brakes, and helm installing on a fresh cluster fails: 
```* Deployment.apps "kubetorch-controller" is invalid: [spec.template.spec.containers[0].resources.requests: Invalid value: "200m": must be less than or equal to cpu limit of 0, spec.template.spec.containers[0].resources.requests: Invalid value: "256Mi": must be less than or equal to memory limit of 0, spec.template.spec.containers[1].resources.requests: Invalid value: "1": must be less than or equal to cpu limit of 0, spec.template.spec.containers[1].resources.requests: Invalid value: "2Gi": must be less than or equal to memory limit of 0]```

cc: @dongreenberg

### DIFF
--- a/charts/kubetorch/templates/controller/deployment.yaml
+++ b/charts/kubetorch/templates/controller/deployment.yaml
@@ -55,9 +55,15 @@ spec:
             requests:
               cpu: {{ .Values.kubetorchController.nginx.resources.cpu.request | quote }}
               memory: {{ .Values.kubetorchController.nginx.resources.memory.request | quote }}
+            {{- if or .Values.kubetorchController.nginx.resources.cpu.limit .Values.kubetorchController.nginx.resources.memory.limit }}
             limits:
+              {{- if .Values.kubetorchController.nginx.resources.cpu.limit }}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by BelSasha** [comment]
* **#2151** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2151?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2151?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

File: charts/kubetorch/values.yaml [hunk 1]
```diff
@@ -29,8 +29,10 @@ kubetorchController:
   resources:
     cpu:
       request: "1"   # Minimum guaranteed CPU
+      limit: ""      # Maximum CPU
```
jlewitt1: i think would be cleaner to remove the empty strings here like we do now and just conditionally handle it in the `deployment.yaml`

PR #2170, opened 2026-01-20
Fix installs in distributed tests

mkandler:
A little strange that `test_pytorch_distributed_cls` was passing because both it and `test_pytorch_distributed_fn` were failing the `uv install` command because it was searching for `numpy` on the torch index.

Might want to make that test a little more robust, since it doesn't seem to actually require the installs to pass.

### DIFF
--- a/python_client/tests/test_distributed.py
+++ b/python_client/tests/test_distributed.py
@@ -232,9 +232,9 @@ def test_pytorch_distributed_fn():
         kt.Compute(
             cpus="0.5",
             memory="1Gi",
-            image=kt.images.Debian().run_bash(
-                "uv pip install --system numpy torch --index-url https://download.pytorch.org/whl/cpu"
-            ),
+            image=kt.images.Debian()
+            .run_bash("uv pip install --system numpy")
+            .run_bash('uv pip install --system torch --extra-index-url "https://download.pytorch.org/whl/cpu"'),
         ).distribute("pytorch", workers=2, num_proc=2)
     )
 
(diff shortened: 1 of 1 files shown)


mkandler:
* **#2170** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2170?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2170?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python_client/tests/test_distributed.py
```diff
@@ -266,9 +266,9 @@ def test_pytorch_distributed_cls():
         kt.Compute(
             cpus="0.5",
             memory="1Gi",
-            image=kt.images.Debian().run_bash(
-                "uv pip install --system numpy torch --index-url https://download.pytorch.org/whl/cpu"
-            ),
+            image=kt.images.Debian()
+            .run_bash("uv pip install --system numpy")
+            .run_bash("uv pip install --system torch --index-url https://download.pytorch.org/whl/cpu"),
```
jlewitt1: good catch, i think it was passing bc we have a conditional in the test, so instead of 
```
        if "pytorch_initialized" in result:
            if result["pytorch_initialized"]:
```

we can change to something like: 
```
assert result.get("pytorch_initialized") is True, \
            f"PyTorch not initialized on rank {i}: {result.get('pytorch_error', 'unknown error')}"
```

mkandler: Yeah good call, updated!

jlewitt1 (approved):

PR #2190, opened 2026-01-22
Allow setting kt config values to None

BelSasha:


BelSasha:
* **#2190** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2190?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2190?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python_client/kubetorch/provisioning/templates/kt_setup_template.sh.j2
```diff
@@ -84,7 +84,11 @@ echo "$uv_pip_cmd" > .kt/kt_pip_install_cmd
     $uv_pip_cmd -e "python_client[server]"
   {% endif %}
 {% else %}
-  $uv_pip_cmd "kubetorch[server]=={{ install_url }}"
+  {% if install_url %}
+    $uv_pip_cmd "kubetorch[server]=={{ install_url }}"
+  {% else %}
```
carolineechen: what's the motivation and expected behavior for this change? `$uv_pip_cmd -e "python_client[server]"` here is happening outside of the `if rsync_kt_local_cmd`, meaning that kubetorch is probably not installed as a local editable or synced over, and `python_client` would not exist on the compute (it likely works in our tests because python_client is also the git working dir)

BelSasha: Actually this change is no longer required, the root problem was in `kt config set`.

PR #2198, opened 2026-01-23
Run image setup steps in lifespan

carolineechen:
run_image_setup was removed in _load_callable_internal in #2185, causing packages to not be installed in setup + CI failures. add this to lifespan to run image setup prior to loading callable

### DIFF
--- a/python_client/kubetorch/serving/http_server.py
+++ b/python_client/kubetorch/serving/http_server.py
@@ -868,8 +868,10 @@ def load_callable(
     With push-based reloads, this is always called fresh on reload - no need to
     check timestamps since the reload is triggered externally.
 
-    Note: run_image_setup() is NOT called here. For reload, _handle_reload() calls
-    it before load_callable(). For subprocesses, the main process already did it.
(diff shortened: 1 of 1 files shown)


carolineechen:
* **#2198** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2198?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2198?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python_client/kubetorch/serving/http_server.py
carolineechen, line 1318: @dongreenberg does it make sense to include `run_image_setup` for selector only mode too? in kt apply work I added an `image` field to `from_manifest`, so it would be possible to for a user to pass in both fields. does it make sense to support that or should we throw an error if both are provided / handle the --dockerfile image differently in kt apply?

PR #2226, opened 2026-01-30
Optimizations: Adjusts timeouts, worker pool singleton

mkandler:
@dongreenberg - some minor optimizations from running the scale test.
- Updated pod_template timeouts to stop showing so many `Readiness probe failed:` warnings
- Retry backoff for pod IPs
- Singleton to skip RemoteWorkerPool recreation on consecutive runs

Main slowness is still recreating ProcessPool after a .to - might need to check a code hash or something to determine if we can keep existing subprocesses.

### DIFF
--- a/python_client/kubetorch/serving/spmd/spmd_supervisor.py
+++ b/python_client/kubetorch/serving/spmd/spmd_supervisor.py
@@ -376,11 +376,8 @@ class SPMDDistributedSupervisor(DistributedSupervisor):
         if subcall_ips:
             logger.debug(f"Have {len(subcall_ips)} remote workers to call")
             if not self.remote_worker_pool:
-                # Create RemoteWorkerPool lazily when first needed
-                # This avoids spawning unnecessary processes for single-pod distributed jobs
-                logger.info(f"Creating RemoteWorkerPool to call {len(subcall_ips)} remote workers")
-                self.remote_worker_pool = RemoteWorkerPool(quorum_timeout=self.quorum_timeout)
-                self.remote_worker_pool.start(max_workers=min(len(subcall_ips) + 50, 200))
+                # Get the singleton RemoteWorkerPool (created lazily, reused across supervisors)
+                self.remote_worker_pool = RemoteWorkerPool.get_instance(quorum_timeout=self.quorum_timeout)
             logger.debug(f"Using RemoteWorkerPool to call {len(subcall_ips)} workers")
 
             def call_remote_workers():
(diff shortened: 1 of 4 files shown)


mkandler:
* **#2226** <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2226?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/run-house/kubetorch/2226?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

python_client/kubetorch/serving/spmd/spmd_supervisor.py
```diff
@@ -376,11 +376,8 @@ def call_worker(worker_ip):
         if subcall_ips:
             logger.debug(f"Have {len(subcall_ips)} remote workers to call")
             if not self.remote_worker_pool:
-                # Create RemoteWorkerPool lazily when first needed
-                # This avoids spawning unnecessary processes for single-pod distributed jobs
-                logger.info(f"Creating RemoteWorkerPool to call {len(subcall_ips)} remote workers")
-                self.remote_worker_pool = RemoteWorkerPool(quorum_timeout=self.quorum_timeout)
-                self.remote_worker_pool.start(max_workers=min(len(subcall_ips) + 50, 200))
+                # Get the singleton RemoteWorkerPool (created lazily, reused across supervisors)
+                self.remote_worker_pool = RemoteWorkerPool.get_instance(quorum_timeout=self.quorum_timeout)
```
dongreenberg, line 380: This is a bit of a funky pattern - I'd rather leave RemoteWorkerPool as a regular class and just pass the singleton into the supervisor, or make it a global.

mkandler: I think the benefit of this pattern is that the class method `get_instance` allows for a combination of global storage (as a singleton) and lazily loading. Otherwise we'd have to implement something like:
```python
if globals.remove_worker_pool:
    self.remote_worker_pool = globals.remove_worker_pool
else:
    self.remote_worker_pool = RemoteWorkerPool(...)
    globals.remote_worker_pool = self.remote_worker_pool
```
is it the singleton pattern you don't like here? This seems a lot cleaner, especially when the singleton is used in more than one place and if we want to keep the lazy loading. If you want to drop the lazy loading, I can see a stronger argument to switch to a global.

mkandler: The class method is essentially just a cleaner way to reference the global

dongreenberg: I guess the singleton is basically just a global, but stored as a class variable, so not a big difference. All good

